PR #93, opened 2020-12-14
Fix/svgs as fonts

meganrm:
This is not a critical change, but also shouldn't effect any of the bug testing. 

[I changed the icons on the step back and step forward buttons from images to fonts](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1241). This allows us to change the color of the icon for different states, in this case, the disabled state. 

Before, there was no visual indication that the button was disabled: 
Screen Shot 2020-12-14 at 11 13 08 AM

Now, the background and the image itself is changed when disabled:
Screen Shot 2020-12-14 at 11 12 49 AM

Thanks for contributing!


EricJIsaac (approved):

src/style.css
```diff
@@ -87,8 +87,18 @@ canvas:focus {
     content: "\e9d9";
 }
 
+.icon-step-back:before {
+    content: "\e906";
+    color: #d8d8d8;
```
schoinh, line 92: You can use `var(--dark-theme-btn-color)` for `#d8d8d8` here and in line 97

schoinh (approved):
You'll have to show me how to do the Icomoon font thing sometime! Or point me to the part of the docs that you used. :)

toloudis (approved):
My impression is that the button background looks too "bright" to be disabled, but the implementation LGTM!

schoinh:
> My impression is that the button background looks too "bright" to be disabled, but the implementation LGTM!

I wasn't gonna say anything because I thought it was just me, but I agree. Maybe just dimming the icon color and leaving the background color as is (or making the background a tad darker) would be better? Don't feel super strong about it 🤷‍♀️

blairlyons (approved):
As is it's better than no indication, but if you want to tweak the colors that's fine with me

meganrm:
yeah, I agree, this is the color change that was also being used for the Load Model button, and for it going from purple to a gray works, but for these, I think just dimming the image is probably better

meganrm:
Screen Shot 2020-12-16 at 5 29 45 PM

PR #97, opened 2020-12-15
Fix/wrap long plot titles

schoinh:
Resolves: [Wrap long plot titles](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1161)

This problem was more obvious when we had listeria as an example simulation, and the latest plots for that had some titles that were too long. Not a problem with current endocytosis and COVID plots but could be a problem with other user- or AICS-uploaded plots.

Unfortunately plotly.js doesn't provide wrapping options so I wrote a util function and some tests. Once the text is wrapped, we also need to make more room vertically, and having a potentially multi-line title also makes it difficult to position the title in the right place with respect to its container. That's why the position of the title doesn't look as consistent as before, but I couldn't figure out how to make it more consistent.

Do you think this PR might be too risky for this week? Maybe we can lower the risk by adding a try-catch for the `wrapText` call in case the function implodes?

### Before
A listeria plot from simularium_text_trajs06:
image

image

### After
image

Notice how the plot title top margins are not as consistent as before. Hopefully it's not too egregious??
image

Also saved some plot trace colors as constants so they're a little easier to identify.

- [x] Provide relevant tests for your feature or bug fix.

Thanks for contributing!


src/util/index.ts
```diff
@@ -28,3 +28,47 @@ export function convertToSentenceCase(string: string): string {
             return c.toUpperCase();
         });
 }
+
+export const wrapText = (
```
toloudis, line 32: Why can't we let the browser do this? Can we not control the width of the title section and just let the browser flow the text in it?

toloudis: I googled and I see that plotly just doesn't support this yet.

schoinh: Haha........ I'm glad your Googling led to the same results I came across 😅 Or am I? I would love to not have to manually wrap

schoinh: I just tried again, just to be extra sure (trying to limit the container of the title), and I couldn't get it to work. I think it might have something to do with the entire plot being rendered as an svg???

blairlyons (approved):

toloudis (approved):
I suggest adding a link to the plotly issue that talks about this, in a comment near where you call your wrap function

https://github.com/plotly/plotly.js/issues/2053

schoinh:
@meganrm I'd like to get your review for this one too!

meganrm (approved):

PR #101, opened 2020-12-17
Fix/stop play on navigation

meganrm:
This pauses the sim anytime the user navigates away from the viewer. This still doesn't solve the problem of: user hits play, navigates away from viewer while first batch is buffering, user selects new or same trajectory. This usually still results in the viewer playing some combination of data, but I think this needs to be handled by the viewer, not the front end. 

I also combined the other location based code for the viewer into one custom hook. 

- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_

Thanks for contributing!


src/index.tsx
```diff
@@ -4,24 +4,84 @@ import "core-js/es6/set";
...
+    getSimulariumFile,
+} from "./state/metadata/selectors";
 
 export const store = createReduxStore();
+interface LocationWithState extends Location {
+    state: {
+        localFile?: boolean;
+    };
+}
+function useLocationChangeToStopPlay() {
```
schoinh, line 34: Maybe this function should be named something more like `cleanUpViewerState`? Since stopping play is only done in one possible scenario (leaving viewer), no?

schoinh (approved):

toloudis (approved):

PR #107, opened 2021-01-12
Fix/simulation keeps playing when user is trying to scrub

schoinh:
Resolves: [Simulation keeps playing when user is trying to scrub](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1230)

**Current behavior:** When a simulation is playing, and a user drags the playhead without pressing the pause button first, the simulation keeps trying to play while the user is trying to scrub, so it becomes a tug-o-war. 

**Behavior implemented in this PR:** The simulation pauses when a user starts scrubbing, and once the user lets go (mouse up) the simulation automatically resumes at the new position. The user has full control when they're trying to scrub and doesn't have to compete with ongoing playback.

Thanks to Megan for tons of help with this issue.

Thanks for contributing!

### DIFF
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -47,7 +47,7 @@ const PlayBackControls = ({
         // but we're just using a single value
         onTimeChange(sliderValue as number);
         if (isPlaying) {
-            // Need to save the sliderValue in local state to use in handleSliderMouseUp,
+            // Need to save the sliderValue as targetPlayTime to use in handleSliderMouseUp,
             // because the sliderValue argument available in handleSliderMouseUp is not accurate
             // when the time between mouse down and mouse up is short.
             setTargetPlayTime(sliderValue as number);
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -39,8 +39,12 @@ const PlayBackControls = ({
     const [wasPlaying, setWasPlaying] = useState(false);
     const [targetPlayTime, setTargetPlayTime] = useState(-1);
 
+    // - Gets called once when the user clicks on the slider to skip to a specific time
+    // - Gets called multiple times when user is scrubbing (every time the play head
+    //     passes through a time value associated with a frame)
     const handleTimeChange = (sliderValue: number | [number, number]): void => {
-        // slider can be a list of numbers, but we're just using a single value
+        // sliderValue can be an array of numbers (representing a selected range),
+        // but we're just using a single value
         setTargetPlayTime(sliderValue as number);
         if (isPlaying) {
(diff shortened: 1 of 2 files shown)


meganrm (approved):

src/containers/ViewerPanel/index.tsx
```diff
@@ -222,7 +224,7 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
             setBuffering,
             setIsPlaying,
         } = this.props;
-        let newTime = time;
+        let newTime = timeOverride || time;
```
toloudis, line 227: what if timeOverride is zero vs undefined?

toloudis: for optional numbers you might have to do an explicit undefined check instead of this falsy-ness check

src/components/PlaybackControls/index.tsx
```diff
@@ -35,11 +35,36 @@ const PlayBackControls = ({
...
+        } else if (targetPlayTime >= 0) {
+            // Update targetPlayTime if user is still dragging
+            setTargetPlayTime(sliderValue as number);
+        }
     };
 
-    const [unitIndex, setUnitIndex] = useState(0);
+    const handleSliderMouseUp = (): void => {
+        // Resume playing if simulation was playing before
+        if (targetPlayTime >= 0) {
```
toloudis, line 63: to see if i understand, you are using -1 to indicate that there is "no target play time"? and that is because all valid times are always non-negative?

src/containers/ViewerPanel/index.tsx
schoinh: That is a really good point, thanks!

src/components/PlaybackControls/index.tsx
schoinh: yes

toloudis: Is targetPlayTime used anywhere else?  If not, then I would consider naming it something like savedPlayTimeWhileScrubbing to make it really obvious.

schoinh: haha, actually its original name was `wasPlayingBeforeScrubbing` (it started out as a boolean)... how about `timeToResumeAfterScrubbing`? @meganrm any opposition?

toloudis: sounds good to me.  targetPlayTime sounds like something that has a bigger meaning

meganrm: sounds good to me too

schoinh: Excellent I have renamed it

toloudis (approved):

PR #109, opened 2021-01-25
Fix/only re-render plots that have the time indicator bar

schoinh:
### Current (problematic) behavior
Endocytosis playback speed on prod and dev maxes out at 3 fps. This is because (1) all of the plots associated with this model are being re-configured at every frame with the `getPlotDataConfiguredForPlotly` selector, which takes the dynamic `currentTime` as an input, and because (2) an unused `time` (equivalent to `currentTime`) prop is being passed into `ResultsPanel` and then into the `Plots` component, making `ResultsPanel` and `Plots` run the `render` method at every frame.

### Proposed solution
During team discussion, we came up with a 2-pronged approach to increase playback speed:
1. Have React only re-render plots that have the time indicator bar
2. Only load/re-render plots that are currently in view

This PR implements part 1 of the above plan. JIRA: [Improve endocytosis playback speed](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1257)

### Changes in this PR
1. So that we can choose which plots get re-rendered and which don't (with the [`shouldComponentUpdate()` lifecycle method](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)), I broke up the `Plots` component (handling all plots collectively) into individual `Plot` components each responsible for one Plotly plot.
2. Any component that receives `currentTime` as a prop would by default re-render at every frame change (e.g. the `ResultsPanel` component which is receives `currentTime` from the redux store and passes it down to `Plots`), so I put the `Plot` component into `containers` and directly connected it to redux so that only this component would be affected by time/frame changes.
3. I made the `getPlotDataConfiguredForPlotly()` selector get called only at the beginning of a trajectory load by removing the `currentTime` input from it. Its previous job was to configure (add Plotly styles and layout to) all of the plots _and_ to add the time indicator bar at the correct location at every time/frame change. Now its job is to configure all the plots when the trajectory first gets loaded, which now includes "tagging" each plot with a the appropriate `hasTimeIndicator` boolean.
4. The job of adding and updating the time indicator bar now falls to the individual `Plot` component, which uses the `hasTimeIndicator` property from the plot data received to determine whether to add/update a vertical bar and to re-render itself at every time change (with `shouldComponentUpdate()`).
5. Variable renaming just for clarity
    * `plotConfig` --> `plotConfigs` (it's an array of plot configurations)
    * `PlotParamsWithKey` --> `PlotConfig` (this is the type interface for a plot configuration)

Phew I hope that makes sense. The main lesson I learned from this is that even if React doesn't actually re-draw components that are the same in the virtual DOM and the real DOM, whenever a state/prop changes it will still do all the computation necessary to determine whether the component has changed. Duh!

### Performance improvement
With the above changes, I've been consistently seeing endocytosis playing at 5 to 6 fps on dev. I expect this to improve even more when we implement part 2 of the optimization.

**Before (master branch on dev):**
image

**After (this branch):**
image

Thanks @meganrm for tons of help!!

Thanks for contributing!


src/containers/Plot/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { PlotConfig } from "../ResultsPanel/types";
+import { getCurrentTime } from "../../state/selection/selectors";
+// import { changeTime } from "../../state/selection/actions";
+// import { ChangeTimeAction } from "../../state/selection/types";
+import { PLOT_STYLE } from "../ResultsPanel/constants";
+
+interface PlotProps {
+    plotConfig: PlotConfig;
+    time: number;
+    // changeTime: ActionCreator<ChangeTimeAction>;
```
meganrm, line 16: remove commented out code (here and in the imports)?

schoinh: I left them in thinking we might bring these back soon, but I can just remove them too :)

src/containers/Plot/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const styles = require("./style.css");
+
+class Plot extends React.Component<PlotProps, {}> {
+    shouldComponentUpdate() {
+        return this.props.plotConfig.hasTimeIndicator;
```
meganrm, line 23: I'm curious if this acts the same as `return "prevTime !== currentTime" `. Did you try that? I dont think plot config is ever changing otherwise, and it's the only other prop, so it I think it would act the same

src/containers/Plot/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public render(): JSX.Element | null {
+        const { plotConfig, time } = this.props;
+        const { hasTimeIndicator, data, layout } = plotConfig;
+        const TIME_INDICATOR_LINE = "timeIndicatorLine";
+
+        if (hasTimeIndicator && time !== 0) {
```
meganrm, line 31: should this be `start time` instead of `0`... since we're allowing for start time to not be zero?

meganrm, line 31: I would call this prop `renderTimeIndicator` or `shouldRenderTimeIndicator` because it doesn't mean it has one, it means it should have one

src/containers/Plot/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+            const lastPlot = data[data.length - 1];
+            // Update the time for the time indicator line if it already exists,
+            // otherwise add a time indicator line
+            if (lastPlot.name === TIME_INDICATOR_LINE) {
```
meganrm, line 35: For clarity I would probably functionalize this into methods 
```findTimeIndicatorPlot() => find(this.props.data), {name : TIME_INDICATOR_LINE}```

 that just checks if there is a plot named `TIME_INDICATOR_LINE` and returns it (the getting the last plot would work fine here, and is faster, so you can do that too, I just think this is an clearer check ). 

Another one that is `updateTimeIndicatorPosition(timePlot) => timePlot.x[this.props.time, this.props.time]` that updates the plot, 

and a `addTimeIndicator()` that does lines 38-53

File: src/containers/Plot/index.tsx [hunk 5]
```diff
@@ -0,0 +1,89 @@
...
+
+function mapStateToProps(state: State) {
+    return {
+        time: getCurrentTime(state),
+    };
+}
+
+// const dispatchToPropsMap = {
+//     changeTime,
+// };
```
meganrm, line 84: more code to remove

src/containers/ResultsPanel/selectors.ts
```diff
@@ -144,48 +141,27 @@ const configureData = (
...
-    }
-
-    return data;
+    return isScatterPlot(plot.data) && isTimePlot;
 };
 
 export const getPlotDataConfiguredForPlotly = createSelector(
-    [getPlotData, getCurrentTime],
-    (plotData: RawPlotParams[], currentTime: number): PlotParamsWithKey[] => {
+    [getPlotData],
```
meganrm, line 150: for consistently you could rename this to `getRawPlotParams` or something. (I know you didn't name this here)

src/containers/Plot/index.tsx
meganrm: oh, because of updating the timebar, yeah, I might put a TODO somewhere that says "allow dragging of timebar to update app time" or something so it doesn't just look like forgotten code.

src/containers/ResultsPanel/selectors.ts
schoinh: Oof, I looked into doing this but then it would be inconsistent with `state.metadata.plotData` and all of its references so I'll pass on this for now 😅

toloudis:
Just a note: when talking about optimizing, FPS can be slightly misleading since it's nonlinear (by virtue of being 1/time).   Cutting the frame time in half by going from 1 to 2 FPS (reducing frame time by 500ms)  is a MUCH bigger optimization than cutting the frame time in half to go from 30 to 60 FPS (reducing frame time by 16ms).  Rather than saying "we went from 1 FPS to 2 FPS" you can say "I saved 500ms of computation per frame". 

From the profiler pictures, it looks like you carved off about 100ms per frame, which is a very big dent in getting our frame times down below 100ms (ideally in a perfect realtime world our frame times will be 16ms (60FPS)).

Also, it's all relative to processor speeds on your hardware -- not everyone will see the same absolute performance difference.

toloudis (approved):

schoinh:
> Just a note: when talking about optimizing, FPS can be slightly misleading since it's nonlinear (by virtue of being 1/time). Cutting the frame time in half by going from 1 to 2 FPS (reducing frame time by 500ms) is a MUCH bigger optimization than cutting the frame time in half to go from 30 to 60 FPS (reducing frame time by 16ms). Rather than saying "we went from 1 FPS to 2 FPS" you can say "I saved 500ms of computation per frame".
> 
> From the profiler pictures, it looks like you carved off about 100ms per frame, which is a very big dent in getting our frame times down below 100ms (ideally in a perfect realtime world our frame times will be 16ms (60FPS)).
> 
> Also, it's all relative to processor speeds on your hardware -- not everyone will see the same absolute performance difference.

Thanks! That's very helpful!

src/containers/Plot/index.tsx
schoinh: We need to factor `hasTimeIndicator` into this method, because otherwise this plot component would try to update at every frame even if it doesn't have a time indicator (due to it receiving a dynamic `time` prop from redux). To double check, I just tried out `return this.props.time !== nextProps.time;` (equivalent to your suggestion I think) and we went back to 3 fps with the Actin Bending Angles scatter plot taking a ton of time to update at every frame even though it doesn't have a time indicator. Let me know if I totally missed what you're saying 😆

meganrm: sorry, i meant `return this.props.plotConfig.hasTimeIndicator && this.props.time !== nextProps.time`; that it should explicitly check for the prop that matters updating. Also, I'm curious how this works when you change between trajectories. Does this component end up remounting, or rerendering? If it remounts with new trajectory, you could just add the time plot on component mount, instead of having to check for it every update.

src/containers/Plot/index.tsx
schoinh: Yes thanks!

src/containers/Plot/index.tsx
```diff
@@ -26,11 +28,11 @@ class Plot extends React.Component<PlotProps, {}> {
     }
 
     public render(): JSX.Element | null {
-        const { plotConfig, time } = this.props;
+        const { plotConfig, time, firstFrameTime } = this.props;
         const { shouldRenderTimeIndicator, data, layout } = plotConfig;
         const TIME_INDICATOR_LINE = "timeIndicatorLine";
 
-        if (shouldRenderTimeIndicator && time !== 0) {
+        if (shouldRenderTimeIndicator && time !== firstFrameTime) {
```
meganrm, line 35: So this is keeping the time plot from being rendered when the plots are first rendered and time is at start time, but if you play through, and then get to time 0/startTime again, won't the time plot still be rendered? Is this bad? if it is, the time plots should also be removed from the data when time is reset to start time

schoinh: Oh you're totally right..... That would create a bug. This check `time !== 0` was only for a very subtle cosmetic thing, where it made the y-axis look ever so slightly thicker at t = 0 when the plots first loaded. I think I'm just going to take it out altogether for this PR and if it bugs us we can fix it later? What do you think?

src/containers/Plot/index.tsx
schoinh: Ah, ok! Yeah they should be equivalent because, like you said, `time` is the only thing that changes after initial mount. I just tried changing trajectories under both scenarios and the entire ResultsPanel and its children all remount. I like the suggestion in your last sentence, will implement

schoinh: Uh, not actually sure if ResultsPanel remounts or re-renders, but will check

src/containers/Plot/index.tsx
meganrm: seems fine to me

src/containers/Plot/index.tsx
schoinh: I verified that, when a new trajectory is loaded, ResultsPanel re-renders and the Plot components re-mount. So I moved the initial creation of the time indicator to `componentDidMount` :)

src/containers/Plot/index.tsx
schoinh: I hope the shifting around of code and some improved variable names / comments in my latest push make things more clear and address this comment!

meganrm (approved):

PR #110, opened 2021-01-27
Feature/lazy update of time indicator on plots

schoinh:
Resolves: [Optimize plot rendering to improve endocytosis playback speed](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1257)

### Summary

This is part 2 of the approach outlined in the Jira issue. Part 1 has already been implemented and merged.

Part 1 improved frame rate by having React only re-render plots that have a time indicator line during trajectory playback. This PR goes further to implement lazy updating of the time indicator line. Only the plots that are visible to the user at any given point ("above the fold" might be the correct term) will update during playback, i.e., have the time indicator line move along with the simulation. If at any point the user scrolls to reveal plots that weren't previously visible, the time indicator lines in those plots jump to the correct spot to get in sync with the current time. The jump happens fast and is barely noticeable.

### Performance

Part 1 improved frame rate (max speed went from ~288 to ~178 seconds per frame, or ~3 fps to ~6 fps, on my machine) for endocytosis. With this change (pt 2) I observed an additional improvement (up to ~8 fps of speed, or down to ~130 seconds per frame). 

image

I did notice that the simulation tends to slow down gradually towards the end, i.e., faster playback at t = 3 than at t = 10. I didn't investigate deeply what is causing this, but profiling indicates that it has to do with the viewer -- the runtime for plotly operations remains consistent (~40 ms) throughout the simulation.

### Code changes

I used the [react-waypoint](https://github.com/civiccc/react-waypoint) library, which Megan suggested, to detect when each plot enters/leaves the view. I set top and bottom offsets as 50 px, so at least 50 px of a plot needs to be in the view in order for it to count as "in view."

Thanks for contributing!

### DIFF
--- a/src/containers/Plot/index.tsx
+++ b/src/containers/Plot/index.tsx
@@ -59,21 +59,13 @@ class Plot extends React.Component<PlotProps, PlotState> {
 
     shouldComponentUpdate(nextProps: PlotProps, nextState: PlotState) {
         const { shouldRenderTimeIndicator } = this.props.plotConfig;
-        if (!shouldRenderTimeIndicator) return false;
-
-        if (this.state.isInView) {
-            return true;
-        } else if (!this.state.isInView && nextState.isInView) {
-            return true;
-        } else {
-            return false;
-        }
+        if (shouldRenderTimeIndicator && nextState.isInView) return true;
+        return false;
     }
 
     public render(): JSX.Element | null {
         const { plotConfig, time } = this.props;
         const { shouldRenderTimeIndicator, data, layout } = plotConfig;
-        console.log("rendering Plot:", layout.title);
 
         if (shouldRenderTimeIndicator) {
             // Position the time indicator line at current time (the time indicator should be the last
@@ -86,9 +78,11 @@ class Plot extends React.Component<PlotProps, PlotState> {
 
         return (
             <Waypoint
-                // debug={true}
+                key={layout.title as string}
                 onEnter={() => this.setState({ isInView: true })}
                 onLeave={() => this.setState({ isInView: false })}
+                topOffset={50}
+                bottomOffset={50}
             >
                 <div className={styles.container}>
                     <PlotlyPlot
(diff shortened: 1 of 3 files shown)


src/containers/Plot/index.tsx
```diff
@@ -46,8 +57,10 @@ class Plot extends React.Component<PlotProps, {}> {
         }
     }
 
-    shouldComponentUpdate() {
-        return this.props.plotConfig.shouldRenderTimeIndicator;
+    shouldComponentUpdate(nextProps: PlotProps, nextState: PlotState) {
```
toloudis, line 60: sanity check: that this doesn't prevent plots without time indicators from ever updating, does it??

toloudis: in other words.. if a plot doesn't have a time indicator, does it still get drawn properly and update when it should?

toloudis (approved):

src/containers/Plot/index.tsx
```diff
@@ -46,8 +57,10 @@ class Plot extends React.Component<PlotProps, {}> {
...
+        const { shouldRenderTimeIndicator } = this.props.plotConfig;
+        if (shouldRenderTimeIndicator && nextState.isInView) return true;
+        return false;
```
meganrm, line 63:
```suggestion
        return shouldRenderTimeIndicator && nextState.isInView
```

src/containers/Plot/index.tsx
meganrm: yeah, I had that same reaction to this, but this component re-mounts when there are new plots, so the non time plots get drawn once on mount and never update.

schoinh: What Megan said -- non-time plots have no reason to update after mounting, at least not at this time. I just verified that built-in plotly user interactions (like zooming in) still works properly.

src/containers/Plot/index.tsx
schoinh: Oh yeah, yup 😆

meganrm (approved):

PR #113, opened 2021-03-17
Feature/make frontend compatible with v2 trajectory files (user-specified units)

schoinh:
Resolves: [implement set time unit in simularium-website](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1290)

Also addresses: [Some local uploads break the playback slider](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1248)
(v1 files with non-second units will still break the slider, but the v2 files do not break it)

---

**To recap recent simulariumio changes,** metadata (`TrajectoryFileInfo`) coming into the frontend from the v1 data files looked like this. There were no units specified for time data. All time data were assumed to be in seconds, and it was up to the React frontend to determine the best unit to display and do associated conversions (i.e., if total time was 1e-9, then instead of "0.000000001 s" we would display "1 ns"). Spatial units were specified through `spatialUnitFactorMeters`.:
```
{
    version: number;
    timeStepSize: number;
    totalSteps: number;
    spatialUnitFactorMeters: number;
    size: {
        x: number;
        y: number;
        z: number;
    };
    typeMapping: EncodedTypeMapping;
}
```

Metadata from the v2 files look like this. `spatialUnitFactorMeters` has been swapped out with `spatialUnits` and `timeUnits` so that the user can specify which units should be displayed in the scale bar label and next to the playback slider:
```
{
    version: number;
    timeStepSize: number;
    totalSteps: number;
	spatialUnits: {
        magnitude: number;
        name: string;
    };
    timeUnits: {
        magnitude: number;
        name: string;
    };
    size: {
        x: number;
        y: number;
        z: number;
    };
    typeMapping: EncodedTypeMapping;
}
```

**This PR enables proper loading/playing of v2 trajectory files in addition to continued support for v1 files.** We should drop support for v1 files at some point (see the previously known bug [Some local uploads break the playback slider](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1248)), but seems like it might be good to keep it until we're fully transitioned (see [Slack conversation](https://allencellscience.slack.com/archives/CMQFRJPPF/p1615330010043200)). 

### Summary of code changes
1. `package.json` and `package-lock.json` - Updated simularium-viewer to v2.6.1 which provides simularium-website with the new interfaces `TrajectoryFileInfoV1` and `TrajectoryFileInfoV2`
2. `src/containers/ViewerPanel/index.tsx`
    * Depending on the version of the incoming metadata, construct the label for the scale bar and receive other metadata appropriately
    * Pass down `timeUnits` as a new prop to `PlaybackControls` component (`timeUnits` is `null` for v1)
3. `src/components/PlaybackControls/index.tsx` - Depending on whether `timeUnits` is `null`, determine the times (current and total times) and the time unit to display accordingly.
4. Everything else is making TS happy and Redux work

You'll see it's much easier to work with v2 data than with v1 data. Phew!

### Notes
* Sample v2 data to test with: https://www.dropbox.com/sh/cau4sl0j8a3sfs8/AABQbBgyH414QMPb3Asn9XDIa?dl=0
* I will add a units library for validation of units in non-simulariumio-generated files in a future PR (Jira: [Add units library to frontend](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1366))

Thanks for contributing!


src/components/PlaybackControls/index.tsx
```diff
@@ -33,15 +35,50 @@ const PlayBackControls = ({
...
+                index = 0;
+            }
+
+            setUnitIndex(index);
+            setUnitLabel(units[index]);
+        } else {
+            // v2 data format
+            setUnitLabel(timeUnits.name);
+        }
+    }, [timeUnits, lastFrameTime]);
```
meganrm, line 80: this all (and below) seems like logic that can go into a selector. I think it would be best if components don't have any idea that there is more than one format version

src/containers/ViewerPanel/index.tsx
```diff
@@ -232,32 +238,58 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+                scaleBarLabelNumber =
+                    tickIntervalLength * dataV2.spatialUnits.magnitude;
+                scaleBarLabelUnit = dataV2.spatialUnits.name;
+
+                receiveMetadata({
+                    numFrames: dataV2.totalSteps,
+                    timeStepSize:
+                        dataV2.timeStepSize * dataV2.timeUnits.magnitude,
+                    timeUnits: dataV2.timeUnits,
+                });
```
meganrm, line 281: This again seems like it shouldn't be in a react component, even if it is a container. I think it could be a good idea to have a version manager utility folder (probably in util/) that has functions for conditionally formatting the data based on version number. My worry about having this here is that when a version changes again or we want to remove them, we'll have to be looking around for all the files that mention versions. In this case it would be a `formatTrajectoryTimeInfoByVersion` or something that takes the `data` here returns the `numFrames` `timeStepSize` and `timeUnits`

src/components/PlaybackControls/index.tsx
toloudis: Amen to that last sentence.

src/containers/ViewerPanel/index.tsx
schoinh: That sounds good. I'm thinking I'll add a new file called `versionHandlers.ts` (or something like that) under `util/` and add 2 functions: 
* one that takes in `tickIntervalLength`, `data`, and `version` and returns a `scaleBarLabel`
* one that takes in `data` and `version` and returns `numFrames`, `timeStepSize`, and `timeUnits` like you said.

schoinh: I can also save the version number as metadata to redux here so that I can use that version number in a redux selector to do the things I'm currently doing in `PlaybackControls`, does that sound good? 

And if I hook up PlaybackControls to redux with selectors I guess I'll need to make PlaybackControls a container. Wonder if we can consider doing away with containers/components distinction like Gabe did for the FMS projects a few weeks ago, not on this ticket though.

schoinh: Better idea than what I just said: I'll try to leave PlaybackControl a component (not connected to redux) and handle everything in ViewerPanel/redux, and pass down `roundedTime`, `roundedLastFrameTime`, and `unitLabel` to PlaybackControl as regular props

meganrm: yeah, I definitely dont think this component needs to be connected to redux, what you said sounds good

schoinh:
Changes made since Megan's last review: 
* I Moved all the logic dependent on version numbers from `ViewPanel` and `PlaybackControl` to `src/util/versionHandlers.ts`. The functions in `versionHandlers` are accessed by selectors in `ViewerPanel/selectors.ts` if they require info from the redux store or directly by the `ViewerPanel` container if they don't (like in line 257 of `ViewerPanel/index.tsx` where I call `getScaleBarLabelByVersion`).

As a result:
* The `PlaybackControls` component now simply receives props from `ViewerPanel` and no longer does or invokes any unit-related calculations. 
* Everything dependent on `TrajectoryFileInfo` version number now lives in one file (`versionHandlers.ts`) so it should be easier to add/delete code related to a specific version going forward.

I didn't add/remove any logic, just moved code around.

src/util/versionHandlers.ts
```diff
@@ -0,0 +1,131 @@
...
+    TrajectoryFileInfoV1,
+    TrajectoryFileInfoV2,
+} from "@aics/simularium-viewer/type-declarations/simularium";
+const si = require("si-prefix");
+
+import { TimeUnits } from "../state/metadata/types";
+
+const invalidVersionNumberError =
+    "Invalid version number in TrajectoryFileInfo";
+
```
meganrm, line 17: This is definitely clearer this way, and will be easier to update. I have one more organization idea I want to throw out there, Dan might have additional thoughts too, but after seeing this file and all the switch statements it seems like it would be nice to have all the functions organized by version number, something like this (this could also be a class, but I thought just doing it in objects would be clear): 
```JavaScript
// an object that holds all the functions organized by version number, functions would mostly have the same names, just different logic
const versionMapping = {
    1: {
        function1: (data) => { return processedData }, // pseudo code, just "do something to the data"
        function2: (data) => { return processedData },
    },
    2: {
        function1: (data) => { return processedData },
        function2: (data) => { return processedData },
        function3: (data) => { return processedData },
    },
};

const versionManager = {
    set version(version) { this.version = version},
    functionCheck(functionName) {
       if (!this.version) {
            throw "Need a version number";
        }
        else if (!versionMapping[this.version][functionName]) {
            throw "Function doesn't exist for version"
        }
    }

    get function1() {
        functionCheck(function1);
        versionMapping[this.version].function1;
    }
    get function2() {
        functionCheck(function1);
        versionMapping[this.version].function2;
    }
    get function3() {
        functionCheck(function1);
        versionMapping[this.version].function2;
    }
}
// how you would use it to call a function (`data` is the unprocessed data coming from the viewer):
versionManager.version = data.version;
const processedData = versionManager.function1(data);```

schoinh: Interesting suggestion! I'd like to hear what @toloudis has to say, too. I wonder if the best solution _would_ be to convert all old data to the latest data before doing anything with it like he mentioned before. And if so where that should happen. If we're going to drop support for version 1 soon, all of this seems overkill, but I know we need to think about this for future versions too.

toloudis: I have to catch up and reread this, but my basic thought is:  at the point when data FIRST arrives (in the viewer code), figure out what version it is.  If it's not the CURRENT version, then convert it on the spot and pass it on as current version data.  Then everything downstream only deals with current version, and the version check only exists in one single place, at the very first point where the data is read.  ( If doing that causes excessive expensive copying, then it might be a performance/memory hit but I doubt that it is, for this data here. )

toloudis: To illustrate, this JS library does something like what I'm describing:  https://github.com/TomFrost/Vers
I'm not saying we should use that library (maybe it's good, idk) but it shows the type of pattern I am thinking about

schoinh:
Closing this PR, I'll submit a new one to simularium-viewer that converts v1 data to v2 upon receipt instead.

PR #114, opened 2021-03-31
Feature/load file url

meganrm:
Problem
=======
Users want to be able to share a link to their own trajectory data [AGENTVIZ-1351](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1351)

Solution
========
1. users can link to a simularium file hosted somewhere (like s3) by using a query param in the url: `trajUrl=`
2. If someone provides a url that isn't formatted correctly, the error message will popup and the url will be cleared out
3. If someone provides a url that returns a 404, the error message will popup and the url will be cleared out. 

## Type of change

- [x] This change requires updated or new tests

Change summary:
---------------
* added a new logic for "load via url" that calls the load local file logic
* added checks for url formatting, and additional error reporting
* added tests for passing/not passing urls so we can be clear why something fails
*

Steps to Verify:
----------------
Once you have it running locally:
1. Working example: http://localhost:9001/viewer?trajUrl=https://aics-agentviz-data.s3.us-east-2.amazonaws.com/trajectory/endocytosis.simularium
1. To see malformed url error: http://localhost:9001/viewer?trajUrl=//aics-agentviz-data.s3.us-east-2.amazonaws.com/trajectory/endocytosis.simularium
1. To see 404 error: http://localhost:9001/viewer?trajUrl=https://not-a-real-url.com

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

Thanks for contributing!

### DIFF
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -254,6 +254,7 @@ const loadFileViaUrl = createLogic({
                         simulariumController
                     )
                 );
+                done();
             })
             .catch((error) => {
                 dispatch(
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -174,7 +174,14 @@ const loadLocalFile = createLogic({
     process(deps: ReduxLogicDeps, dispatch, done) {
         const { action, getState } = deps;
         const currentState = getState();
-        const simulariumController = getSimulariumController(currentState);
+        let simulariumController = getSimulariumController(currentState);
+        console.log(action.controller);
+        if (!simulariumController) {
+            if (action.controller) {
+                simulariumController = action.controller;
+                dispatch(setSimulariumController(simulariumController));
+            }
+        }
         const lastSimulariumFile: LocalSimFile = getSimulariumFile(
             currentState
         );
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -178,7 +178,6 @@ const loadLocalFile = createLogic({
         const currentState = getState();
         const simulariumController =
             getSimulariumController(currentState) || action.controller;
-        console.log(simulariumController);
         const lastSimulariumFile: LocalSimFile = getSimulariumFile(
             currentState
         );
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -261,7 +261,7 @@ const loadFileViaUrl = createLogic({
                     setViewerStatus({
                         status: VIEWER_ERROR,
                         errorMessage: error.message,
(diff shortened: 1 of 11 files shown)


src/state/metadata/logics.ts
```diff
@@ -219,9 +222,64 @@ const loadLocalFile = createLogic({
...
+                status: VIEWER_LOADING,
+            })
+        );
+        let simulariumController = getSimulariumController(currentState);
+        if (!simulariumController) {
+            if (action.controller) {
+                simulariumController = action.controller;
+                dispatch(setSimulariumController(simulariumController));
+            }
+        }
```
toloudis, line 241: Can it ever happen that (!simulariumController) and (!action.controller) ?

src/util/index.ts
```diff
@@ -72,3 +72,18 @@ export const wrapText = (
         numLines: numLines,
     };
 };
+
+export const urlCheck = (urlToCheck: any): string => {
+    if (typeof urlToCheck !== "string") {
+        return "";
+    }
+    const regEx = /(https?:\/\/)([\w\-]){0,200}(\.[a-zA-Z][^\-])([\/\w]*)*\/?\??([^\n\r]*)??([^\n\r]*)/g;
```
toloudis, line 80: can you cite where the regex came from?

toloudis (approved):

src/state/metadata/logics.ts
meganrm: this comment: https://github.com/allen-cell-animated/simularium-website/blob/master/src/state/metadata/logics.ts#L4

schoinh: Ah... yeah... I vaguely remember trying to see if removing Mocha would solve the problem but it didn't solve it. So unless it magically fixed itself it is still broken.

schoinh (approved):
This is great! The first time I loaded the linked endocytosis model it took 52 seconds to load. But when I hit play, there was no wait! Makes sense that this is essentially a drag-and-drop.... but I actually kind of liked it better than our current setup for networked files where it loads quickly but takes forever to start playing......... 🤔 Anyway, just a couple small comments

src/state/metadata/logics.ts
```diff
@@ -219,9 +222,64 @@ const loadLocalFile = createLogic({
...
+                    )
+                );
+                done();
+            })
+            .catch((error) => {
+                dispatch(
+                    setViewerStatus({
+                        status: VIEWER_ERROR,
+                        errorMessage: error.message,
+                        htmlData: `${url} failed` || "",
```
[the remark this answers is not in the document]
meganrm: never... mistake from copying other code, thanks for catching

PR #115, opened 2021-04-06
Feature/display user-set time units

schoinh:
Resolves:
* Feature - [implement set time unit in simularium-website](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1290)
* Bug - [Some local uploads break the playback slider](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1248)

### Summary
simularium-viewer v2.7.0 (installed this new version in this PR) converts all data to the latest simularium file version upon receipt. This PR makes simularium-website work with the latest v2 data format.

### Results
I tested with v1 and v2 local uploads. All are as expected:
* v2 files - Times and scale bars display correctly and properly. Trajectories with both non-second and second time units play properly.
image
* v1 files - Times and scale bars display correctly but are "ugly" for some trajectories (because everything displays in seconds and meters). Functionality is same as before - trajectories with very small time values break the slider.
image

I also tested with the 2 streamed trajectories. They are broken, but I believe it's unrelated to this PR (localhost:8080/public/ shows the same bug on simularium-viewer test site when rolled back to the commit before the version update changes were merged). I let @EricJIsaac know in case it's related to array buffer changes.
image

### Remaining related work
* [Determine scale bar label in simularium-viewer](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1381)

Thanks for contributing!

### DIFF
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -38,12 +38,6 @@ const PlayBackControls = ({
     timeUnits,
     isEmpty,
 }: PlayBackProps) => {
-    const [unitLabel, setUnitLabel] = useState("s");
-    useEffect(() => {
-        if (!timeUnits) return;
-        setUnitLabel(timeUnits.name);
-    }, [timeUnits]);
-
     // Where to resume playing if simulation was playing before scrubbing
     const [
         timeToResumeAfterScrubbing,
@@ -173,7 +167,8 @@ const PlayBackControls = ({
                 <p>
                     {roundedTime}{" "}
                     <span className={styles.lastFrameTime}>
-                        / {roundedLastFrameTime} {unitLabel}
+                        / {roundedLastFrameTime}{" "}
+                        {timeUnits ? timeUnits.name : "s"}
                     </span>
                 </p>
             </div>
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -5,6 +5,7 @@ import classNames from "classnames";
 import { TimeUnits } from "../../state/metadata/types";
 import { TOOLTIP_COLOR } from "../../constants/index";
 import Icons from "../Icons";
+import { DisplayTimes } from "../../containers/ViewerPanel/types";
 
 const styles = require("./style.css");
 interface PlayBackProps {
@@ -20,6 +21,7 @@ interface PlayBackProps {
     loading: boolean;
     timeStep: number;
     timeUnits: TimeUnits;
+    displayTimes: DisplayTimes;
     isEmpty: boolean;
 }
 
@@ -36,6 +38,7 @@ const PlayBackControls = ({
     loading,
     timeStep,
     timeUnits,
+    displayTimes,
     isEmpty,
 }: PlayBackProps) => {
     // Where to resume playing if simulation was playing before scrubbing
@@ -70,15 +73,6 @@ const PlayBackControls = ({
             setTimeToResumeAfterScrubbing(-1);
         }
     };
-
-    const roundNumber = (num: number) => parseFloat(Number(num).toPrecision(3));
-    let roundedTime = 0;
-    let roundedLastFrameTime = 0;
-
-    if (timeUnits) {
-        roundedTime = time ? roundNumber(time * timeUnits.magnitude) : 0;
-        roundedLastFrameTime = roundNumber(lastFrameTime * timeUnits.magnitude);
-    }
 
     const btnClassNames = classNames([styles.item, styles.btn]);
 
@@ -165,10 +159,10 @@ const PlayBackControls = ({
             />
             <div className={styles.time}>
                 <p>
-                    {roundedTime}{" "}
+                    {displayTimes.roundedTime}{" "}
                     <span className={styles.lastFrameTime}>
-                        / {roundedLastFrameTime}{" "}
-                        {timeUnits ? timeUnits.name : "s"}
+                        / {displayTimes.roundedLastFrameTime}{" "}
+                        {displayTimes.unitLabel}
                     </span>
                 </p>
             </div>
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { Button, Slider, Tooltip } from "antd";
 import classNames from "classnames";
 
+import { TimeUnits } from "../../state/metadata/types";
 import { TOOLTIP_COLOR } from "../../constants/index";
 import Icons from "../Icons";
 
@@ -18,6 +19,7 @@ interface PlayBackProps {
     onTimeChange: (time: number) => void;
     loading: boolean;
     timeStep: number;
+    timeUnits: TimeUnits;
     isEmpty: boolean;
 }
 
@@ -33,9 +35,15 @@ const PlayBackControls = ({
     onTimeChange,
     loading,
     timeStep,
+    timeUnits,
     isEmpty,
 }: PlayBackProps) => {
-    const [unitIndex, setUnitIndex] = useState(0);
+    const [unitLabel, setUnitLabel] = useState("s");
+    useEffect(() => {
+        if (!timeUnits) return;
+        setUnitLabel(timeUnits.name);
+    }, [timeUnits]);
+
     // Where to resume playing if simulation was playing before scrubbing
     const [
         timeToResumeAfterScrubbing,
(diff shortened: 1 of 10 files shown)


EricJIsaac:
As a note - the yellow bug pre-dates both this change and the array-buffer changes:
https://allencellscience.slack.com/archives/CMQFRJPPF/p1608252745398400?thread_ts=1608248656.391700&cid=CMQFRJPPF

schoinh:
> As a note - the yellow bug pre-dates both this change and the array-buffer changes:
> https://allencellscience.slack.com/archives/CMQFRJPPF/p1608252745398400?thread_ts=1608248656.391700&cid=CMQFRJPPF

Oh that's right, it did happen before. The only thing is that the all-yellow bug only happened sometimes previously, but now the it happens every time reliably for both of the trajectory files. 🤔

src/components/PlaybackControls/index.tsx
```diff
@@ -33,9 +35,15 @@ const PlayBackControls = ({
     onTimeChange,
     loading,
     timeStep,
+    timeUnits,
     isEmpty,
 }: PlayBackProps) => {
-    const [unitIndex, setUnitIndex] = useState(0);
+    const [unitLabel, setUnitLabel] = useState("s");
+    useEffect(() => {
```
meganrm, line 42: I'm not sure why you're setting this in local state here since it's coming from app state. Couldn't you just do `timeUnits.name || "s"` when you're setting the label down below?

src/components/PlaybackControls/index.tsx
```diff
@@ -69,33 +77,14 @@ const PlayBackControls = ({
...
-        }
+    let roundedTime = 0;
+    let roundedLastFrameTime = 0;
 
-        setUnitIndex(index);
-    }, [lastFrameTime]);
+    if (timeUnits) {
+        roundedTime = time ? roundNumber(time * timeUnits.magnitude) : 0;
+        roundedLastFrameTime = roundNumber(lastFrameTime * timeUnits.magnitude);
+    }
```
meganrm, line 87: I'm glad there is so much less logic in this component now, but it seems like this last little bit could be moved to a selector. You could also have the selector default the timeUnits.name to "s".

src/containers/ViewerPanel/index.tsx
```diff
@@ -6,12 +6,14 @@ import SimulariumViewer, {
     SelectionStateInfo,
 } from "@aics/simularium-viewer";
 import "@aics/simularium-viewer/style/style.css";
-import { TrajectoryFileInfo } from "@aics/simularium-viewer/type-declarations/simularium";
+import {
+    TrajectoryFileInfo,
+    TrajectoryFileInfoV2,
+} from "@aics/simularium-viewer/type-declarations/simularium";
```
toloudis, line 12: I would suggest making `TrajectoryFileInfo` always represent the latest version so that you don't need to import two different types here.  Maybe for another PR?

src/components/PlaybackControls/index.tsx
toloudis, line 87: I love how clean this became. looks good.

toloudis (approved):

src/components/PlaybackControls/index.tsx
toloudis: I was wondering the same thing

src/containers/ViewerPanel/index.tsx
```diff
@@ -234,32 +238,24 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
 
     public onTrajectoryFileInfoChanged(data: TrajectoryFileInfo) {
         const { receiveMetadata, simulariumController } = this.props;
-        const { spatialUnitFactorMeters } = data;
-
+        const updatedData = data as TrajectoryFileInfoV2;
```
meganrm, line 241: I'm confused by this typing, with the updated version, shouldn't `onTrajectoryFileInfoChanged` already be typed to the most recent version. Seeing this here also makes me think that type `TrajectoryFileInfo` should be the current version and the only type that the front end ever needs to use, and the viewer can set `TrajectoryFileInfo = TrajectoryFileInfoV2` or whatever is the current version, but it shouldn't be exporting `TrajectoryFileInfoV2`

meganrm:
did you also check this with `?trajUrl=https://aics-agentviz-data.s3.us-east-2.amazonaws.com/trajectory/endocytosis.simularium`? Should be the same as loading it locally, just want to be certain

schoinh:
> did you also check this with `?trajUrl=https://aics-agentviz-data.s3.us-east-2.amazonaws.com/trajectory/endocytosis.simularium`? Should be the same as loading it locally, just want to be certain

I totally did, and it worked great!

src/components/PlaybackControls/index.tsx
schoinh: You're totally right, I removed the local state and now just rendering `timeUnits ? timeUnits.name : "s"`

src/components/PlaybackControls/index.tsx
```diff
@@ -184,7 +167,8 @@ const PlayBackControls = ({
                 <p>
                     {roundedTime}{" "}
                     <span className={styles.lastFrameTime}>
-                        / {roundedLastFrameTime} {units[unitIndex]}
+                        / {roundedLastFrameTime}{" "}
+                        {timeUnits ? timeUnits.name : "s"}
```
toloudis, line 171: when is timeUnits falsy?

schoinh: When no file is loaded! And before TrajectoryFileInfo arrives?

toloudis: default initialize it to {1, "s"} ?  just to remove this conditional :)

toloudis: only if it makes sense.  no serious objection here.

src/containers/ViewerPanel/index.tsx
schoinh: Yeah, I agree with you and @toloudis that there shouldn't be references to `TrajectoryFileInfo` AND `TrajectoryFileInfoV2` in simularium-website. I don't think I can use `TrajectoryFileInfo` here though, because that's defined as `TrajectoryFileInfoV1 | TrajectoryFileInfoV2` in the viewer type definitions. 

Instead I can make a Jira ticket to have [`updateTrajectoryFileInfoFormat` in simularium-viewer](https://github.com/allen-cell-animated/simularium-viewer/blob/master/src/simularium/versionHandlers.ts#L8) output a `TrajectoryFileInfoV2` instead of `TrajectoryFileInfo`. After that, I can modify simularium-website to only use `TrajectoryFileInfoV2`. Yes it has a version number and yes it'll need to be changed in name when we go to V3 but I think that makes sense since it would reflect a change in the shape of the incoming data. What do you guys think of that?

src/containers/ViewerPanel/index.tsx
schoinh: See https://github.com/allen-cell-animated/simularium-website/pull/115#discussion_r609134135

src/containers/ViewerPanel/index.tsx
meganrm: I think the front end should never have to change the import.... the viewer should export something like `CurrentFileInfoFormat` that can be set to whatever the current one is. So in the viewer it would be `export type CurrentFileInfoFormat = TrajectoryFileInfoV2` and you would change that to `export type CurrentFileInfoFormat = TrajectoryFileInfoV3` on the viewer when the time comes

schoinh: Oh I see what you're saying. Can do

src/components/PlaybackControls/index.tsx
meganrm: I think it's a UX question if we should have any time units when there is no data. It could be "-/-" or some other kind of placeholder

schoinh: Hm..... Okay, I'll just leave this here as is for now. This preserves the current behavior which is displaying "0 / 0 s" by default

src/containers/ViewerPanel/index.tsx
schoinh: Made a ticket: https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1382

src/components/PlaybackControls/index.tsx
schoinh: Done, check it out!

src/components/PlaybackControls/index.tsx
schoinh: Actually, I cleaned this up in my latest commits, followed @meganrm 's [suggestion](https://github.com/allen-cell-animated/simularium-website/pull/115#discussion_r609004851) to take care of the unit label in the selector.

EricJIsaac (approved):

meganrm (approved):
Nice job

PR #117, opened 2021-04-20
Update simularium viewer to 2.7.2

EricJIsaac:
Problem
=======
All of the agents are yellow on staging & production. 
The binary broadcast for pc4covid19.simularium is parsed incorrectly.

Solution
========
Update the viewer to 2.7.2, which includes the bug-fixes for both of these

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)


package.json
```diff
@@ -21,6 +21,7 @@
     "test": "BABEL_ENV=commonjs TS_NODE_FILES=true TS_NODE_PROJECT=tsconfig.commonjs.json NODE_ENV=production mocha --opts mocha.opts src/**/*.test.ts"
   },
   "devDependencies": {
+    "@aics/simularium-viewer": "^2.7.2",
```
toloudis, line 24: I think this should be a dependency still, and not a devDependency.   devDependencies are more for things like build tools, transpiler, linter etc

meganrm:
for some reason the `package.json` doesn't seem to be updated? I only see one file changed

meganrm (approved):

schoinh (approved):

PR #127, opened 2021-05-20
Fix/Remove Cloudfront instructions from prod deployment

schoinh:
Doesn't look like it's needed, plus bust-cloudfront-cache.sh doesn't exist on this repo?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,9 +64,6 @@ steps are captured in this project's Jenkinsfile and can be executed by setting 
 Automatically builds from `master`
 
 #### Production deployment
-Note: The final two steps require the AWS CLI as well as permissions on AWS to copy to the production S3 bucket and to run a Cloudfront invalidation.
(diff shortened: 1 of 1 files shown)


schoinh:
> Should we be going the other way and including the cache busting script?

Yeah, or that! 😄 The readme seemed outdated, for example the Jenkins link pointed to the CFE page (I replaced that yesterday with a URL @EricJIsaac gave me), so I wasn't sure if this was an unnecessary artifact of copy/paste or what. @meganrm Can/should we just copy over the cache busting script from the CFE repo?

schoinh:
Just looked at the [CFE script](https://github.com/allen-cell-animated/cell-feature-explorer/blob/master/scripts/bust-cloudfront-cache.sh), we need an ID specific to Simularium, so can't just copy it over from CFE. 🤷🏻‍♀️

meganrm:
I'm working on setting up the deploy through github actions. It will include the cloudfront bust in the script.

PR #128, opened 2021-05-20
Feature/publish gh actions

meganrm:
Problem
=======
Move deployment from Jenkins to github actions 
[Link to story or ticket](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1404)

Solution
========
I copied over jobs from CFE, with one change: CFE just copies the staging build into the production bucket, but since we want a different backend for staging and production, the production script does a fresh build

## Type of change

- [x] This change requires a documentation update

Thanks for contributing!


.github/workflows/auto-deploy-staging.yml
```diff
@@ -0,0 +1,68 @@
+name: Auto deploy staging AWS
+env: 
+  NODE_VERSION: '12.x'
```
toloudis, line 3: I keep forgetting, why are we stuck on node 12?  gh-actions workers have node 14 something auto installed

schoinh:
Probably unrelated, but what's daily-new-viewer.yml in the github workflows folder? Is that in use? If not should it go away?

schoinh (approved):
Looks good to me... This is gonna be great! Thanks!

.github/workflows/auto-deploy-staging.yml
meganrm: It's the version we were using when we started the cfe repo I think, we can update it. For a while we had both 12 and 14. I think it's pretty standard to include 2 so you can feel comfortable going up to the next version.

EricJIsaac (approved):

PR #132, opened 2021-06-09
Fix/enable Dropbox links in URL param

schoinh:
Problem
=======
When a user copies and uses the sharing link for a file in their Dropbox as the `trajUrl` param, the trajectory file doesn't load.

Resolves: [modify dropbox links in URL params to enable download](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1453)

Solution
========
Replace `dropbox.com` in the user-supplied URL param with `dl.dropboxusercontent.com`

(I decided to replace `dropbox.com` instead of `www.dropbox` as specified in the Jira issue just in case, because if for some weird reason the user's URL is missing `www.` then replacing `www.dropbox` won't work, whereas replacing `dropbox.com` works in either case)

Sorry for lumping this together, but I also put in a few extra lines of code to hide the misleading "modified" date tag if it's a URL upload. If someone doesn't like it I'll take it out. I just thought not showing a tag at all would be better than supplying false information (currently, the download time is shown as last modified time for URL uploads).

## Type of change

Steps to Verify:
----------------
2. Go here and check that the trajectory loads successfully: http://localhost:9001/viewer?trajUrl=https://www.dropbox.com/s/juhh7os0ae31ju8/example_default.simularium?dl=0
3. This S3 link should still also work: http://localhost:9001/viewer?trajUrl=https://aics-agentviz-data.s3.us-east-2.amazonaws.com/trajectory/springsalad_condensate_formation_Run17.simularium

### DIFF
--- a/src/components/ViewerTitle/index.tsx
+++ b/src/components/ViewerTitle/index.tsx
@@ -37,7 +37,9 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
 
     let tagText = "";
 
-    // lastModified is a date in the future if file was loaded via external link in URL param
+    // If the trajectory file was loaded from an external link (trajUrl param), we set
+    // lastModified to a date in the future when we received it (in loadFileViaUrl logic)
+    // to indicate that it's not a real "last modified" date and that it shouldn't be displayed as such
     const shouldShowModifiedDate =
         lastModified && lastModified - Date.now() < 0;
     if (shouldShowModifiedDate) {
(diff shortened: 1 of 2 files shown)


meganrm (approved):

src/components/ViewerTitle/index.tsx
```diff
@@ -36,7 +36,11 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
     const version = currentTrajectory ? currentTrajectory.version : "";
 
     let tagText = "";
-    if (lastModified) {
+
+    // lastModified is a date in the future if file was loaded via external link in URL param
```
toloudis, line 40: how does it end up getting a time in the future?

toloudis, line 40: the comment made me think that it happened by external means as opposed to your code doing it intentionally

toloudis (approved):

src/components/ViewerTitle/index.tsx
schoinh: Thanks for pointing it out, I made it more verbose and hopefully less confusing

PR #133, opened 2021-06-10
Feature/make error message more informative upon trajUrl fail

schoinh:
Problem
=======
When the user attempts to load a trajectory with the trajUrl param and it fails, the error message displayed to the user is not informative enough.

Example of message currently displayed when there is a 404 error due to bad URL:
image

Example of message currently displayed when there is a CORS error:
image

Resolves: [Show more verbose error messages to user when files don't load](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1455)

Solution
========
If the URL fetch failed but we still got a response of some kind (i.e., not a CORS error), show the HTTP error code and indicate there was a problem fetching:
image

If we didn't even get a response back (likely because of a CORS error), suggest the user use Dropbox or S3 instead:
image

While I was at it I also dropped any search params (starting with a `?`) from the file name so that the top nav bar after a Dropbox upload would display `myTraj.simularium` instead of `myTraj.simlarium?dl=0`. It's probably not perfect but I think it's an improvement.

## Type of change
I'm not sure which

Steps to Verify:
----------------
2. Going to this URL should display a message suggesting that you use Dropbox or S3 instead: http://localhost:9001/viewer?trajUrl=https://drive.google.com/file/d/1HH5KBpH7QAiwqw-qfm0_tfkTO3XC8afR/view?usp=sharing
3. Going to this URL (not a valid Dropbox link) should display a 404 Failed to fetch message: http://localhost:9001/viewer?trajUrl=https://www.dropbox.com/s/zh6qdn4ey58x55
4. Going to this URL should successfully load a trajectory and show the file name correctly: http://localhost:9001/viewer?trajUrl=https://www.dropbox.com/s/zh6qdn4ey58x55f/nanoparticle_wrapping.simularium?dl=0

### DIFF
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -285,7 +285,7 @@ const loadFileViaUrl = createLogic({
                 // If there was a CORS error, error.message does not contain a status code
                 if (error.message === "Failed to fetch") {
                     errorDetails +=
-                        " Try uploading your trajectory file from a Dropbox or Amazon S3 link instead.";
+                        "<br/><br/>Try uploading your trajectory file from a Dropbox or Amazon S3 link instead.";
                 }
                 dispatch(
                     setViewerStatus({
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -262,7 +262,9 @@ const loadFileViaUrl = createLogic({
                 }
             })
             .then((json) => {
+                // ex) "mysite.com/myTraj.simularium?dl=0" -> ["mysite.com", "myTraj.simularium?dl=0"]
                 const urlSplit = url.split("/");
+                // ex) ["mysite.com", "myTraj.simularium?dl=0"] -> "myTraj.simularium"
                 const name = urlSplit[urlSplit.length - 1].split("?")[0];
                 dispatch(
                     changeToLocalSimulariumFile(
--- a/src/state/metadata/logics.ts
+++ b/src/state/metadata/logics.ts
@@ -253,10 +253,17 @@ const loadFileViaUrl = createLogic({
             }
         }
         fetch(url)
-            .then((data) => data.json())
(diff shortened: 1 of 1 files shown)


src/state/metadata/logics.ts
```diff
@@ -272,11 +281,17 @@ const loadFileViaUrl = createLogic({
                 done();
             })
             .catch((error) => {
+                let errorDetails = `"${url}" failed to load.`;
+                // If there was a CORS error, error.message does not contain a status code
+                if (error.message === "Failed to fetch") {
+                    errorDetails +=
+                        " Try uploading your trajectory file from a Dropbox or Amazon S3 link instead.";
```
meganrm, line 288: I think you can pass html to this error, if I'm right about that, I would put this on a new line, because as it it, I might miss it thinking it just a bunch of error text.

toloudis (approved):
Ok for a quick fix but we still need to address url encoding with respect to nesting a url within another url, the question mark character, and parsing query parameters.  
For example, in a proper url-in-a-url, the inner url's question marks would be `%3F` so as not to pollute the outer url's query parameters.

src/state/metadata/logics.ts
schoinh: Agreed, I just put in 2 line breaks to make it stand out extra. Better?
image

schoinh:
> Ok for a quick fix but we still need to address url encoding with respect to nesting a url within another url, the question mark character, and parsing query parameters.
> For example, in a proper url-in-a-url, the inner url's question marks would be `%3F` so as not to pollute the outer url's query parameters.
> I only raise this point because splitting on `?` makes me nervous!

Agreed, @toloudis would you mind making a ticket for it? (I would but I think you know better what we really need)

meganrm (approved):

PR #134, opened 2021-06-12
Fix/build script

meganrm:
Problem
=======
the builds are not connecting to the backend

Solution
========
add the inline env var, a better fix might be to rewrite the build to use NODE_ENV since that is more standard. I think the env.env might be a gradle holdover. 

## Type of change


toloudis (approved):
I believe in this fix but it's now ambiguous as to whether this or the NODE_ENV is the right one in these two lines. (Or do you really need both)

meganrm:
I don't think you need both currently... but if we use other tooling it will rely on NODE_ENV for deployment settings (heroku for example). So it seems prudent to set it

PR #137, opened 2021-06-17
Feature/set time with text input

schoinh:
Problem
=======
Users can't specify a specific time to skip to.

Resolves: [set time with text input](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1293)
Also resolves: [Slider length changes when displayed times have many digits](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1421)

Solution
========
* Turned the current time display into an interactive number input. 
    * Enter or Tab to trigger time change
    * Esc to cancel entry
* The appropriate width of the input box for a given trajectory is determined and set in the beginning, which also solves the problem of the time display width and slider length changing throughout playback.

## Type of change

- [x] This PR requires updated tests

Steps to Verify:
----------------
2. Load any trajectory and try to specify the time you want to skip to. It should work without breaking any existing functionality 😛 
3. Check other expected behaviors (these were not explicitly specified in the issue so I didn't _try_ to make most of these behaviors happen, but this is how things behave and they seems desirable/acceptable for now):
    * Input box is disabled when the trajectory is loading or playing or if there is no trajectory loaded
    * If you input anything invalid (non-number, a time value outside of the min-max times, etc.) and hit enter, nothing happens
    * If you input a time value that matches a frame time exactly (for example, 2.3 when step size is 0.1) then the viewer should skip to that exact time
    * If you input a time value that's inside the min-max time range but doesn't match a frame time exactly (for example, 2.35 when step size is 0.1) then the viewer will skip to a frame that's closest to that time

Screenshots (optional):
-----------------------
User is interacting:
image
User is not interacting (no change from before):
image

-----------------------
1. src/components/PlaybackControls/index.tsx - See code comments
2. src/containers/ViewerPanel/selectors.ts - I modified `getDisplayTimes` to supply `roundedTimeStep` instead of `unitLabel`. 
    * I'm importing `timeUnits` (containing unit label and magnitude (multiplier)) directly into the `ViewerPanel` container now, so don't need to get the unit label here.
    * On the other hand `rounedTimeStep` is needed in `PlaybackControls` to determine the width of the time input box.
3. src/state/metadata/selectors.ts - New unit tests revealed that the `getTimeStepSize` selector was erroneously getting `metadata.timeStepSize` instead of the correct `metadata.timeStep` which is used elsewhere in the repo like in `metadata.initialState`. So I fixed it.
4. src/containers/ViewerPanel/index.tsx - I realized that we shouldn't multiply the time step by `timeUnits.magnitude` upon receipt, because the multiplication should only happen for display purposes!! So I removed that. It's been inconsequential so far because [almost?] all the magnitudes have been 1.

### DIFF
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { Button, Slider, Tooltip } from "antd";
+import { Button, Slider, Tooltip, InputNumber } from "antd";
 import classNames from "classnames";
 
 import { TOOLTIP_COLOR } from "../../constants/index";
@@ -43,6 +43,7 @@ const PlayBackControls = ({
         timeToResumeAfterScrubbing,
         setTimeToResumeAfterScrubbing,
     ] = useState(-1);
+    const [timeInput, setTimeInput] = useState(firstFrameTime);
 
     // - Gets called once when the user clicks on the slider to skip to a specific time
     // - Gets called multiple times when user is scrubbing (every time the play head
@@ -69,6 +70,18 @@ const PlayBackControls = ({
             playHandler(timeToResumeAfterScrubbing);
             setTimeToResumeAfterScrubbing(-1);
         }
+    };
+
+    const handleTimeInputChange = (
+        timeInput: number | string | undefined
+    ): void => {
+        if (timeInput !== undefined) {
+            setTimeInput(timeInput as number);
+        }
+    };
+
+    const handleTimeInputEnter = (): void => {
+        onTimeChange(timeInput as number);
     };
 
     const btnClassNames = classNames([styles.item, styles.btn]);
@@ -159,13 +172,21 @@ const PlayBackControls = ({
                 disabled={loading || isEmpty}
             />
             <div className={styles.time}>
-                <p>
-                    {displayTimes.roundedTime}{" "}
-                    <span className={styles.lastFrameTime}>
-                        / {displayTimes.roundedLastFrameTime}{" "}
-                        {displayTimes.unitLabel}
-                    </span>
-                </p>
+                <InputNumber
+                    size="small"
+                    step={timeStep}
+                    min={firstFrameTime}
+                    max={lastFrameTime}
+                    value={displayTimes.roundedTime}
+                    onChange={handleTimeInputChange}
+                    onPressEnter={handleTimeInputEnter}
+                    disabled={loading || isEmpty || isPlaying}
+                    // formatter={}
+                />
+                <span className={styles.lastFrameTime}>
+                    / {displayTimes.roundedLastFrameTime}{" "}
+                    {displayTimes.unitLabel}
+                </span>
             </div>
         </div>
     );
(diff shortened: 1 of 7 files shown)


src/components/PlaybackControls/index.tsx
```diff
@@ -159,13 +192,20 @@ const PlayBackControls = ({
...
-                        {displayTimes.unitLabel}
-                    </span>
-                </p>
+                <InputNumber
+                    // key is necessary to re-render this component and override user input
+                    key={displayTimes.roundedTime}
+                    size="small"
+                    value={displayTimes.roundedTime}
+                    onChange={handleTimeInputChange}
+                    onPressEnter={handleTimeInputPressEnter}
```
toloudis, line 201: not sure if this is required, but some users may expect "tab" to also enter the value?  or any defocus event?  or does defocus cancel the text entry?

src/containers/ViewerPanel/test/selectors.test.ts
```diff
@@ -41,33 +42,7 @@ describe("ViewerPanel selectors", () => {
...
-                },
-            };
-
-            const displayTimes = getDisplayTimes(mockState);
-
-            expect(displayTimes).to.deep.equal({
-                roundedTime: 3,
-                roundedLastFrameTime: 15,
-                unitLabel: "ns",
+                roundedTimeStep: 0.1,
```
toloudis, line 45: did we lose a whole test here?

schoinh:
I'm still fixing some things, will let everyone know when this is ready for review again :)

schoinh:
Tests are fixed and this PR is ready for review again. Please see no. 3 and 4 under "Keyfiles" for a description of new changes!

src/containers/ViewerPanel/test/selectors.test.ts
schoinh: Yes. `getDisplayTimes` no longer returns a unit label, so the purpose of this test "getDisplayTimes return user-set non-second time unit" is no longer applicable.

src/components/PlaybackControls/index.tsx
schoinh: Right, good points, I just enabled Tab to commit time, but defocus/ESC is going to require a bit more work (since I have to figure out the best way to clear the text entry and make the component re-render... let me know if you know an obvious way). Ok if we do that on a separate ticket?

schoinh: Another quirk I just found that I'd need to figure out with defocus is, it seems like if I make an `onBlur` handler it doesn't detect clicking away if I click on somewhere on the viewer canvas (clicking on elsewhere in the playback controls, or the left or right side panels, triggers it but not the viewer canvas 🙄 )

schoinh: Actually I just found out that the defocus to clear entry is already automatically implemented without an explicit `onBlur` handler, except it faces the same issue above which is clicking away on the viewer canvas doesn't do anything.

schoinh: Okay, enabled Esc to cancel input!

toloudis (approved):

meganrm (approved):

PR #138, opened 2021-06-18
Feature/add nanoparticle and Smoldyn models to landing page

schoinh:
Problem
=======
We want to stream more models.

Resolves: [Add more example models](https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1465)

Solution
========
* Added the 3 new models to the landing page and dropdown menu
* Also added info about where to find the software used to make the trajectory (`code`), per specs
* While I was at it I also fixed something that's bugging me and probably at least @blairlyons for a while -- now the legalese in tiny gray text is pinned to the bottom of each card.

## Type of change

Steps to Verify:
----------------
1. Pull this branch, `npm start`, and verify you see 5 networked trajectories total on the landing page and in the Load Model dropdown
2. Verify that the models load
    ⚠️ **NOTE** I'm having trouble loading the nanoparticle trajectory for some reason. This should be a separate issue.

2. Also check that the copy looks correct and that all the external links go where they're supposed to

Screenshots (optional):
-----------------------
image

-----------------------
1. src/constants/networked-trajectories.ts
Most others are just styling changes to pin the legalese at the bottom of card.


src/components/ModelCard/index.tsx
```diff
@@ -102,6 +103,10 @@ const ModelCard: React.FunctionComponent<ModelCardProps> = (
                     </p>
                 </a>
                 <p>{description}</p>
+                {code && <p dangerouslySetInnerHTML={{ __html: code }} />}
```
toloudis, line 106: I love names like that

toloudis (approved):

meganrm (approved):

PR #140, opened 2021-06-21
Feature/rotate pan buttons

### LINKED ISSUE #139: implement camera buttons in website
## Use Case

Allow users to switch between pan and rotate, and use key board keys to control the camera in addition to the buttons. 
## Solution

- camera buttons are on bottom right of viewer panel oriented vertically
- text is only in tooltips, not next to camera buttons
- mockup can be found here https://zpl.io/V0oLjdE
- hotkeys (moved from viewer ticket since it's better to implement these in website)
- (SHIFT/CMD modifier keys are already implemented)

**tooltips:

- rotate: "Rotate mode (SHIFT or CMD)"
- pan: "Pan mode (SHIFT or CMD)"
- zoom in: "Zoom in (↑)"
- zoom out: "Zoom out (↓)"
- reset: "Reset camera (R)"
- the first time a user toggles focus without an agent selected, warn them "Select a target agent to focus the camera."
- the rotate/pan and focus buttons' color state updates to match the current state, even if it is toggled or overridden with a key

## Alternatives

I think it would be nice to have hotkeys for the rotate and pan modes, I'll implement this as is, but here is my alternative proposal ( I know we are going to want to user test, so this is just for the future if the current system seems confusing): 

- rotate: "Rotate camera (Shift & click or R)" 
- pan: "Pan camera (Shift & click or P)"
- zoom in: "Zoom in (↑)"
- zoom out: "Zoom out (↓)"
- reset: "Reset camera(H)" (since we're going to use the home icon for this instead of the arrows)

meganrm:
Problem
=======
Need to allow user to switch between rotate and pan modes. 
resolves #139 

Solution
========
added buttons for rotate, pan, and additional hotkeys for zoom in, zoom out, reset. 
Used a font to render the icons so they are most similar to how antd renders svg. 
NOTE: I used hooks and a library to attach the keystroke listeners, and for some reason, using radio buttons for the pan/rotate buttons removed the listener. Ie, clicking one of the radio buttons disabled the hotkey listener so clicking "up" or "shift" had no effect. Clicking anything else on the UI then re-enabled the listener. So I used buttons and classnames to make a radio group. 

## Type of change

Steps to Verify:
----------------
1. npm start
2. shift + click will temporary switch you to pan camera mode
3. click up, down arrows to zoom in and out 
4. click R to reset the view
5. click the pan button to switch to pan mode permanently 
6. hold shift or command and you can temporary rotate the camera again. 

Screenshots (optional):
-----------------------
Screen Shot 2021-06-21 at 11 35 58 AM
Screen Shot 2021-06-21 at 11 36 23 AM
Screen Shot 2021-06-21 at 11 36 42 AM

-----------------------
1. [main file/entry point](https://github.com/allen-cell-animated/simularium-website/compare/feature/rotate-pan-buttons?expand=1#diff-642e7c0b8edb5c03031dbb5e9275f522e0d0ef25bf63e0d38fe5691de3842aab)

Thanks for contributing!


src/components/CameraControls/index.tsx
```diff
@@ -1,26 +1,152 @@
...
+    setPanningMode: (value: boolean) => void;
 }
 
 const CameraControls = ({
     resetCamera,
     zoomIn,
     zoomOut,
+    setPanningMode,
 }: CameraControlsProps) => {
+    const [mode, setMode] = useState("rotate");
```
toloudis, line 32: should be `ROTATE` ?

src/components/CameraControls/index.tsx
```diff
@@ -1,26 +1,152 @@
...
     zoomIn,
     zoomOut,
+    setPanningMode,
 }: CameraControlsProps) => {
+    const [mode, setMode] = useState(ROTATE);
+    const [keyPressed, setKeyPressed] = useState("");
+    const lastKeyPressed = useRef("");
+    const isModifierKey = (key: string) =>
+        CAMERA_MODE_MODIFIER_KEYS.includes(key);
+    useHotkeys(
```
schoinh, line 37: [nit] A blank line between lines 36 and 37 would be good; I got confused for a bit thinking `useHotkeys` was a part of `isModifierKey` and that we were missing brackets and indentation 😅

src/components/CameraControls/index.tsx
```diff
@@ -1,26 +1,152 @@
...
+        }
+    }, [mode]);
+
+    useEffect(() => {
+        if (
+            (isModifierKey(keyPressed) && !lastKeyPressed.current) ||
+            (isModifierKey(lastKeyPressed.current) && !keyPressed)
+        ) {
+            // toggle the mode, so the radio buttons reflect the true state
+            lastKeyPressed.current = keyPressed;
```
schoinh, line 75: Is this line necessary (since same thing is in 91)?

src/components/CameraControls/index.tsx
```diff
@@ -1,26 +1,152 @@
...
+                break;
+            default:
+                break;
+        }
+        lastKeyPressed.current = keyPressed;
+    }, [keyPressed]);
+
     return (
         <div className={styles.container}>
+            <div className={styles.moveButtons}>
```
schoinh, line 96: Is this div necessary? (I don't see a `move-buttons` selector in the CSS)

src/styles/ant-vars.less
```diff
@@ -137,11 +137,14 @@ https://github.com/ant-design/ant-design/blob/master/components/style/themes/def
...
-@radio-button-active-color : #fff;
-
+@radio-button-bg: @black;
+@radio-button-checked-bg: @dark-three;
+@radio-button-color: @greyish-brown;
+@radio-button-hover-color: @primary-5;
+@radio-button-active-color: @dark-three;
+@radio-disabled-button-checked-bg: tint(@black, 90%);
+@radio-disabled-button-checked-color: @disabled-color;
+@radio-wrapper-margin-right: 8px;
```
schoinh, line 147: Did you mean to keep these changes after removing the antd radio group? I guess it doesn't hurt in case we want them for future 🤷🏻‍♀️

schoinh (commented):
This is exciting! One thing I noticed while testing it out was that after panning, the up and down arrow keys for zooming don't work properly anymore. (Scroll mouse for zooming works fine.) Seems like a separate issue to file?

src/components/CameraControls/index.tsx
meganrm: I use it in the next PR to group the move buttons together

src/styles/ant-vars.less
meganrm: yeah, I figured we'd probably use the same styling for any radio buttons

meganrm:
> This is exciting! One thing I noticed while testing it out was that after panning, the up and down arrow keys for zooming don't work properly anymore. (Scroll mouse for zooming works fine.) Seems like a separate issue to file?

Yeah, to make sure we're seeing the same behavior: for up, the camera still moves, but seems to step rotate and down does nothing. I see it both clicking the button and using the arrow keys which makes me thing the keys are successfully calling the function but something is weird with the camera on the viewer.

src/components/CameraControls/index.tsx
schoinh: Oh right I can read code

schoinh (approved):

meganrm:
> > This is exciting! One thing I noticed while testing it out was that after panning, the up and down arrow keys for zooming don't work properly anymore. (Scroll mouse for zooming works fine.) Seems like a separate issue to file?
> 
> Yeah, to make sure we're seeing the same behavior: for up, the camera still moves, but seems to step rotate and down does nothing. I see it both clicking the button and using the arrow keys which makes me thing the keys are successfully calling the function but something is weird with the camera on the viewer.

I verified that this happens in the example app on the viewer too and made a [ticket](https://github.com/allen-cell-animated/simularium-viewer/issues/125)

toloudis (approved):
One thing I noticed is that holding down shift still let me click to select an agent.  It would have felt more natural to me to disable that selection when shift is held, but not a major deal.

I wonder if this conflicts with the internal shift key handling down in the OrbitControls in the viewer.

PR #143, opened 2021-06-21
Feature/focus button

### LINKED ISSUE #142: Implement focus toggle
## Use Case

Allow users to toggle on and off the camera follow mode, while maintaining a selected agent

## Solution

A button that toggles between focus mode and non focus mode. for focus, pressing a hotkey will permanently set the state of the toggle, even when the key is released
the first time a user toggles focus without an agent selected, warn them "Select a target agent to focus the camera."

meganrm:
Problem
=======
resolves #142 

Solution
========
added a focus button that toggles state, and a hotkey 

## Type of change

Change summary:
---------------
* Updated the font (added instructions on how I do this to the [wiki](https://github.com/allen-cell-animated/simularium-website/wiki/Generate-a-font-with-svgs))

Steps to Verify:
----------------
1. npm start
2. select an agent
3. try to pan, won't allow you to really,  *NOTE: should we give the user some info here the first time they try this, ie, "turn off focus to pan away from selected agent?"
4. push focus button or "f"
5. try to pan, will be able to, and agent stays selected

Screenshots (optional):
-----------------------
Screen Shot 2021-06-21 at 1 12 00 PM
Screen Shot 2021-06-21 at 1 12 13 PM

-----------------------
1. [main file](https://github.com/allen-cell-animated/simularium-website/compare/feature/focus-button?expand=1#diff-642e7c0b8edb5c03031dbb5e9275f522e0d0ef25bf63e0d38fe5691de3842aab)

Thanks for contributing!


EricJIsaac (approved):

blairlyons (approved):
looks good, works 👍 

For the conflict between pan/focus, I think we can leave it as is for now and user test later. I had thought about making panning past a certain threshold toggle focus off automatically, but it would be good to have a UX designer weigh in first.

PR #162, opened 2021-07-01
PR template - use bullets instead of checkboxes under "Type of change"

schoinh:
Checkboxes in PRs make GitHub interpret them as tasks to be completed and displays a distracting blue circle, so using plain bullets instead. If approved I will go ahead and commit this change to simularium-viewer as well.


meganrm (approved):

toloudis (approved):
Could you change this in allen-cell-animated/github-boilerplate as well?

PR #173, opened 2021-07-02
Fix/excessively wide time input box

### LINKED ISSUE #158: Time input box is excessively wide for Membrane Wrapping a Nanoparticle
## Description

The time input box next to the playback slider seems excessively wide for the trajectory "Membrane Wrapping a Nanoparticle."

image

## Expected Behavior

The box should have a width close to the width of the trajectory's final frame time + time step size.

## Reproduction

1. On staging, load Membrane Wrapping a Nanoparticle
2. See that the time input box is much wider than it needs to be

schoinh:
Problem
=======
Resolves #158 

Solution
========
Two fixes needed to be made to address the above bug with Membrane Wrapping a Nanoparticle:
1. Round the `refTimeValue` (`lastFrameTime + timeStep`) before determining its character length, because 5.44 + 0.08 = 5.520000000212455 of course
2. If the firstFrameTime is a very small (and long) number like 0.0008 (longer than `refTimeValue`), the input box needs to be long enough to accommodate that

## Type of change

Steps to Verify:
----------------
1. `npm start` and load Membrane Wrapping a Nanoparticle
2. The time input box should have sufficient (but not excessive) width to display the current time value at all times
3. Check the other networked trajectories

Screenshots (optional):
-----------------------
Before:
image

After:
image

-----------------------
* src/containers/ViewerPanel/selectors.ts - I moved the logic in `getTimeInputWidth` from PlaybackControls into here (as `getMaxNumChars`), and made it a function that the `getDisplayTimes` selector calls. `getMaxNumChars` now has the fixes described under "Solution".
* src/util/index.ts - I moved the `roundNumber` function from the `getDisplayTimes` selector into here and made it a util function that's called whenever we need to round a time value for display.

### DIFF
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -104,7 +104,7 @@ const PlayBackControls = ({
     const getTimeInputWidth = (): string => {
         // If maxNumChars is 5 then the input box width will be 6 character widths long
         // (+ 1 is arbitrary padding)
-        return displayTimes.maxNumChars + 1 + "ch";
+        return `${displayTimes.maxNumChars + 1}ch`;
     };
 
     const btnClassNames = classNames([styles.item, styles.btn]);
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -107,8 +107,18 @@ const PlayBackControls = ({
         // the max number of characters for this trajectory
         const refTimeValue =
             displayTimes.roundedLastFrameTime + displayTimes.roundedTimeStep;
-        const roundedRefTime = roundTimeForDisplay(refTimeValue);
-        const maxNumChars = roundedRefTime.toString().length;
+        const roundedRefTime = roundTimeForDisplay(refTimeValue).toString();
+
+        // Edge case: firstFrameTime is a very small but long number like this 0.000008,
+        // so we need to accommodate that.
+        const roundedFirstFrameTime = roundTimeForDisplay(
+            firstFrameTime
+        ).toString();
+        const maxNumChars = Math.max(
+            roundedFirstFrameTime.length,
+            roundedRefTime.length
+        );
+
         // If maxNumChars is 5 then the input box width will be 6 character widths long
         // (+ 1 is arbitrary padding)
         return maxNumChars + 1 + "ch";
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -6,7 +6,6 @@ import { TOOLTIP_COLOR } from "../../constants/index";
 import Icons from "../Icons";
 import { DisplayTimes } from "../../containers/ViewerPanel/types";
 import { TimeUnits } from "../../state/metadata/types";
-import { roundTimeForDisplay } from "../../util";
 
 const styles = require("./style.css");
 interface PlayBackProps {
@@ -101,7 +100,7 @@ const PlayBackControls = ({
         }
     };
 
-    // Determine the width of the input box
+    // Determine the width of the input box based on a likely max number of characters
     const getTimeInputWidth = (): string => {
         // If maxNumChars is 5 then the input box width will be 6 character widths long
         // (+ 1 is arbitrary padding)
--- a/src/components/PlaybackControls/index.tsx
+++ b/src/components/PlaybackControls/index.tsx
@@ -6,6 +6,7 @@ import { TOOLTIP_COLOR } from "../../constants/index";
 import Icons from "../Icons";
 import { DisplayTimes } from "../../containers/ViewerPanel/types";
 import { TimeUnits } from "../../state/metadata/types";
+import { roundTimeForDisplay } from "../../util";
 
 const styles = require("./style.css");
 interface PlayBackProps {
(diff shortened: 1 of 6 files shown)


src/containers/ViewerPanel/selectors.ts
```diff
@@ -52,32 +54,73 @@ export const convertUIDataToColorMap = (
     }, returnData);
 };
 
+// Determine the likely max number of characters for the time input box
+const getMaxNumChars = (
```
toloudis, line 58: i would unit test this function all by itself

toloudis (approved):
This looks great to me.  How about a simpler test that getMaxNumChars() returns the expected thing for a variety of numbers?

src/components/PlaybackControls/index.tsx
```diff
@@ -100,16 +100,11 @@ const PlayBackControls = ({
...
     const getTimeInputWidth = (): string => {
-        // If lastFrameTime is 15 and step size is 0.25 then 15.25 is probably going to have
-        // the max number of characters for this trajectory
-        const refTimeValue =
-            displayTimes.roundedLastFrameTime + displayTimes.roundedTimeStep;
-        const maxNumChars = refTimeValue.toString().length;
         // If maxNumChars is 5 then the input box width will be 6 character widths long
         // (+ 1 is arbitrary padding)
-        return maxNumChars + 1 + "ch";
+        return displayTimes.maxNumChars + 1 + "ch";
```
toloudis, line 107: this might be less ambiguous as a formatted string `${(displayTimes.maxNumChars+1)}ch` or maybe just putting parens around the numbers...

toloudis (approved):

src/containers/ViewerPanel/test/selectors.test.ts
```diff
@@ -70,7 +74,95 @@ describe("ViewerPanel selectors", () => {
...
+                },
+            };
+
+            const displayTimes = getDisplayTimes(mockState);
+
+            expect(displayTimes).to.deep.equal({
+                roundedTime: 3,
+                roundedLastFrameTime: 15,
+                roundedTimeStep: 0.08,
+                maxNumChars: 8,
```
meganrm, line 106: i'm not sure i understand why this is 8 since the rounded times are all to 2-3 sig figures.

EricJIsaac (approved):

src/containers/ViewerPanel/test/selectors.test.ts
```diff
@@ -70,7 +74,95 @@ describe("ViewerPanel selectors", () => {
...
+            });
+        });
+    });
+
+    describe("getMaxNumChars", () => {
+        it("determines correct maxNumChars when time values are simple integers", () => {
```
meganrm, line 112: nit but I think it would be helpful to define what is "correct maxNum" here. Like, "if time values are simple integers, returns a max based on the largest value time can possibly be"

src/containers/ViewerPanel/test/selectors.test.ts
```diff
@@ -70,7 +74,95 @@ describe("ViewerPanel selectors", () => {
...
+            const lastFrameTime = 44.1;
+            const timeStep = 0.8;
+
+            const maxNumChars = getMaxNumChars(
+                firstFrameTime,
+                lastFrameTime,
+                timeStep
+            );
+
+            expect(maxNumChars).to.equal(4); // "44.9".length()
```
meganrm, line 137: you could just put `expect(maxNumChars).to.equal("44.9".length())

src/util/index.ts
```diff
@@ -76,3 +76,7 @@ export const wrapText = (
 export const clearUrlParams = () => {
     history.replaceState({}, "", `${location.origin}${location.pathname}`);
 };
+
+export const roundTimeForDisplay = (time: number): number => {
+    return parseFloat(time.toPrecision(3));
+};
```
meganrm, line 82: i think we said in the test meeting that we want to cover utils too. It's a super simple function to test, but could be important incase we ever change the precision expectations

src/containers/ViewerPanel/test/selectors.test.ts
schoinh: Ah, yeah it's because `firstFrameTime` is 0.000004. I guess I could do this instead like you suggested in another comment:
```
            expect(displayTimes).to.deep.equal({
                roundedTime: 3,
                roundedLastFrameTime: 15,
                roundedTimeStep: 0.08,
                maxNumChars: "0.000004".length(),
            })
```

src/containers/ViewerPanel/test/selectors.test.ts
schoinh: Yesss

src/containers/ViewerPanel/test/selectors.test.ts
meganrm: ah, i also can't quite count :)

src/containers/ViewerPanel/test/selectors.test.ts
schoinh: Good call, I made them better

src/containers/ViewerPanel/test/selectors.test.ts
schoinh: Done!

src/util/index.ts
schoinh: Done!

meganrm (approved):

PR #186, opened 2021-07-13
Update markdown files

### LINKED ISSUE #185: Update markdown files
## What needs to happen?
* A lot of the readme-type files (all the markdown files) contain outdated info. Go through them and remove/update any incorrect/outdated information. 
* The documentation doesn't need to be super comprehensive yet, but it should be free of incorrect/misleading info.
* The highest priority is README.md which is shown on the repo's landing page

## Why should we do this?
We want our repo to be up-to-date and well-maintained both for ourselves and for anybody who visits.

Ideally for the paper1 submission

schoinh:
Problem
=======
Closes #185 

Solution
========
I went through all the markdown files and looked for any outdated info. There is still a lot more documentation we can / should add, but in this PR I just tried to get any outdated info updated. I wrote some questions I have as comments in the files themselves; I'd like to get them resolved before merging.

I also deleted src/data because there's only one json file in it and I don't think we use that anymore.

## Type of change


CONTRIBUTING.md
```diff
@@ -55,6 +58,8 @@ ___
 
 
 ### Deployment
+TODO: Update this section
+
 Once built, Webpack outputs (e.g., index.html, JS and CSS files) are put into a tar archive, gzipped, and stored in
 Artifactory in the `maven-snapshot-local` repo. From there, deployments involve: a) pulling a particular artifact (referenced by git tag) out of Artifactory
 and b) copying the contents of the artifact to an S3 website bucket. For both staging and production deployments, these
```
schoinh, line 65: @meganrm How much of this is still valid?

meganrm: it can all be deleted

meganrm (approved):

SUPPORT.md
```diff
@@ -1,5 +0,0 @@
-# Support
```
toloudis, line 1: my only hesitation here is removing this file entirely.  This came from the github-boilerplate, I believe, which was informed by some standards org's suggested github files.   Would we remove this from the boilerplate and all repos?

meganrm: oh yeah, shouldn't this just be updated with the link to the simularium forum?

schoinh: I moved the contents of this file to README.md because I thought this info didn't warrant an entire separate file, and I thought README.md should have that info anyway. But I don't mind if we keep this file for completeness' sake if that's what we want.

toloudis: I guess I would feel fine with removing it in a project that thought it wasn't necessary but I would leave it in the github-boilerplate

schoinh: Okay, then I will leave this PR as is :) And I won't do anything with the boilerplate template.

toloudis (approved):

PR #187, opened 2021-07-13
Fix/remove Gradle, Jenkins, Pipfile, and Docker

### LINKED ISSUE #183: Remove Gradle and Jenkins
## What needs to happen?
We should remove any folders/files related to Gradle and Jenkins from this repo.

## Why should we do this?
We no longer use Gradle or Jenkins to build/deploy. We use npm and GitHub Actions now. Having the old files and folders doesn't hurt anything but is completely unnecessary and a little confusing.

As part of the repository cleanup for the first paper?

schoinh:
Problem
=======
Resolves #183 

Solution
========
* I'm positive we can get rid of Gradle and Jenkins files. So deleted those.
* I'm not sure about Pipfile and Docker (because I'm not sure what they did in the first place). But I deleted those too. Should we keep these?

Relevant documentation update is in #186


Pipfile
```diff
@@ -1,11 +0,0 @@
...
-name = "pypi"
-url = "https://pypi.org/simple"
-verify_ssl = true
-
-[dev-packages]
-
-[packages]
-
-[requires]
-python_version = "3.7"
```
meganrm, line 11: yeah this is fine to get rid of

meganrm (approved):

toloudis (approved):
i love deleting code

Issue #174, opened 2021-07-06
Post example models for SpringSaLaD and MEDYAN


blairlyons:
## Use Case

It would be nice for the posted examples to include models from SpringSaLaD, MEDYAN, and Vivarium, while still keeping the "Load your own data" card near the top of the page.

## Solution

Add cards for the new models and routing for their links to the website. Data for the models is highlighted on this sheet https://docs.google.com/spreadsheets/d/1QmcCY6qu6oRQH-qzj_TMjz0CHeHMa32evP09KTBsV0w/edit?usp=sharing and thumbnails are in drive folder https://drive.google.com/drive/folders/1o_MTacfBgRCvRdq1NuAXE2OuRqEXdB82?usp=sharing

Adding these models will push the "Load your own data" card far down the page. It could be ordered so that it's always the third card to prevent that.

schoinh:
Will add Vivarium later when backend bug is fixed.

PR #190, opened 2021-07-13
Feature/post example models for SpringSaLaD and MEDYAN

schoinh:
Problem
=======
Resolves: #174 

Solution
========
Added the new models and modified the `LandingPage` component to always place the blank model card as the 3rd card per specs.

## Type of change

Screenshots (optional):
-----------------------
Screen Shot 2021-07-13 at 11 33 32 AM
Screen Shot 2021-07-13 at 11 34 09 AM
Screen Shot 2021-07-13 at 11 34 28 AM


src/constants/networked-trajectories.ts
```diff
@@ -123,6 +135,136 @@ const TRAJECTORIES: TrajectoryDisplayData[] = [
...
+            journal: "PLoS Computational Biology",
+            year: 2019,
+            url:
+                "https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1007156",
+        },
+        description:
+            "A MEDYAN model of interacting actin filaments, α–actinin crosslinkers, and myosin motors. High myosin activity causes formation of an aster-like structure.",
+        code:
+            "Software used to generate this data is available <a href='https://bitbucket.org/jkomianos/medyan/src/master/'>here</a>. The outputs that were visualized can be downloaded <a href='https://drum.lib.umd.edu/handle/1903/21856'>here</a>.",
+        legalese:
```
meganrm, line 181: this seems like something that can be factored out since it's repeated

schoinh: Okay, I made a function returning a template literal for the legalese. I didn't do anything with the `code` text because that one varies slightly, and I didn't want to make it too complicated especially if we're going to make this all part of the trajectory file metadata soon ;)

src/constants/networked-trajectories.ts
```diff
@@ -123,6 +135,136 @@ const TRAJECTORIES: TrajectoryDisplayData[] = [
...
+            "Please note that the use of this model is subject to third party licensing requirements, which are currently available <a href='https://bitbucket.org/jkomianos/medyan/src/master/license.txt'>here</a>.",
```
blairlyons: From talking to the lawyers this recently changed to "Please note that the use of this **software** is subject to third party licensing requirements, which are currently available here." It was a subtle change, and I missed it last time.

schoinh: Good thing I turned this text into a variable per @meganrm 's suggestion above...... Fixed!

src/components/LandingPage/index.tsx
```diff
@@ -50,7 +50,7 @@ const LandingPage: React.FunctionComponent<{}> = () => {
                         related plots.
                     </p>
                     <div className={styles.cards}>
-                        {TRAJECTORIES.map((trajectory) => {
+                        {TRAJECTORIES.slice(0, 2).map((trajectory) => {
```
meganrm, line 53: probably overkill, but it might be a good idea to make a variable named "NUMBER_OF_CARDS_PER_ROW" and make this NUMBER_OF_CARDS_PER_ROW-1, just so if we ever want to change the layout we know what value to change.

meganrm (approved):

toloudis (approved):

Issue #166, opened 2021-07-01
Reorganize redux state


schoinh:
## Use Case

For a Simularium developer, it would be nice to have redux state that is organized in a way that makes sense.

## Solution

Reorganize state to (per @meganrm):

1. simFileMetaData/modelMetaData or something
would hold all the data that comes from the back end when you select a model
2. status
would hold stuff like isBuffering, isPlaying, viewerStatus
3. selection
the same as it is, just move "draggedOverViewer" to status
4. simularium
right now would just have the controller, but in the future might have all the available trajectories

Also create a base state to revert to for all the state resetting tasks, like this https://medium.com/@justintulk/best-practices-for-resetting-an-es6-react-components-state-81c0c86df98d

See discussion: https://github.com/allen-cell-animated/simularium-website/pull/88#discussion_r541460619

schoinh:
Current state branches (`initialState`):
```javascript
metadata:
    firstFrameTime: 0,
    lastFrameTime: 0,
    timeStep: 0,
    timeUnits: null,
    agentIds: [],
    agentUiNames: [],
    plotData: [],
    simulariumFile: {
        name: "",
        data: null,
        lastModified: null,
    },
    simulariumController: null,
    viewerStatus: VIEWER_EMPTY,
    viewerError: null,

selection:
    time: 0,
    numberPanelsCollapsed: 0,
    visibleAgentKeys: {},
    highlightedAgentKeys: {},
    draggedOverViewer: false,
    agentColors: {},
    isBuffering: false,
    isPlaying: false,
```

Proposed state branches:
```javascript
metadata:   
// trajectory data received from backend
    firstFrameTime: 0,
    lastFrameTime: 0,
    timeStep: 0,
    timeUnits: null,
    agentIds: [],
    agentUiNames: [],
    plotData: [],
    simulariumFile: {
        name: "",
        data: null,
        lastModified: null,
    },

selection:    
// selections by user
    time: 0,
    numberPanelsCollapsed: 0,
    visibleAgentKeys: {},
    highlightedAgentKeys: {},
    agentColors: {},

status: // or app?
// app status
    viewerStatus: VIEWER_EMPTY,
    viewerError: null,
    draggedOverViewer: false,
    isBuffering: false,
    isPlaying: false,

simularium:
    simulariumController: null,
```

toloudis:
Would it make sense to just stash the whole TrajectoryFileInfo as read-only state?

meganrm:
[...]

Issue #182, opened 2021-07-10
Get 100% unit test coverage on redux selectors


schoinh:
## What needs to happen?
Check if all the selector functions have test coverage. If not, write more tests.

## Why should we do this?
We rely on selectors for most of the data handling/manipulation in this repo, so they need to be rock solid.

## When does this need to get done?
Ideally in the next month or so? Would be easier with Jest and its coverage reporting in place

meganrm:
A thing I just realized that I should have thought of before, but the basic selectors are a lot of these files and we don't want to test them. One solution if we really want to enforce 100% is to break the files up into basic selectors and composed selectors and enforce 100% on composed. we could structure it so that everything still gets exported from `selectors/index` so we wouldn't have to change the import statements in the other files. 
```
selectors/
     index <-- exports composed and basic
     basic
     composed```

meganrm:
Alternatively, jest does as an ignore next: `/* istanbul ignore next */` but you would have to add it before each basic selector. They don't have a "open ignore" "close ignore" as far as I can tell.

PR #225, opened 2021-07-29
Reorganize Redux state

schoinh:
Problem
=======
Accomplishes most of #166

(did everything except renaming the following states in the `viewer` state branch, I'll do that in a separate PR in a feeble attempt to make this PR less huge)
```
viewerStatus -> status,
viewerError -> error,
draggedOverViewer -> draggedOver,
```

Solution
========
There is no new or deleted code here. Just moving code and renaming variables.

* Renamed `metadata` branch to `trajectory`
* Added `viewer` and `simularium` branches
* Moved state slices(?) around as laid out in the issue #166 
* Lots of find-and-replace to make the app work with the reorganized state
* Added a short description of each state branch in the readme

## Why are we calling these "branches"?
While I can't find the specific use of the word "branch" in the Redux docs, the Redux docs do [describe a state tree](https://redux.js.org/api/store). But from there they seem to veer from the tree analogy and use the term "slice" instead of "branch." (See https://redux.js.org/usage/structuring-reducers/basic-reducer-structure#basic-state-shape)

It seems like the "branch" terminology was used in our repo from the get-go because that was what was in our template's code and readme. I think "branch" makes sense with the tree concept and think it's fine to keep them. 🤷🏻‍♀️ 

## Type of change

* Maintenance

Steps to Verify:
----------------
1. Check that all tests pass
2. Check there are no type errors
3. Run `npm start` and make sure that basic functions work as before, such as loading a networked trajectory, loading a trajectory by drag and drop, loading by trajUrl, and playback.

-----------------------
1. All the files under `src/state`
2. src/state/README.md - please provide input


src/state/README.md
```diff
@@ -91,3 +50,53 @@ export default {
...
 ```
+
+### State branches
+
+The Redux state tree in simularium-website is currently divided into 4 branches:
+
+* selection - selections made by the user through the UI, such as `time` and `numberPanelsCollapsed`
+* simularium - the Simularium controller
+* trajectory - current trajectory data, such as `timeStep` and `simulariumFile`
+* viewer - viewer status, such as `isBuffering` or `viewerError`
```
schoinh, line 61: I just added this little section.

schoinh:
I just temporarily removed the `src/state/*/selectors.ts` test coverage threshold from the jest config so the CI would pass. There are new selector files with only basic selectors that don't need testing which is making this fail. We need to deal with this in some way in a future PR, see some of @meganrm 's suggestions [here](https://github.com/allen-cell-animated/simularium-website/issues/182#issuecomment-887725994).

src/state/trajectory/actions.ts
```diff
@@ -4,32 +4,27 @@ import {
...
-    SetViewerStatusAction,
     RequestNetworkFileAction,
     RequestLocalFileAction,
     ClearSimFileDataAction,
     LoadViaUrlAction,
 } from "./types";
 import { SimulariumController } from "@aics/simularium-viewer/type-declarations";
 
-export function receiveMetadata(payload: MetadataStateBranch): ReceiveAction {
+export function receiveMetadata(payload: TrajectoryStateBranch): ReceiveAction {
```
meganrm, line 27: is this being used? should probably be named something more specific

schoinh: Yeah, `receiveMetadata` is used. Would `receiveTrajectoryMetadata` be better? I might do that in a separate PR though.

src/containers/ViewerPanel/index.tsx
```diff
@@ -15,24 +15,30 @@ import Bowser from "bowser";
...
 import { State } from "../../state/types";
 import selectionStateBranch from "../../state/selection";
-import metadataStateBranch from "../../state/metadata";
-import { VIEWER_EMPTY, VIEWER_SUCCESS } from "../../state/metadata/constants";
+import trajectoryStateBranch from "../../state/trajectory";
+import viewerStateBranch from "../../state/viewer";
+import {
+    VIEWER_EMPTY,
+    VIEWER_SUCCESS,
+    VIEWER_ERROR,
```
toloudis, line 23: was ERROR added intentionally?  didn't see it referenced yet

schoinh: It was moved up from line 35 in the "before" version :)

src/state/configure-store.ts
```diff
@@ -4,14 +4,29 @@ import { createLogicMiddleware } from "redux-logic";
...
+    simularium,
+    State,
+} from "./";
 
 const reducers = {
-    metadata: metadata.reducer,
+    trajectory: trajectory.reducer,
     selection: selection.reducer,
+    viewer: viewer.reducer,
+    simularium: simularium.reducer,
```
toloudis, line 21: are these last two actually new?  just wondering why they were absent before and added now

src/state/configure-store.ts
```diff
@@ -4,14 +4,29 @@ import { createLogicMiddleware } from "redux-logic";
...
 };
 
-const logics = [...metadata.logics, ...selection.logics];
+const logics = [
+    ...trajectory.logics,
+    ...selection.logics,
+    ...viewer.logics,
+    ...simularium.logics,
+];
```
toloudis, line 29: same as above comment.  do these lists need to always be complete?

src/state/configure-store.ts
schoinh: Oh oops, yes this IS actually new! 😅 Previously there were only 2 branches (`metadata` and `selection`). Now we have 4. See https://github.com/allen-cell-animated/simularium-website/issues/166#issuecomment-887147239

src/state/trajectory/logics.ts
```diff
@@ -4,38 +4,41 @@ import queryString from "query-string";
 // NOTE: importing @aics/simularium-viewer here currently breaks ability to compile in testing setup
 // TODO: work on test babel setup, or switch to jest?
```
toloudis, line 5: The above comments are very intriguing :)

toloudis (approved):

src/containers/ViewerPanel/index.tsx
toloudis: I see, thanks!

src/state/configure-store.ts
schoinh: That's a good question. @meganrm what do you think? Right now `selection.logics` and `simularium.logics` are just empty arrays, but I thought I'd include them here now so we don't have to worry about it down the road if/when we do add logic functions to those branches.

src/state/configure-store.ts
toloudis: I see.  I guess my real thought is: in this file, it looks like these two lists really need to accumulate up the reducers and logics from every state branch.  Is that right?

schoinh: Yes. That is correct. This is where they all come together into one unified state tree

src/state/configure-store.ts
toloudis: It's also good to know that if you add more state you have to come here and insert the new stuff.

src/state/trajectory/logics.ts
schoinh: Yeahhh I think a long time ago I tried switching (partially?) to Jest to see if it fixed anything per @meganrm 's suggestion but it didn't make a difference. Maybe the import will work now? haha

src/state/configure-store.ts
meganrm: yeah, so @schoinh you're right that it's better to add the empty arrays. It can be really frustrating working on some logic that hits an external source and have it not work because you forgot it actually import it into state. 

@toloudis the separation of state into the different folders/branches is purely for our own organization and sanity. Redux only has one big reducer. Thats why we take the trouble to make sure the constants we use in the reducer are named for their branch, otherwise you can have namespace overlap and have an action get called you don't expect.

src/state/trajectory/actions.ts
meganrm: I think it should be renamed in this work, you're renaming the branch from 'metadata' to 'trajectory' so anything that references metadata should now reference trajecory. I think it could just be called "recieveTrajectory"

schoinh: Okay. The reason I thought something like `receiveTrajectoryMetadata` would be better than `receiveTrajectory` is because we have a `receiveSimulariumFile` action in the trajectory state branch as well. I thought having both `receiveTrajectory` and `receiveSimulariumFile` might sound like it's the same thing when `receiveTrajectory` receives the metadata only (stuff from `TrajectoryFileInfo`) while `receiveSimulariumFile` receives a `simulariumFile` object which has `name`, `data`, and `lastModified`. Do you still think `receiveTrajectory` is better to use since it matches the name of the state branch?

schoinh: I just renamed a bunch of variables containing `metadata` (including `receiveMetadata`).

meganrm (approved):

PR #227, opened 2021-08-02
Put basic and composed selectors in separate files

schoinh:
Problem
=======
We want to aim for 100% test coverage on Redux selectors, but only on selectors that are more than simple getters ("basic selectors"). Currently the basic and composed selectors are all in the same file, and there is no good way to make Jest ignore a block of code for test coverage purposes. 

This is a baby step for #224 

Solution
========
Separate basic and composed selectors into different files so that we can set a code coverage threshold for just the composed selectors

* Instead of having a single `selectors.ts` file for every state branch, I added a `selectors` folder that contains a `basic.ts` and `index.ts`. 
    * `basics.ts` holds all the basic selectors. 
    * `index.ts` holds all the composed selectors and also exports the basic selectors as well so that the rest of the app doesn't have to worry about the basic/composed distinction or has to have any import paths changed. 
* The Jest config file now only has a coverage threshold for `selectors/index.ts` so that it doesn't matter that we don't have tests for the basic selectors. 

Thanks @meganrm for discussion.

## Type of change

* Maintenance

Steps to Verify:
----------------
1. `npm run test` and check that Jest is measuring test coverage on selectors/index.ts and basics.ts separately
1. `npm start` and make sure the app still works


schoinh:
The commit history above has commits from the Redux reorg because of git squash etc. BUT the diffs in this PR show the correct diffs.

meganrm (approved):

EricJIsaac (approved):

PR #229, opened 2021-08-05
100% test coverage of state selectors + some cleanup

schoinh:
Problem
=======
We want more test coverage. A significant step towards completing #182 

Solution
========
* Add more tests to state selectors to get 100% test coverage.

    Work still remaining to close out this issue: test coverage for container-specific selector files. For example, src/containers/ModelPanel/selectors.ts has no tests. Because the test file doesn't exist, Jest currently doesn't calculate code coverage for that container's selectors. Ironic!

* I also renamed `visibleAgentKeys` in `state.selection` to `agentVisibilityMap` and `highlightedAgentKeys` to `agentHighlightMap` to make them more accurate and less confusing, per discussion with @meganrm 

* Removed some old unused selectors

## Type of change

Steps to Verify:
----------------
1. Check that test coverage is 100% for all selectors under src/state - this should be taken care of by GitHub Actions check
2. Check that variable renaming didn't break anything - `npm start`

-----------------------
1. src/state/selection/selectors/selectors.test.ts
2. src/state/trajectory/selectors/selectors.test.ts

Everything else is variable renaming or deleting of unused selectors.

Thanks for contributing!


src/containers/ViewerPanel/selectors.test.ts
```diff
@@ -1,12 +1,12 @@
-import { initialState } from "../../../state/index";
```
schoinh, line 1: I took this test file out of the `test` folder because it was the only test file in there. Kinda irrelevant to this PR, sorry

src/state/selection/selectors/index.ts
```diff
@@ -11,7 +11,7 @@ import {
     getVisibleAgentsNamesAndTags,
 } from "./basic";
 
-export const getHightLightedAgents = createSelector(
```
schoinh, line 14: Typo fix

src/state/selection/selectors/selectors.test.ts
```diff
@@ -35,50 +39,137 @@ const mockState: State = {
...
+                        agent1: ["agent1", "v-agent2-state2"],
+                        agent2: [""],
+                    },
                 },
             };
             const hiddenNames = getAgentsToHide(stateWithSelection);
             expect(hiddenNames).toBeInstanceOf(Array);
             expect(hiddenNames).toEqual([
                 { name: "agent1", tags: ["state1", ""] },
+                { name: "agent2", tags: ["state1", "state2"] },
```
schoinh, line 132: @meganrm I'm actually still kinda confused about this function `getAgentsToHide`. Why is it written so that if there are 3 agents, and only 2 agents are under `visibleAgentsNamesAndTags`, the third agent is not included in the list of hidden names? i.e. if `agent3` is not listed under `visibleAgentsNamesAndTags` shouldn't it be listed under `hiddenNames`? I see that `getAgentsToHide` is written to keep `agent3` from being added to the output, but why do we want/need it that way?

src/state/trajectory/selectors/index.ts
```diff
@@ -29,31 +24,6 @@ export const getIsNetworkedFile = createSelector(
     }
 );
 
-export const getKeysOfTrajectory = createSelector(
```
schoinh, line 32: Removed unused selectors

src/state/selection/selectors/selectors.test.ts
schoinh: Just talked with Megan and renamed some more variables and rewrote these tests

src/containers/ModelPanel/index.tsx
```diff
@@ -51,8 +51,8 @@ const styles = require("./style.css");
 
 interface ModelPanelProps {
     uiDisplayDataTree: AgentDisplayNode[];
-    highlightedAgentKeys: VisibilitySelectionMap;
-    visibleAgentKeys: VisibilitySelectionMap;
+    agentHighlightMap: VisibilitySelectionMap;
+    agentVisibilityMap: VisibilitySelectionMap;
```
meganrm, line 55: much better naming

src/state/selection/reducer.ts
```diff
@@ -69,8 +69,8 @@ const actionToConfigMap: TypeToDescriptionMap = {
         ) => {
             return {
                 ...state,
-                visibleAgentKeys: {
-                    ...state.visibleAgentKeys,
+                visibleAgentsNamesAndTags: {
+                    ...state.visibleAgentsNamesAndTags,
```
meganrm, line 73: why isn't this called `agentVisibilityMap`?

src/state/selection/reducer.ts
```diff
@@ -28,8 +28,8 @@ import {
 export const initialState = {
     time: 0,
     numberPanelsCollapsed: 0,
-    visibleAgentKeys: {},
-    highlightedAgentKeys: {},
+    agentVisibilityMap: {},
+    agentHighlightMap: {},
     agentColors: {},
```
meganrm, line 33: based on the basic selectors for `selection` it seems like initalState is either missing some keys, or there are some old selectors that can be deleted. like `export const getAgentsOnById = (state: State) => state.selection.agentIdsOn;`

src/state/selection/reducer.ts
```diff
@@ -54,7 +54,7 @@ const actionToConfigMap: TypeToDescriptionMap = {
         ) => {
             return {
                 ...state,
-                visibleAgentKeys: action.payload,
+                visibleAgentsNamesAndTags: action.payload,
```
meganrm, line 57: `agentVisibilityMap`?

src/state/selection/reducer.ts
schoinh: Oh because I missed it! Thanks for catching it, just fixed it

src/state/selection/reducer.ts
schoinh: Fixed

src/state/selection/reducer.ts
schoinh: Good catch, looks like `getAgentsOnById` is not used anywhere. I just checked all the basic selectors (had only checked composed ones) and removed the unused ones.

src/state/selection/selectors/index.ts
toloudis: nit: HighLighted --> Highlighted

schoinh: Oh good eye, thanks! Fixed

meganrm (approved):

toloudis (approved):

EricJIsaac (approved):

PR #230, opened 2021-08-09
Fix/ "Show/Hide All" checkbox bugs (Add tests for ModelPanel selectors)

schoinh:
Problem
=======
I wrote tests for the `ModelPanel` container's selectors and found some bugs having to do with the "Hide/Show All" checkbox. So I fixed the bugs too.

Repro for bugs:
1. Go to https://staging.simularium.allencell.org/viewer?trajFileName=endocytosis.simularium
2. Uncheck an agent in the left panel. Notice that the "Hide/show All" checkbox at the top of the list changes from a check mark to a hyphen (["indeterminate" checkbox state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate)) (Correct behavior)
3. Now go to https://staging.simularium.allencell.org/viewer?trajFileName=pc4covid19.simularium
4. Uncheck "Dendritic cell" in the left panel. Notice that the "Hide/show All" checkbox at the top changes from a check mark to a blank checkbox (Incorrect behavior -- this checkbox should only be blank if nothing on the list is checked)
5. Now click on the arrow next to "Dendritic cell" to unfurl its states
6. Check a box on one of the dendritic cell's states (like "G0/G1 phase") to make it visible. Notice that the "Hide/show all" checkbox changed to a check mark (Incorrect behavior -- this checkbox should only be checked if everything on the list is checked)

Solution
========
1. I wrote tests for the `ModelPanel` selectors, including the selector that determines whether the "Show/Hide All" checkbox should be in an indeterminate state or not
2. The old selector `getCheckboxAllIsIntermediate` was failing I believe 3 of the new tests. I confirmed something was wrong by testing out the UI in staging (see repro steps above). 
3. I rewrote the above selector to fix the bugs and make the tests pass. It might not be the most optimal way to do it, but it seems to function correctly!

Also renamed some variables for clarity:
* `getCheckboxAllIsIntermediate` -> `getIsSharedCheckboxIndeterminate`
* `convertUITreeDataToSelectAll` -> `getSelectAllVisibilityMap`
* `convertUITreeDataToSelectNone` -> `getSelectNoneVisibilityMap`

And added an `npm run test-watch` script to package.json so we don't have to run `npm run test` after every change

## Type of change

Steps to Verify:
----------------
1. `npm start` and check that the Hide/Show All checkbox behaves correctly now
2. `npm run test` - taken care of by GitHub Actions check

-----------------------
1. src/containers/ModelPanel/selectors.ts - Rewrite of `getCheckboxAllIsIntermediate`
2. src/containers/ModelPanel/selectors.test.ts - All new tests

Thanks for contributing!


src/containers/ModelPanel/selectors.ts
```diff
@@ -32,35 +36,37 @@ export const convertUITreeDataToSelectNone = createSelector(
...
-                    visibleAgents[agent.key].length < agent.children.length &&
-                    visibleAgents[agent.key].length > 0;
-                return false;
-            } else if (visibleAgents[agent.key]) {
-                return true;
+            if (!visibleStates.length) {
+                numInvisibleAgents++;
+            } else if (visibleStates.length < agent.children.length) {
+                isAgentPartiallyVisible = true;
+                break;
```
meganrm, line 61: cant you just return true here, and then you don't need the `isAgentPartiallyVisible` at all?

schoinh: Yes totally! Thanks! Fixed

src/containers/ModelPanel/selectors.ts
```diff
@@ -32,35 +36,37 @@ export const convertUITreeDataToSelectNone = createSelector(
...
+        if (isEmpty(agentVisibilityMap)) return false;
+
+        let isAgentPartiallyVisible = false;
+        let numInvisibleAgents = 0;
+
+        // Loop through each agent and count how many are invisible. If an agent is partially
+        // visible, break out of loop and just return true
+        for (let i = 0; i < allAgents.length; i++) {
+            const agent = allAgents[i];
+            const visibleStates = agentVisibilityMap[agent.key];
```
meganrm, line 54: I'm surprised that typeScript doesn't tell you `object this possibility undefined here` since it's using the key from one map to access another. Did giving types to the two inputs remove that?

schoinh: I think adding `if (isEmpty(agentVisibilityMap)) return false;` in line 46 took care of that

schoinh: Oh wait, I get what you're asking now. I didn't get an error like that, maybe because I put in the types for the inputs :)

schoinh: Maybe I should add a check here just in case `agentVisibilityMap` for some reason doesn't have `agent.key` as a key?

src/containers/ModelPanel/selectors.test.ts
```diff
@@ -0,0 +1,142 @@
...
+            },
+            {
+                label: "G0/G1 phase",
+                value: "G0/G1 phase",
+            },
+        ],
+    },
+    {
+        title: "Dendritic cell",
+        key: "Dendritic cell",
```
meganrm, line 38: This is a preference thing, and I don't think we've been consistent with it either, but I kinda like naming test cases like this by a description instead of just test data names. Like, "agentWithNoStates" and "agentWithOneState"

schoinh: Ok, just updated the names.

src/containers/ModelPanel/selectors.ts
schoinh: Added a check for when `visibleStates` is undefined

EricJIsaac:
simularium-website-checkbox-delay

Watch the **All** box in the gif. I don't think this should block this PR (the same is present but less noticeable on staging) - but this might have something to do with how long it takes to determine visibility/the appropriate mark for the box if I hazard a guess

EricJIsaac (approved):

schoinh:
> simularium-website-checkbox-delay
> 
> Watch the **All** box in the gif. I don't think this should block this PR (the same is present but less noticeable on staging) - but this might have something to do with how long it takes to determine visibility/the appropriate mark for the box if I hazard a guess

Interesting! I think we can keep an eye on it and see if it gets significantly more distracting with a much longer agent list. Unless someone has an alternative idea now that is.

meganrm (approved):

PR #231, opened 2021-08-10
Add tests for ResultsPanel selectors (100% coverage for all selectors)

### LINKED ISSUE #182: Get 100% unit test coverage on redux selectors

schoinh:
Problem
=======
Resolves #182 

This is the last PR for the above issue! After this is merged, all selectors in simularium-website will be 100% covered.

Solution
========
Added tests to the `ResultsPanel` container's selector and its helper functions, breaking down the functions further to make them more easily testable on the way.

## Type of change

Steps to Verify:
----------------
1. Check that refactoring the selector and helper functions didn't break the plots. Run `npm start`, load some models with plots (like endocytosis) and make sure the plots look formatted the same way as before.
2. GitHub Actions should take care of test coverage check

-----------------------
1. src/containers/ResultsPanel/selectors.test.ts - All new tests
2. src/containers/ResultsPanel/selectors.ts - Broke out `configureLayout` helper function into smaller functions to make them more easily testable; turned some style values into constants


src/containers/ResultsPanel/selectors.ts
```diff
@@ -149,7 +167,6 @@ const getShouldRenderTimeIndicator = (plot: RawPlotParams): boolean => {
 export const getPlotDataConfiguredForPlotly = createSelector(
     [getPlotData],
     (plotData: RawPlotParams[]): PlotConfig[] => {
-        if (!plotData) return [];
```
schoinh, line 152: This is duplicate logic so I removed it. TypeScript wouldn't allow a `null` or `undefined` or other falsy value to be passed in as `plotData`. And if `plotData` were an empty array, the `map` function below would just return an empty array anyway.

meganrm (approved):

EricJIsaac (approved):

PR #235, opened 2021-08-13
Fix/time input box is too narrow sometimes

### LINKED ISSUE #226: The time input box (current time display) is too narrow sometimes
## Description

For some models, the current time display (= time input box) is too narrow, so the current time gets cut off. 

image

## Expected Behavior

The time display / input box should be wide enough to show the entire time value without obscuring a part of it.

## Reproduction

1. Go to https://staging.simularium.allencell.org/viewer?trajUrl=https://www.dropbox.com/s/ipqad4n8c3bny75/actin.h5.simularium?dl=0
2. Hit play
3. Notice that you can only see the first digit of current time value. When you click on the time input box you can use your right arrow key to see that there are actually 2 to 3 sig figs but they are cut off because the input box is too narrow.

schoinh:
Problem
=======
Closes #226 

Solution
========
I wrote an additional test and modified the `getMaxNumChars` function so that it passes.

In a nutshell, `getMaxNumChars` was only considering the lengths of `round(lastFrameTime + timeStep)` and `firstFrameTime` as contenders for the longest value, but when `firstFrameTime` is 0 and `timeStep` is small enough, `round(lastFrameTime + timeStep)` just becomes `lastFrameTime`. So I added `round(firstFrameTime + timeStep)` into consideration.

## Type of change

Steps to Verify:
----------------
1. Try to reproduce the bug outlined in the issue. You shouldn't be able to.
2. All other tests for `getMaxNumChars` still pass.

Screenshots (optional):
-----------------------
### Before:
image

### After:
image

### DIFF
--- a/src/containers/ViewerPanel/selectors.ts
+++ b/src/containers/ViewerPanel/selectors.ts
@@ -60,16 +60,18 @@ export const getMaxNumChars = (
     lastFrameTime: number,
     timeStep: number
 ) => {
-    // If lastFrameTime is 15 and step size is 0.2 then 15.2 is probably going to have
-    // the max number of characters for this trajectory
-    const refTimeValue = lastFrameTime + timeStep;
-    const roundedRefTime = roundTimeForDisplay(refTimeValue).toString();
+    // These two time values are likely to have the most digits
+    const refTime1Value = firstFrameTime + timeStep;
+    const refTime2Value = lastFrameTime + timeStep;
+    const roundedRefTime1 = roundTimeForDisplay(refTime1Value).toString();
+    const roundedRefTime2 = roundTimeForDisplay(refTime2Value).toString();
 
     // Edge case: If firstFrameTime is a very small but long number like 0.000008,
     // we need to accommodate that.
     const maxNumChars = Math.max(
         firstFrameTime.toString().length,
-        roundedRefTime.length
+        roundedRefTime1.length,
+        roundedRefTime2.length
     );
 
     return maxNumChars;
--- a/src/containers/ViewerPanel/selectors.test.ts
+++ b/src/containers/ViewerPanel/selectors.test.ts
@@ -106,6 +106,20 @@ describe("ViewerPanel selectors", () => {
             expect(maxNumChars).toBe("44.1".length);
         });
 
+        it("returns length of firstFrameTime + timeStep when time step is a lengthy float and firstFrameTime is 0", () => {
+            const firstFrameTime = 0;
+            const lastFrameTime = 5;
+            const timeStep = 0.005;
+
+            const maxNumChars = getMaxNumChars(
+                firstFrameTime,
+                lastFrameTime,
+                timeStep
+            );
+
+            expect(maxNumChars).toBe("0.005".length);
+        });
+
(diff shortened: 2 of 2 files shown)


src/containers/ViewerPanel/selectors.ts
```diff
@@ -60,16 +60,18 @@ export const getMaxNumChars = (
...
 ) => {
-    // If lastFrameTime is 15 and step size is 0.2 then 15.2 is probably going to have
-    // the max number of characters for this trajectory
-    const refTimeValue = lastFrameTime + timeStep;
-    const roundedRefTime = roundTimeForDisplay(refTimeValue).toString();
+    // These two time values are likely to have the most digits
+    const refTime1Value = firstFrameTime + timeStep;
+    const refTime2Value = lastFrameTime + timeStep;
+    const roundedRefTime1 = roundTimeForDisplay(refTime1Value).toString();
+    const roundedRefTime2 = roundTimeForDisplay(refTime2Value).toString();
```
toloudis, line 67: Is roundTimeForDisplay clamping between firstFrameTime and lastFrameTime?  Just wondering if you might expect lastframe-1 time + timestep to be a good "large" value?

toloudis: What I'm getting at is that I would prefer numbers closer to the last frame rather than the first, as they will hopefully be bigger.

toloudis (approved):

src/containers/ViewerPanel/selectors.ts
schoinh: Talked with @toloudis . In the long run it would be better to have `roundTimeForDisplay` round to a number of decimal points (which is determined by `stepSize` probably -- and if we don't have any decimal points, `maxNumChars` would simply be `lastFrameTime.length`, I think) rather than to a number of sig figs as it is doing currently. That would probably allow us to determine the `maxNumChars` with more certainty than just guessing. It would also make the time display less chaotic during playback because the length of the time value won't shrink and grow repeatedly. We should discuss this at the next UX meeting.

For now I can't think of another test case that would make `getMaxNumChars` fail, so I think it won't hurt to merge these changes in until we decide exactly what to do in the long run?

toloudis: If we didn't come up with a case to break things, that's probably a good sign.  There's still the question of watching numbers tick by and the decimal point is potentially shifting.

EricJIsaac: It might work well to...
1. Observe how many decimal digits are in the time-step size
2. Fix the display to always display that number of decimal places (fill in '0's when needed)
3. Use (Non-decimal digits in last time-step) + (decimal digits in time-step size) to determine the set size for the time display

This would fix the decimal point in one place, making it easier to tell that the number is continuously increasing (it will visually get longer), and remove some of the guesswork in the display size

EricJIsaac (approved):

src/containers/ViewerPanel/selectors.ts
schoinh: Agreed @EricJIsaac !

PR #253, opened 2021-10-07
Feature/load local geometry files

### LINKED ISSUE #237: Accept multiple geometry files to load in viewer
## Use Case
It would be nice if users could drag and drop their geometry files with their trajectory file.

## Acceptance Criteria
A user can provide multiple files, either as a selected set or as a directory, and see them load in the viewer, including custom geometry where it was provided.

## Details
Support for user specified geometry files is being added to simularium-viewer, but the website needs to allow loading multiple files in order to use this feature (so this depends on https://github.com/allen-cell-animated/simularium-viewer/milestone/1)

This should work both for drag/drop and the file dialog.

schoinh:
Problem
=======
Closes #237 

Solution
========
Modified the drag-and-drop upload component (`ViewerOverlayTarget`), the file dialog upload component (`LocalFileUpload`), and the file-processing module that these components both use (`custom-request-upload.ts`) to accept multiple files, separate out .simularium file and geometry files, and tell the viewer to load the trajectory with the given geometry assets.

Antd version update was necessary for using the `onDrop` prop for the `Upload` component.

#### What this PR does:
* Allow user to upload multiple files at once through drag and drop
* Allow user to upload multiple files at once through the file upload dialog
* Lets user know if they didn't upload a geometry file/files that is/are needed (propagates error messages from simularium-viewer). Trajectory will still load but missing geometry will default to spheres, just like in the viewer testbed.

#### This PR does _not_:
* Allow user to upload a folder through drag and drop or the file upload dialog
* Handle all edge cases (ex/ if user only uploads a geometry file, the viewer will be stuck at loading forever) - I've made an issue for error handling (#254) and work on it right after this one

## Type of change

* Breaking change (some minor styling broke because of Antd version update -- in the screenshot, notice checkbox misalignment and carets purple instead of white -- I've also made an issue for this (#255) and will work on that after the error handling task)

Steps to Verify:
----------------
1. Pull this branch, `npm i` and `npm start`
2. Download this zip file [actin_bundle.zip](https://github.com/allen-cell-animated/simularium-website/files/7297736/actin_bundle.zip) and unzip it
3. Drag and drop or use the upload file dialog to load the 4 files inside the unzipped folder into the viewer
4. The trajectory should load successfully, you shouldn't see any spheres (it should look like the screenshot below), and it should play normally
5. Switching to and from different trajectories, and everything else, should work like normal

Screenshots (optional):
-----------------------
image

-----------------------
1. src/components/LocalFileUpload/index.tsx - File upload dialog `Upload` component
2. src/components/ViewerOverlayTarget/index.tsx - Drag and drop `Upload` component
3. src/components/LocalFileUpload/custom-request-upload.ts - used by the above 2 copmonents

### DIFF
--- a/src/components/LocalFileUpload/custom-request-upload.ts
+++ b/src/components/LocalFileUpload/custom-request-upload.ts
@@ -46,7 +46,9 @@ export default (
             numCustomRequests = 0;
         }
         return;
-    } else if (selectedFiles.length === 1) {
+    } 
+    
+    if (selectedFiles.length === 1) {
         // numCustomRequests and selectedFiles.length are both 1, so reset
         numCustomRequests = 0;
     }


src/components/HelpMenu/index.tsx
```diff
@@ -29,13 +29,13 @@ const HelpMenu: React.FunctionComponent<{}> = () => {
         );
     const menu = (
         <Menu theme="dark" className={styles.menu}>
-            <Menu.Item>{tutorialLink}</Menu.Item>
-            <Menu.Item>
+            <Menu.Item key="tutorial">{tutorialLink}</Menu.Item>
```
schoinh, line 32: These `key` additions were required as a result of Antd update

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,113 @@
-import { RcCustomRequestOptions } from "antd/lib/upload/interface";
+import { SimulariumFileFormat } from "@aics/simularium-viewer";
+import { findIndex } from "lodash";
+import {
+    UploadRequestOption,
+    UploadRequestError,
+} from "rc-upload/lib/interface";
```
schoinh, line 6: The `UploadRequestOption` here is the same thing as `{ RcCustomRequestOptions } from "antd/lib/upload/interface"`, but with the Antd update the old import statement was not working, so I had to modify it

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,113 @@
...
 import { LocalSimFile } from "../../state/trajectory/types";
 import { CLEAR_SIMULARIUM_FILE } from "../../state/trajectory/constants";
 import { store } from "../..";
+
+// Typescript's File definition is missing this function
+//  which is part of the HTML standard on all browsers
+//  and needed below
+interface FileHTML extends File {
+    text(): Promise<string>;
+}
```
schoinh, line 17: This and much of this file from line 57 onward is copied (with some modifications) from the simularium-viewer testbed

src/components/LocalFileUpload/index.tsx
```diff
@@ -11,30 +11,55 @@ interface FileUploadProps {
...
+                to={{
+                    pathname: VIEWER_PATHNAME,
+                    state: { localFile: true },
+                }}
             >
                 <Button type="ghost">From your device</Button>
-            </Upload>
-        </Link>
+            </Link>
+        </Upload>
```
schoinh, line 62: Another minor change required by Antd update that doesn't change any behavior

src/components/ViewerTitle/index.tsx
```diff
@@ -44,7 +44,7 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
         lastModified && lastModified - Date.now() < 0;
     if (shouldShowModifiedDate) {
         tagText = `modified: ${moment(lastModified).format(
-            "YYYY-MM-DD, h:m A"
+            "YYYY-MM-DD, h:mm A"
```
schoinh, line 47: Unrelated change that shows "6:09 PM" instead of "6:9 PM" for example for last modified time

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,113 @@
...
+    Promise.all(files.map((file: FileHTML) => file.text())).then(
+        (parsedFiles: string[]) => {
+            try {
+                const simulariumFileIndex = findIndex(files, (file) =>
+                    file.name.includes(".simularium")
+                );
+                // TODO: handle when user doesn't load a .simularium file
+                // if (simulariumFileIndex === -1) {
+                //    throw new Error("Please upload a .simularium file.")
+                // }
```
toloudis, line 66: why is this code commented out? Is it not sufficient to throw Error?

schoinh: It's not sufficient -- I need to do something with it in the `catch` block in line 100

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,113 @@
...
+                    onSuccess(
+                        {
+                            name: simulariumFile.name,
+                            status: "done",
+                            url: "",
+                        },
+                        new XMLHttpRequest() // onSuccess needs an XMLHttpRequest arg
+                    );
+                }
+                isLoading = false;
```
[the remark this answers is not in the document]
schoinh: That is a really good point. I'll make this change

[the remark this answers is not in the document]
meganrm: yeah this seems good, but it does mean you're making the user wait *slightly longer, I was thinking you'd process on the first call but not reset anything, and then reset the counter on the last call? I am not sure I understand why that would cause a problem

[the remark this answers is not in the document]
schoinh: Oh my god why didn't I think of that.... hang on :)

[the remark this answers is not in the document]
schoinh: Done

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,117 @@
...
+                    onSuccess(
+                        {
+                            name: simulariumFile.name,
+                            status: "done",
+                            url: "",
+                        },
+                        new XMLHttpRequest() // onSuccess needs an XMLHttpRequest arg
+                    );
+                }
+                numCustomRequests = 0;
```
schoinh, line 103: Delete this

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,117 @@
...
-                    url: "",
-                },
-                file
-            )
-        )
-        .catch((error) => {
-            console.log(error);
-            onError(error);
+    numCustomRequests++;
+    if (numCustomRequests === 1) {
```
schoinh, line 39: Flip the order of if and else blocks (don't need else)

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -1,39 +1,119 @@
...
+        // If the user loads multiple files at once (.simularium file + geometry file(s)),
+        // this function is called multiple times, but we only need to process
+        // and load the trajectory once.
+        if (numCustomRequests === selectedFiles.length) {
+            // We're at the last customRequest for this batch of files,
+            // so reset counter
+            numCustomRequests = 0;
+        }
+        return;
+    } else if (selectedFiles.length === 1) {
```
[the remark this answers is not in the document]
toloudis:
```suggestion
    } 
    
    if (selectedFiles.length === 1) {
```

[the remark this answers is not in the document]
schoinh: Right, thanks!

meganrm (approved):

PR #259, opened 2021-10-12
Fix/file load error handling

### LINKED ISSUE #254: Better error handling for geometry file uploads
## Use Case

As a user, I want to see helpful information when something is not right about the files I tried to load to Simularium.

## Acceptance Criteria

* The viewer displays a helpful message(s) instead of freezing, crashing, or getting stuck in a "loading" state if anything goes wrong with file upload.
* When I upload a trajectory file but not all of the geometry files that are required by that trajectory, I get notified of which ones are missing.
* When I upload a geometry file(s) but not a .simularium file, I get notified that a .simularium file is missing.
* When I try to upload a folder through drag-and-drop (it's impossible through the file dialog), I get notified I need to upload files instead.

@blairlyons @meganrm Did I miss anything?

schoinh:
Problem
=======
Closes #254 

Solution
========
* Replaced the try-catch block inside of the `then` callback in the customRequest function with a `Promise.catch()` to catch all errors, including exceptions that can happen inside the `Promise.all()` async function
* In the `catch` callback, propagate the caught error message to the app state using the `SET_STATUS` action
* Supply appropriate, helpful error messages depending on what the problem is

Bonus changes
* Made the error notification display for 10 seconds instead of 4.5 seconds (can manually close anytime)
* Tweaked/improved the util function that turns strings into sentence case so that a file extension mentioned in an error message doesn't get capitalized

## Type of change

Steps to Verify:
----------------
1. Download this zip file [actin_bundle.zip](https://github.com/allen-cell-animated/simularium-website/files/7297736/actin_bundle.zip) and unzip it
2. With the downloaded files, try these scenarios. None of these scenarios should break the viewer (= viewer freezes/hangs without giving you a helpful error message).
    1. Only load a geometry file or files (or any other file(s) that don't include a .simularium file)
    2. Drag and drop a folder instead of a group of files
    3. Load a trajectory file that requires local geometry files, but without any geometry files
    4. Load a trajectory file that requires local geometry files, but without one of the geometry files
    5. Load a trajectory that doesn't require any local geometry files
3. Try to break the viewer with some other kind of file upload scenario I didn't list above

Screenshots (optional):
-----------------------
- File(s) did not include a .simularium file:
    
    image
- Didn't include geometry files (error message text is from the viewer):

    image

- Tried to load a folder:

    image

-----------------------
src/components/LocalFileUpload/custom-request-upload.ts

### DIFF
--- a/src/components/LocalFileUpload/custom-request-upload.ts
+++ b/src/components/LocalFileUpload/custom-request-upload.ts
@@ -6,8 +6,6 @@ import {
 } from "rc-upload/lib/interface";
 
 import { LocalSimFile } from "../../state/trajectory/types";
-import { CLEAR_SIMULARIUM_FILE } from "../../state/trajectory/constants";
-import { store } from "../..";
 import { VIEWER_ERROR } from "../../state/viewer/constants";
 import { ViewerStatusInfo } from "../../state/viewer/types";
 
@@ -35,6 +33,7 @@ into the viewer.
 export default (
     { onSuccess, onError }: UploadRequestOption,
     selectedFiles: File[],
+    clearSimulariumFile: (isNewFile: { newFile: boolean }) => void,
     loadFunction: (simulariumFile: LocalSimFile) => void,
     setViewerStatus: (status: ViewerStatusInfo) => void
 ) => {
@@ -59,10 +58,7 @@ export default (
     // want the loading indicator to show without any lag time
     // as soon as user hits "Open" button or drops files,
     // and not have to have this action called multiple places in the code.
-    store.dispatch({
-        payload: { newFile: true },
-        type: CLEAR_SIMULARIUM_FILE,
-    });
+    clearSimulariumFile({ newFile: true });
 
     const files: FileHTML[] = Array.from(selectedFiles) as FileHTML[];
 
--- a/src/components/LoadFileMenu/index.tsx
+++ b/src/components/LoadFileMenu/index.tsx
@@ -6,6 +6,7 @@ import { Menu, Dropdown, Button } from "antd";
 import TRAJECTORIES from "../../constants/networked-trajectories";
 import { URL_PARAM_KEY_FILE_NAME } from "../../constants";
 import {
+    ClearSimFileDataAction,
     RequestLocalFileAction,
     RequestNetworkFileAction,
 } from "../../state/trajectory/types";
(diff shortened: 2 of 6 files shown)


src/containers/ViewerPanel/index.tsx
```diff
@@ -179,11 +179,11 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
     public componentDidUpdate(prevProps: ViewerPanelProps) {
         const { status, error } = this.props;
         const current = this.centerContent.current;
-        if (
+        const isNewError: boolean =
             status === VIEWER_ERROR &&
-            prevProps.status !== VIEWER_ERROR &&
-            error.message
-        ) {
+            error.message !== prevProps.error.message;
```
schoinh, line 184: This change is necessary to display multiple errors from the same file load (e.g. when a trajectory is missing 3 geometry files, there are 3 error message passed up from the viewer but only the first one was displaying before)

schoinh:
> I think this branch still has a problem: there are some errors that are supposed to warn the user but not stop the load process. This may mean we have to introduce the idea of "warning" but currently that's how it works on the test bed viewer app. Here is my collection of files, that purposely is missing some data, and it should warn the user that it is defaulting to spheres but it should still load. [geo-test-colors.zip](https://github.com/allen-cell-animated/simularium-website/files/7332514/geo-test-colors.zip)

Okay, yeah I didn't test any cases where the trajectory file itself is missing data, only cases having to do with the user loading the wrong thing or forgetting files. Let me see if this is a simple fix

meganrm:
> > I think this branch still has a problem: there are some errors that are supposed to warn the user but not stop the load process. This may mean we have to introduce the idea of "warning" but currently that's how it works on the test bed viewer app. Here is my collection of files, that purposely is missing some data, and it should warn the user that it is defaulting to spheres but it should still load. [geo-test-colors.zip](https://github.com/allen-cell-animated/simularium-website/files/7332514/geo-test-colors.zip)
> 
> Okay, yeah I didn't test any cases where the trajectory file itself is missing data, only cases having to do with the user loading the wrong thing or forgetting files. Let me see if this is a simple fix

Nevermind, this file is also missing spatialUnits and such, which is what was causing it to error out, when I fixed that it does load with the errors displaying. It would be nice if we sent in a "warning" attribute though so that the front end could display it with a different icon that the red X

src/components/ErrorNotification/index.tsx
```diff
@@ -29,7 +29,7 @@ const errorNotification = ({
                     }}
                 />
             ) || "",
-        duration: htmlData ? 0 : 4.5,
+        duration: htmlData ? 0 : 10,
```
toloudis, line 32: just curious, what is this number and why did it change?

src/components/LocalFileUpload/custom-request-upload.ts
```diff
@@ -56,66 +58,68 @@ export default (
...
+                        name: simulariumFile.name,
+                        status: "done",
+                        url: "",
                     },
-                    {}
+                    new XMLHttpRequest() // onSuccess needs an XMLHttpRequest arg
                 );
+            }
+        })
+        .catch((error) => {
```
toloudis, line 108: just so i understand, this catch happens if any of the files in the `map` call throws or rejects from file.text() ?

schoinh: Yup that's right! The `DOMException` in line 110 for example is thrown from `file.text()` and the error thrown in line 72 is also caught here

PR #265, opened 2021-10-21
remove unmodified state

### LINKED ISSUE #263: Remove "<unmodified>" from front end, send it from viewer
## Description

Unchecking or highlighting <unmodified> doesn't do anything in models where it isn't present. 

## Expected Behavior

Only have an <unmodified> tag if its present in the data. 

## Reproduction

Use one of the actin models and try turning the unmodified actin agent on and off. 

## Environment

_Any additional information about your environment. E.g. OS version, python version, browser etc._

meganrm:
Problem
=======
Some files don't have an unmodified state in their data, but we're always adding it. Which means sometimes there is a checkbox that doesn't do anything. 
closes #263 

Solution
========
Remove it from the front end, send it in with uiDisplayData

## Type of change

Steps to Verify:
----------------
1. npm start
2. won't show unmodified state for any of our current networked trajectories
3. install version of the viewer with that sends the unmodified state up
4. use actin test file to see unmodified state under arp3 but not the other two

Screenshots (optional):
-----------------------
Screen Shot 2021-10-21 at 12 43 43 PM

### DIFF
--- a/src/containers/ViewerPanel/selectors.test.ts
+++ b/src/containers/ViewerPanel/selectors.test.ts
@@ -31,7 +31,7 @@ describe("ViewerPanel selectors", () => {
     });
 
     describe("convertUIDataToSelectionData", () => {
-        it("converts UI display data to map with agent name as the key, and an array of display state names is the value (adding the empty string for unmodified state", () => {
+        it("converts UI display data to map with agent name as the key, and an array of display state ids is the value", () => {
             const result = convertUIDataToSelectionData(mockDisplayData);
             expect(result).toEqual({ name1: ["id1"] });
         });


src/containers/ViewerPanel/selectors.test.ts
```diff
@@ -33,7 +33,7 @@ describe("ViewerPanel selectors", () => {
     describe("convertUIDataToSelectionData", () => {
         it("converts UI display data to map with agent name as the key, and an array of display state names is the value (adding the empty string for unmodified state", () => {
```
schoinh, line 34: This test description should be updated

PR #275, opened 2021-11-03
Fix/agent list sometimes not loading for drag-and-drop

### LINKED ISSUE #274: Sometimes agent types in left panel don't load
## Description
Agent types don't load in left panel after first load of a trajURL page followed by drag and drop.

## Expected Behavior
Agent types should always load in the left panel.

## Reproduction
1. Go to a trajURL link, e.g. https://staging.simularium.allencell.org/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/example_md.simularium 
2. Make sure your cache was clear (i.e. force reload page), this only happens on the first load
3. Drag and drop https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/json_v3/actin012_3_obj-local-url_arp2-green_arp3-pink_actin-purple.simularium
4. The left panel goes back to it's empty state 
Screen Shot 2021-11-01 at 2 33 47 PM

## Environment
Staging

schoinh:
Problem
=======
Closes #274 

Solution
========
Instead of automatically rendering a `NoTypeMappingText` component ("Unabled to load UI controls") every time an error message is received from the viewer (e.g., when a geometry file can't be found), only render it if a list of agents (type mappings) doesn't exist in the data.

The `CheckBoxTree` component does its own check to see if the array for populating the agent list is empty and displays the `NoTypeMappingText` component if empty. 

## Type of change

Steps to Verify:
----------------
0. `npm start` on the main branch, throttle your browser to "Fast 3G" and try to reproduce the bug described in the issue. You should see the bug.
1. Pull this branch
2. Make sure your browser is still throttled and try to reproduce the bug. The agent list should show up now and the "Unable to load UI controls" message shouldn't ever flash in the left panel.

### DIFF
--- a/src/containers/ModelPanel/index.tsx
+++ b/src/containers/ModelPanel/index.tsx
@@ -100,7 +100,7 @@ class ModelPanel extends React.Component<ModelPanelProps, {}> {
             [VIEWER_ERROR]: isNetworkedFile ? (
                 <NetworkFileFailedText />
             ) : (
-                <NoTypeMappingText />
+                checkboxTree
             ),
         };
 
--- a/src/containers/ModelPanel/index.tsx
+++ b/src/containers/ModelPanel/index.tsx
@@ -40,7 +40,6 @@ import {
     VIEWER_SUCCESS,
 } from "../../state/viewer/constants";
 import NoTrajectoriesText from "../../components/NoTrajectoriesText";
-import NoTypeMappingText from "../../components/NoTrajectoriesText/NoTypeMappingText";
 import { RequestNetworkFileAction } from "../../state/trajectory/types";
 import { ViewerStatus } from "../../state/viewer/types";
 import NetworkFileFailedText from "../../components/NoTrajectoriesText/NetworkFileFailedText";


schoinh:
I think this fix might actually be robust enough to keep permanently and not just as a quick-and-dirty fix until we send up different levels of errors from the viewer, but I'm not positive 🤔

src/containers/ModelPanel/index.tsx
```diff
@@ -100,7 +100,7 @@ class ModelPanel extends React.Component<ModelPanelProps, {}> {
             [VIEWER_ERROR]: isNetworkedFile ? (
                 <NetworkFileFailedText />
             ) : (
-                <NoTypeMappingText />
+                checkboxTree
```
toloudis, line 103: can remove line 43 too now (linter says so!).  It won't let me put the comment on that line because it's not part of the diff, but the linter warning is right there

toloudis (approved):

EricJIsaac (approved):

src/containers/ModelPanel/index.tsx
schoinh: Oh thanks!!

PR #284, opened 2021-11-11
Feature/treat warning messages as warnings

### LINKED ISSUE #269: Handle different types of messages (error, warning, info) from viewer
## Use Case

As a Simularium user, I would like to see a distinction between critical errors (error), non-critical / gracefully handled errors (warning), and helpful messages (info). Currently, all "something went wrong / was unexpected" messages from the viewer are shown as errors, like this:

image

## Acceptance Criteria

* Display a yellow warning icon with a warning message from the viewer, and a blue info icon with an info message from the viewer. 
* Error messages continue to have a red error icon

## Details

_Please provide any helpful specifications_

schoinh:
Problem
=======
Now that we get warning, info, and error messages differentiated from the viewer, we should treat warnings and info messages differently from error messages. 

Closes #269 

Solution
========
1. Currently, the `setStatus` action does 2 things: set the viewer status (empty, loading, error, or success) AND set the error message in state. Setting the viewer status to "error" has effects we don't want with a warning or info (for example clearing out the agent panel), and in fact we don't want to change the viewer status when we get a warning or info at all. So I split up `setStatus` into `setStatus` and `setError`. Now, `setStatus` only sets the viewer status (empty, loading, error, or success) and `setError` puts the error message (with various error levels like warning, info, error) in our state so they can be turned into notifications for the user.
2. Modify the `ErrorNotification` component so that it displays warning and info icons (instead of a red X) for warnings and infos

### Other stuff
* Installed esbuild-jest and esbuild (required by esbuild-jest) to modify the jest config, which fixed this annoying problem we've had forever, where the test suite would break if we imported `@aics/simularium-viewer` into src/state/trajectory/logics.ts. This was necessary in this PR because I needed to import simularium-viewer into that logics file! (I'm honestly not sure why this config fixed the problem or why the problem existed in the first place, it was just a lot of trial and error)

## Type of change

Steps to Verify:
----------------
1. Pull this branch and `npm i` then `npm start`
2. Load a file that needs local geometry files, but don't supply the geometry files
3. You should see some warning messages pop up (instead of error). Everything else should work as before.

Screenshots (optional):
-----------------------
image

-----------------------
2. src/state/viewer/reducer.ts & src/state/viewer/types.ts - separate out `setStatus` into `setStatus` and `setError`
1. src/components/ErrorNotification/index.tsx - take error level into account when rendering a notification

Almost everything else is just having to update code everywhere `setStatus` has been used.


package.json
```diff
@@ -106,7 +108,7 @@
         "react-router-dom": "^5.2.0",
         "react-waypoint": "^9.0.3",
         "redux": "4.0.1",
-        "redux-logic": "^2.1.1",
+        "redux-logic": "^3.0.3",
```
schoinh, line 111: I updated this to try to fix something I think is a bug with redux -- it didn't fix it, but it didn't break anything else either, so I kept the update

src/containers/ModelPanel/index.tsx
```diff
@@ -99,7 +100,7 @@ class ModelPanel extends React.Component<ModelPanelProps, {}> {
             [VIEWER_ERROR]: isNetworkedFile ? (
                 <NetworkFileFailedText />
             ) : (
-                checkboxTree
+                <NoTypeMappingText />
```
schoinh, line 103: This is rolling back the temporary fix I put in for #274

src/state/viewer/types.ts
```diff
@@ -23,25 +25,20 @@ export interface ToggleAction {
...
-    status: ViewerStatus;
-    onClose?: () => void;
+export interface SetErrorAction {
+    payload: ViewerError;
+    type: string;
 }
 
 export interface ViewerError {
-    htmlData?: string;
+    level: ErrorLevel;
```
meganrm, line 38: small thing, but if we're exporting FrontEndError, I think you could do `interface ViewerError extends FrontEndError`

src/state/viewer/reducer.ts
```diff
@@ -32,17 +34,31 @@ const actionToConfigMap: TypeToDescriptionMap = {
...
-                          message: action.payload.errorMessage,
-                          htmlData: action.payload.htmlData,
-                          onClose: action.payload.onClose,
-                      }
-                    : "",
+            error: {
+                level: action.payload.level,
+                message: action.payload.message,
+                htmlData: action.payload.htmlData,
+                onClose: action.payload.onClose,
```
meganrm, line 60: you could just spread payload here. `error : {...action.payload}`

src/state/viewer/reducer.ts
```diff
@@ -32,17 +34,31 @@ const actionToConfigMap: TypeToDescriptionMap = {
     [SET_STATUS]: {
         accepts: (action: AnyAction): action is SetViewerStatusAction =>
             action.type === SET_STATUS,
-        perform: (state: ViewerStateBranch, action: SetViewerStatusAction) => ({
+        perform: (state: ViewerStateBranch, action: SetViewerStatusAction) => {
+            if (action.payload.status !== VIEWER_ERROR) {
+                return {
+                    ...state,
+                    status: action.payload.status,
+                    error: "", // Clear out error
```
meganrm, line 42: set this to initial state

src/state/trajectory/logics.ts
```diff
@@ -112,14 +110,16 @@ const loadNetworkedFile = createLogic({
...
         const simulariumFile = action.payload;
-        dispatch(
-            setStatus({
-                status: VIEWER_LOADING,
-            })
-        );
-        dispatch({
-            payload: { newFile: true },
-            type: CLEAR_SIMULARIUM_FILE,
+        batch(() => {
```
meganrm, line 113: 👍🏻

src/state/viewer/types.ts
schoinh: Ooh, I just tried it but I'm getting some errors because now I need to supply a `name` every time I use `setError`

image

src/state/viewer/reducer.ts
schoinh: You mean (pseudocode) `return {...initialState}`? I don't think we want to do that because we want to set the status to whatever the payload.status is.

meganrm: no I meant, `error: initialState.error`

meganrm (approved):
I just tested this with a file I have that currently crashes staging, and it works, so thumbs up. I did leave one comment that I think it's best to always be setting things in state to either something coming through the payload or to initialState[key] just so it's clear what's happening, but otherwise, looks good

EricJIsaac (approved):

toloudis (approved):

PR #285, opened 2021-11-13
Fix/old error message comes back after going to landing page

### LINKED ISSUE #282: Old error message pops up with new trajectory
## Description

When I fail to load a trajectory, go back to the landing page, then load another trajectory by clicking on a card, the previous error message shows up.

## Expected Behavior

I shouldn't see any irrelevant, old error messages popping up for a new trajectory.

## Reproduction

1. Go to staging viewer and load something that will fail, like an .obj geometry file by itself
2. Go to the landing page by clicking on "Simularium Home" in the nav bar (NOT by clicking the back button)
3. Wait for the error message in the upper right hand corner, if it's not gone already (We should probably also kill the error message once we leave the viewer page... that might be a different ticket)
3. Click on a trajectory card to load a different trajectory
4. The trajectory loads successfully, but the old error message about the .obj geometry file pops back up on the upper right hand corner!

## Details
I know why this is happening, we need to not return an `ErrorNotification` at `ViewerPanel` component mount. Returning `ErrorNotification` at component update is sufficient.

schoinh:
Problem
=======
Closes #282 

Solution
========
Don't render an error notification at `ViewerPanel` component mount -- it is rendered at component update, and that is enough.

## Type of change

Steps to Verify:
----------------
Try to reproduce the error described in the issue -- you shouldn't be able to.

### DIFF
--- a/src/containers/ViewerPanel/index.tsx
+++ b/src/containers/ViewerPanel/index.tsx
@@ -166,10 +166,6 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
                 this.resize(current);
             }, 200);
         }
-
-        if (error) {
-            return ErrorNotification({ ...error });
-        }
     }
 
     public componentDidUpdate(prevProps: ViewerPanelProps) {
--- a/src/containers/ViewerPanel/index.tsx
+++ b/src/containers/ViewerPanel/index.tsx
@@ -127,7 +127,6 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
     }
 
     public componentDidMount() {
-        const { error } = this.props;
         const browser = Bowser.getParser(window.navigator.userAgent);
         // Versions from https://caniuse.com/webgl2
         const isBrowserSupported = browser.satisfies({
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -4,7 +4,7 @@ import "core-js/es6/set";
 
 import React from "react";
 import { render } from "react-dom";
-import { Provider, useDispatch } from "react-redux";
+import { Provider, useDispatch, batch } from "react-redux";
 import { Layout } from "antd";
 import { BrowserRouter, Switch, Route, useLocation } from "react-router-dom";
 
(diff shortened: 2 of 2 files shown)


toloudis (approved):

src/containers/ViewerPanel/index.tsx
```diff
@@ -127,7 +127,6 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
     }
 
     public componentDidMount() {
```
meganrm, line 129: but doesn't this mean there is still erroneously an old error in state? Even though we aren't showing it?

EricJIsaac (approved):

src/containers/ViewerPanel/index.tsx
schoinh: That is a good point. I was thinking that whenever a viewer loads a new trajectory it does clear out the error, but it just happens after `componentDidMount`. That is true, but it doesn't account for when you leave the viewer with errors and return to the viewer without loading a new trajectory. I'm about to push up another commit

schoinh:
Let me know what you guys think of the most recent commit (which is meant to address @meganrm 's concern). Thanks!

src/index.tsx
```diff
@@ -34,31 +32,12 @@ function useLocationChange() {
...
+        if (location.pathname !== VIEWER_PATHNAME) {
+            batch(() => {
+                // if we're navigating away from the viewer, stop playing and reset state
+                dispatch(setIsPlaying(false));
                 dispatch(clearSimulariumFile({ newFile: false }));
-            }
-        } else {
-            // if we're navigating away from the viewer, stop playing
-            dispatch(setIsPlaying(false));
+            });
```
schoinh, line 40: Essentially, I'm proposing clearing out the state when leaving the viewer page instead of when arriving at the viewer page.

EricJIsaac (approved):

src/index.tsx
meganrm: this seems a lot cleaner. I had to think a bit about if this misses anything, but it seems good? Have you tested a lot of different navigation combos? Like loading a file, going back to home, loading a file from home, etc?

schoinh: Cool. Yeah, I haven't been able to break it yet :)

meganrm (approved):

PR #290, opened 2022-01-04
Fix/nav bar overflowing when title is long

### LINKED ISSUE #268: Nav bar contents overflow if the trajectory title is long
## Description

Nav bar contents overflow if the trajectory is too long to fit in one line.

image

## Expected Behavior

Nothing should overflow from the nav bar. My thought is to truncate the trajectory title like this, to fit the available space:

image

Also, display whole name on hover (tooltip)

## Reproduction

Go to staging and upload a trajectory file with a really long file name.

schoinh:
Problem
=======
Closes #268 

Solution
========
* Prevent nav bar contents from spilling out of the nav bar, for any reason
* Truncate a long trajectory title with "..." 
* Upon hovering over the trajectory title, show the full trajectory title in a tooltip after 1 second. Tooltip disappears immediately upon leaving the trajectory title.

## Type of change

Steps to Verify:
----------------
1. Clone this repo 
2. Pull this branch
3. `npm i` then `npm start`
4. Try to repro the bug as outlined in the issue, nothing should overflow from the nav bar anymore.

Screenshots (optional):
-----------------------
image

Upon hover:
image

### DIFF
--- a/src/components/ViewerTitle/index.tsx
+++ b/src/components/ViewerTitle/index.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { useLocation } from "react-router-dom";
-import { Tag } from "antd";
+import { Tag, Popover } from "antd";
 import moment from "moment";
 
 import { VIEWER_PATHNAME } from "../../routes";
@@ -55,11 +55,15 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
             <span />
         );
 
+    const popover = title;
+
     return (
-        <div className={styles.container}>
-            <div className={styles.title}>{title}</div>
-            <div>{tag}</div>
-        </div>
+        <Popover content={popover}>
+            <div className={styles.container}>
+                <div className={styles.title}>{title}</div>
+                <div>{tag}</div>
+            </div>
+        </Popover>
     );
 };
 
--- a/src/components/ViewerTitle/index.tsx
+++ b/src/components/ViewerTitle/index.tsx
@@ -19,11 +19,9 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
     const { simulariumFileName, lastModified } = props;
     const location = useLocation();
     const title =
-        location.pathname.startsWith(VIEWER_PATHNAME) && simulariumFileName ? (
-            <span>{simulariumFileName}</span>
-        ) : (
-            <span />
-        );
+        location.pathname.startsWith(VIEWER_PATHNAME) && simulariumFileName
+            ? simulariumFileName
+            : "";
 
     // Grab the trajectory ID from the URL and find the corresponding trajectory object in
     // networked-trajectories.ts to get its version info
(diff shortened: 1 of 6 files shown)


src/styles/ant-vars.less
```diff
@@ -263,4 +263,24 @@ https://github.com/ant-design/ant-design/blob/master/components/style/themes/def
...
+// Popover arrow width
+@popover-arrow-width: 6px;
+// Popover arrow color
+@popover-arrow-color: @popover-bg;
+// Popover outer arrow width
+// Popover outer arrow color
+@popover-arrow-outer-color: @popover-bg;
+// Popover distance with trigger
+@popover-distance: 0px;
+@popover-padding-horizontal: @padding-md;
```
schoinh, line 286: I mainly just changed the colors here. You can see the original version here: https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less#L522

src/components/ViewerTitle/index.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { useLocation } from "react-router-dom";
-import { Tag } from "antd";
+import { Tag, Popover } from "antd";
```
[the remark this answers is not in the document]
toloudis: sounds delicious

src/components/ViewerTitle/index.tsx
```diff
@@ -58,9 +56,12 @@ const ViewerTitle: React.FunctionComponent<ViewerTitleProps> = (
         );
 
     return (
-        <span className={styles.container}>
-            {title} {tag}
-        </span>
+        <div className={styles.container}>
+            <Popover content={title} mouseEnterDelay={1} mouseLeaveDelay={0}>
```
meganrm, line 60: So this popover shows the title even if the title is fully showing? Seems like we only want this if the title is truncated

toloudis (approved):
For the record, I tend to agree with whoever said that the modified date should only show in the popover and not on the main screen.  Just to reduce information clutter.

src/components/ViewerTitle/index.tsx
schoinh: Yeah, it shows regardless. I agree it would be ideal to only show when the title is truncated, but I thought about the additional work to make that happen, and the possibility of putting additional information in the popover in the future, and instead just put in a 1 second delay for the tooltip appearing so it's not too annoying.  After I saw your comment I tried to make something work with useRef and useLayoutEffect and haven't had success yet. I'll give it another try tomorrow if you think this is really important!

schoinh:
I nudged the tooltip position up a little bit per @lynwilhelm 's request. 

image

Per today's discussion, we will leave the tooltip as is for now (tooltip is available both when title is truncated and when it isn't), with the long term idea being we want a little info icon you can hover over or click that shows more info (the full file name, trajectory title, modified date, etc.) in a popover.

meganrm (approved):

PR #303, opened 2022-01-21
Fix/server error messages not displaying

### LINKED ISSUE #302: Server error message doesn't display
## Description

When the server is down, and the user tries to stream a trajectory, the viewer just spins in place forever without any notification to the user. 

## Expected Behavior

The viewer should reset, and the user should get a message that connecting to server has failed (per https://github.com/simularium/simularium-viewer/pull/173).

## Reproduction

Change the dev server URL to something invalid and try to load a trajectory.

schoinh:
Problem
=======
Server error messages aren't displaying to the user like they're supposed to. 

Closes #302 

Solution
========
This bug was happening because I was misusing a React hook (`useDispatch`) -- React hooks can only be used inside a function component. I was using it in a helper function for a redux logic function. 

I fixed the issue by passing in `dispatch` to the helper function from the invoking logic function instead. 

## Type of change

Steps to Verify:
----------------
1. Change the dev server URL in webpack/plugins.js to something invalid
2. `npm start`
3. Try to load a model
4. After a few seconds of attempting to load, the viewer should reset, and you should see an error message like this:

image

### DIFF
--- a/src/state/trajectory/logics.ts
+++ b/src/state/trajectory/logics.ts
@@ -1,6 +1,7 @@
 import { AxiosResponse } from "axios";
-import { batch, useDispatch } from "react-redux";
+import { batch } from "react-redux";
 import { createLogic } from "redux-logic";
+import { ArgumentAction } from "redux-logic/definitions/action";
 import queryString from "query-string";
 import { ErrorLevel, FrontEndError } from "@aics/simularium-viewer";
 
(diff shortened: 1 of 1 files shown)


src/state/trajectory/logics.ts
```diff
@@ -106,9 +107,12 @@ const requestPlotDataLogic = createLogic({
...
 });
 
-const handleFileLoadError = (error: FrontEndError) => {
-    const dispatch = useDispatch();
-
+const handleFileLoadError = (
+    error: FrontEndError,
+    dispatch: <T extends ArgumentAction<string, undefined, undefined>>(
+        action: T
+    ) => T
```
schoinh, line 114: If someone has a better idea for how to type `dispatch` let me know. I just copied and pasted what TS was telling me to be the type of `dispatch` in the logic functions that call this function.

toloudis: it looks odd to me but passing `dispatch` in does seem to be the pattern that the parent function uses too.   If that type def makes ts linting and typecheck happy, then I'm ok with it.

toloudis (approved):

meganrm (approved):

PR #306, opened 2022-01-28
Update eslint from v5.14.1 to v8.7.0

schoinh:
Problem
=======
We're using an old version of eslint that's less secure and outdated in general.

Solution
========
* Updated eslint and typescript-eslint plugins to latest versions
* Fixed new linting errors and warnings that resulted from above update
* Updated node version in GitHub Actions workflows from v12 to v14
* Fixed `npm run lint` script so that it actually works across platforms (current script was not picking up errors in MacOS)

Unrelated small change: 
* Changed server node URL from dev to staging in webpack/plugins.js (dev server doesn't exist at the moment)

Steps to Verify:
----------------
1. GitHub Actions checks (which includes linting) pass
2. Locally running `npm run lint` and `npm run typeCheck` works
3. Application still runs 

-----------------------
1. package.json

Most changes are linting error fixes.


src/components/BlankCard/index.tsx
```diff
@@ -7,7 +7,7 @@ import ImportThumbnail from "../../assets/import-thumbnail.svg";
 
 const styles = require("../ModelCard/style.css");
 
-const BlankCard: React.FunctionComponent<{}> = () => {
+const BlankCard = () => {
```
toloudis, line 10: removing the React.FunctionComponent typing?

toloudis: I found this https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/ which recommends against it too.  I do like the idea of typing the return value but it shouldn't hold back this PR.

toloudis (approved):

src/components/BlankCard/index.tsx
schoinh: Yeah, I removed them because the linter gave these errors:
```
  17:63  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead  @typescript-eslint/ban-types
```
And then I thought `React.FunctionComponent<Record<string, never>>` looked way too ugly for something so routine/simple, so I looked up some examples of how to write a TS function component and saw that those return types are unnecessary. 😄

toloudis: This one is my favorite:
```
const App = ({ message }: AppProps): JSX.Element => <div>{message}</div>;
```
because at least it says it returns JSX.Element

schoinh: Ooh I like that. I think I should make a separate issue/PR for that though, because it'd be best to go through all the function components (the lint fixes in this PR only touch those without any props) and make them all use it.

meganrm (approved):

PR #307, opened 2022-01-29
Add return type to React components

schoinh:
Problem
=======
Since #306 was merged, some React components are missing return types. 

Solution
========
* ~90% of the changes in this PR are just adding the return type `JSX.Element` to function components.
* ~9% of the changes are just removing `Record<string, never>` (implying no component state) from some of the class components because they're unnecessary.
* The other few changes are pointed out in the diff

## Type of change

* Maintenance (non-breaking change which should have no effect on functionality)

Steps to Verify:
----------------
1. `npm run typeCheck` still passes
2. App still runs like before

### DIFF
--- a/src/components/ErrorNotification/index.tsx
+++ b/src/components/ErrorNotification/index.tsx
@@ -11,12 +11,14 @@ interface ErrorNotificationProps {
     onClose?: () => void;
 }
 
+// Technically this function does not return a JSX.Element, but it is
+// like a React component
 const ErrorNotification = ({
     level,
     message,
     htmlData,
     onClose,
-}: ErrorNotificationProps): JSX.Element => {
+}: ErrorNotificationProps) => {
     // Sometimes the message that comes in at runtime is an entire
     // Error object instead of a string
     if (typeof message !== "string") {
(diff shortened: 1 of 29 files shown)


src/components/ErrorNotification/index.tsx
```diff
@@ -11,6 +11,8 @@ interface ErrorNotificationProps {
     onClose?: () => void;
 }
 
+// Technically this function does not return a JSX.Element, but it is
+// like a React component
```
schoinh, line 15: Anomalous change no. 1

src/components/ScrollToTop/index.tsx
```diff
@@ -9,7 +9,7 @@ Source: https://reactrouter.com/web/guides/scroll-restoration
 import { useEffect } from "react";
 import { useLocation } from "react-router-dom";
 
-const ScrollToTop = () => {
+const ScrollToTop = (): null => {
```
schoinh, line 12: Anomalous change no. 2 (expand the diff up to see the comment block about this "component" -- maybe I should just move this into src/index?)

toloudis (approved):

meganrm (approved):

PR #317, opened 2022-02-15
Fix/current time is hard to read when number of decimal places keeps changing

schoinh:
Problem
=======
The current time display in the playback controls is hard to read when the number of decimal places keeps changing:

Feb-15-2022 13-17-10

Solution
========
Left-align the current time value instead of centering it:

Feb-15-2022 13-20-15

@blairlyons brought this issue up again (issue has been mentioned before but we never discussed solutions) and this was the simplest solution I could think of.

## Type of change

* Bug fix (non-breaking change which fixes an issue) - Actually this is more like a design change, but it feels like a bug fix

### DIFF
--- a/src/components/PlaybackControls/style.css
+++ b/src/components/PlaybackControls/style.css
@@ -14,7 +14,6 @@
 
 .container :global(.ant-input-number input) {
     color: var(--dark-theme-btn-color);
-    text-align: center;
     padding: 0 3px;
 }
 


blairlyons (approved):

toloudis:
I wonder if it would be possible to keep the decimal point in the same spot...

toloudis (approved):

PR #323, opened 2022-02-18
Fix/can't use arrow keys in input box of Upload by URL modal

### LINKED ISSUE #283: Can't use arrow keys in "Load Model from URL" modal input
## Description

When inputting a URL into the "Load Model from URL" modal, I can't use left and right arrow keys to move my text cursor back and forth. I can't use home or end keys either. (I _can_ use my mouse to click on a different spot in the text input and move the cursor there.)

Maybe it's because we have some arrow keys mapped to things like zoom.

## Expected Behavior

My arrow keys should work! And ideally home and end keys too! (Delete and backspace work 🤷🏻‍♀️)

## Reproduction

1. Go to simularium.allencell.org and go to Load Model -> From a URL
2. Type something
3. Press left arrow, nothing happens, the blinking cursor in the text box doesn't move.

## Environment

MacOS Big Sur + Microsoft Edge/Chrome/Safari

schoinh:
Problem
=======
Closes #283 

Solution
========
This is an Antd bug. I was just about to file a bug report with them and then saw someone else already did earlier today! https://github.com/ant-design/ant-design/issues/34125

For now, I found a temporary workaround which is to avoid nesting the `Modal` inside `Menu.Item` but rather place it as a sibling to `Menu.Item`. This required taking some components out of the existing `UrlUpload` component, only leaving the `Modal` component in it. So I renamed `UrlUpload` to `UrlUploadModal`.

Bonus: I also made it so that the modal re-renders after closing and re-opening so that the previous input box contents don't stick around after closing.

## Type of change

Steps to Verify:
----------------
1. Try to repro the bug with steps outlined in issue, it should be fixed


src/components/UrlUploadModal/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const styles = require("./style.css");
+
+interface UrlUploadModalProps {
+    setIsModalVisible: (isModalVisible: boolean) => void;
+}
+
+const UrlUploadModal = ({
+    setIsModalVisible,
+}: UrlUploadModalProps): JSX.Element => {
```
schoinh, line 15: The only new code in this file

src/components/LoadFileMenu/index.tsx
```diff
@@ -74,8 +85,19 @@ const LoadFileMenu = ({
                 ))}
             </Menu.SubMenu>
             <Menu.Item key="url-upload">
-                <UrlUpload />
+                <Button type="ghost" onClick={showModal}>
+                    From a URL
+                </Button>
```
schoinh, line 90: This used to be inside UrlUpload

meganrm (approved):

toloudis (approved):

PR #332, opened 2022-03-22
fix horizontal scroll

meganrm:
Problem
=======
On chrome the landing page had horizontal scroll 

Solution
========
I first narrowed it down to the `max-width` setting on the text content, but it was just an arbitrary cutoff: If i set it to over 1040px, there was no scroll. 
So I finally figured out it was the markdown content, we had both column number and height set, which I guess caused an invisible overflow even though the text fit. 

## Type of change

Change summary:
---------------
* Moved text content into it's own div
* Experimented with changing the max width settings
* Removed the height setting
* Revert max width to original setting 

Steps to Verify:
----------------
1. npm start
2. open in chrome
3. should not be able to horizontally scroll on landing page

Screenshots (optional):
-----------------------
**Before**:
Screen Shot 2022-03-22 at 10 58 16 AM
NOTICE the text is longer in column 2
Screen Shot 2022-03-22 at 12 12 38 PM
**After**:
Screen Shot 2022-03-22 at 12 12 25 PM

### DIFF
--- a/src/components/LandingPage/index.tsx
+++ b/src/components/LandingPage/index.tsx
@@ -24,37 +24,43 @@ const LandingPage = (): JSX.Element => {
         <React.Fragment>
             <Content className={styles.content}>
                 <div className={styles.panel}>
-                    <h1>Simularium {BetaTag}</h1>
-                    <h2>
-                        Share, visualize, & interrogate biological simulations
-                        online
-                    </h2>
-                    <br />
-                    <p>
-                        The Simularium Viewer makes it easy to share and
-                        interrogate interactive 3D visualizations of biological
-                        simulation trajectories and related plots directly in a
-                        web browser. Its primary goal is to facilitate
-                        collaborations among experimental and computational
-                        biologists by removing major challenges to sharing,
-                        accessing, and comparing simulation results.
-                    </p>
+                    <div className={styles.textContent}>
+                        <h1>Simularium {BetaTag}</h1>
+                        <h2>
+                            Share, visualize, & interrogate biological
+                            simulations online
+                        </h2>
+                        <br />
+                        <p>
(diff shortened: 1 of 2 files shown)


src/components/LandingPage/index.tsx
```diff
@@ -24,37 +24,43 @@ const LandingPage = (): JSX.Element => {
...
-                    <p>
-                        The Simularium Viewer makes it easy to share and
-                        interrogate interactive 3D visualizations of biological
-                        simulation trajectories and related plots directly in a
-                        web browser. Its primary goal is to facilitate
-                        collaborations among experimental and computational
-                        biologists by removing major challenges to sharing,
-                        accessing, and comparing simulation results.
-                    </p>
+                    <div className={styles.textContent}>
```
meganrm, line 27: this looks like a lot of changes but all i did was add a div around the text content instead of styling with `.panel *` which I initially thought might be causing the problem, but left because it seems safer

src/components/LandingPage/style.css
```diff
@@ -85,7 +85,6 @@
 
 .markdown {
     column-count: 2;
-    height: 540px;
```
meganrm, line 88: this is the change that fixes the problem (see screen shots for before and after of the acknowledgments)

schoinh: That's what I would have done (I had made this issue for this bug but looks like it got buried - https://github.com/simularium/simularium-website/issues/329)

schoinh (approved):

src/components/LandingPage/style.css
meganrm: Oh man! I totally missed this, I'll be better about searching in the issues in the future. Thanks for reviewing!

PR #336, opened 2022-04-05
fix issue after merge main

meganrm:
Problem
=======
There were style changes on `main` I hadn't merge in that caused the new content to lose the `max-width` setting

Solution
========
Added a composes call to make sure the new panel gets the same styling. 
I also added a few mobile styling tweaks 

## Type of change

Screen Shot 2022-04-04 at 8 36 10 PM
Screen Shot 2022-04-04 at 8 44 58 PM

### DIFF
--- a/src/components/LandingPage/style.css
+++ b/src/components/LandingPage/style.css
@@ -79,12 +79,13 @@
 }
 
 .citation-content {
+    composes: text-content;
     display: flex;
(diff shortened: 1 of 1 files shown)


blairlyons (approved):

toloudis (approved):
LGTM?

PR #343, opened 2022-05-13
won't show time controls if there is no time range

meganrm:
Problem
=======
The cellPack files don't need playback controls when they're a single frame

Solution
========
Only show playback controls if the first time and last time are different 

## Type of change

Steps to Verify:
----------------
1. npm start
2. http://localhost:9001/viewer?trajUrl=https://www.dropbox.com/s/hhqrwp9bc8huufa/BloodPlasma.simularium?dl=0
3. Should not show a time bar
4. go to an example trajectory
5. time bar should be back

Screenshots (optional):
-----------------------
showing a cellpack file with no time bar 
Screen Shot 2022-05-12 at 5 17 16 PM

### DIFF
--- a/src/containers/ViewerPanel/index.tsx
+++ b/src/containers/ViewerPanel/index.tsx
@@ -392,22 +392,24 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
                         this.onTrajectoryFileInfoChanged
                     }
                 />
-                <PlaybackControls
-                    playHandler={this.startPlay}
-                    time={time}
-                    timeStep={timeStep}
-                    displayTimes={displayTimes}
-                    timeUnits={timeUnits}
-                    onTimeChange={this.skipToTime}
-                    pauseHandler={this.pause}
-                    prevHandler={this.playBackOne}
-                    nextHandler={this.playForwardOne}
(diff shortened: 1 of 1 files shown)


toloudis (approved):

src/containers/ViewerPanel/index.tsx
```diff
@@ -392,22 +392,24 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
-                    pauseHandler={this.pause}
-                    prevHandler={this.playBackOne}
-                    nextHandler={this.playForwardOne}
-                    isPlaying={isPlaying}
-                    firstFrameTime={firstFrameTime}
-                    lastFrameTime={lastFrameTime}
-                    loading={isBuffering}
-                    isEmpty={status === VIEWER_EMPTY}
-                />
+                {firstFrameTime !== lastFrameTime && (
```
toloudis, line 395: i wonder if there is a better check, like count of number of frames === 1

meganrm: yeah, this component doesn't know the total number, and I didn't investigate when/if that gets sent in. I don't see any way this would be wrong when the count would be right though

blairlyons (approved):

PR #364, opened 2022-09-19
Feature/update workflows

### LINKED ISSUE #352: Change nightly build to use main branch
## What needs to happen?
We need to switch the nightly build to use the main branch rather than the released version.

We use github actions to run nightly builds. The nightly builds are published using github-pages to public websites. We need to make sure that the "main" branch is being used to create those builds, so that it picks up any merged PRs as they come in.

We also should put the nightly build url someplace discoverable, such as the readme?

QUESTION:
Key detail is that there are situations when we would like simularium-website to use simularium-viewer's main branch as well, but sometimes that may cause breakage. This probably needs to be discussed - could we have two flavors of nightly for simularium-website?

## Why should we do this?
Allows us to use merged PRs when doing new dev work without a release. 

By end of sprint 7/26

meganrm:
Problem
=======
Closes #352 

Solution
========
Using github actions upload and download, I was able to build the app twice with two different builds and then upload them both to the gh-pages branch. I tested this on https://github.com/meganrm/react-redux-github-actions/actions/runs/3085116920 because I couldn't get this action to trigger while it's on a branch. 

I also used conditionals to remove the redundant workflow files

## Type of change


gh-pages/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Daily Builds with Updated Viewers</title>
+</head>
+<body>
+    <a href="https://simularium.github.io/simularium-website/dev">App with viewer at development head</a>
+    <a href="https://simularium.github.io/simularium-website/stable">App with latest released stable viewer</a>
```
toloudis, line 11: would relative urls "dev" and "stable" work here?

.github/workflows/ci-and-auto-deploy.yml
```diff
@@ -1,11 +1,8 @@
-name: Deploy AWS production
+name: Continuos Integration
```
toloudis, line 1: Continuous

gh-pages/index.html
meganrm: oh yeah they would

gh-pages/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Daily Builds with Updated Viewers</title>
+</head>
+<body>
+    <a href="/dev">App with viewer at development head</a>
+    <a href="/stable">App with latest released stable viewer</a>
```
toloudis, line 11: i'm not 100% sure it works with the leading / ... unless you tested it

toloudis (approved):
This is super cool and I can't wait to see it for real.  Thanks for doing this.

blairlyons (approved):

Issue #322, opened 2022-02-18
Automatically focus in input box when "Load by URL" modal opens


schoinh:
## Use Case

As a Simularium user, it would be more convenient/less annoying if my cursor automatically focused in the input box when the "Load by URL" modal opens. Currently, I have to click in the text input box to focus before I can paste my URL.

## Acceptance Criteria

When the user clicks on the "Load Model -> From a URL" menu item to open the Load by URL modal, they can hit ctrl/cmd + V immediately afterwards to paste a URL into the text input box, without having to click inside the input box to focus the cursor first.

## Details

There is an `autofocus` prop that can be added to an Antd `Input` component that _should_ work, but it doesn't. That might be a bug with Antd.

frasercl:
Resolved by #373

PR #373, opened 2022-10-27
Url Upload Auto Focus

BritneyAC:
Problem
=======
Solves #322
React, I believe, had a concurrent rendering issue causing 2 modals to be created, and the auto focus didn't know which to auto focus on

Solution
========
Moving the UrlUploadModal component to be on the same level as the menu instead of being rendered inside of the menu as the menu closes seems to fix the issue

Also added highlighting to the "Learn more" link when focused.

Steps to Verify:
----------------

Loading models still works and tab targeting to everything else still works.
Can't think of any other issues that would arise.

### DIFF
--- a/src/components/LoadFileMenu/index.tsx
+++ b/src/components/LoadFileMenu/index.tsx
@@ -89,12 +89,6 @@ const LoadFileMenu = ({
                     From a URL
                 </Button>
             </Menu.Item>
-            {/* 
-            1. Putting UrlUploadModal inside Menu.Item causes keyboard bugs.
-               https://github.com/ant-design/ant-design/issues/34125
-            2. Conditionally rendering the modal this way instead of as a `visible` prop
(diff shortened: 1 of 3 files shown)


toloudis:
Thank you for this contribution! We will be looking at it very soon

frasercl (approved):
Looks like an issue with antd. I don't know what implementation detail would cause it to create two copies of the same component, but at a glance it appears that children of `Menu` get chewed up by [this recursive function](https://github.com/ant-design/ant-design/blob/90b3823c9291f6b977dabc61fe93821365a5b961/components/menu/hooks/useItems.tsx#L37), which from the type signature is clearly hostile to anything other than menu items. Moving the modal outside the menu seems like a reasonable solution, and I was able to confirm it works. Good catch and thanks for this contribution!

toloudis (approved):
LGTM  assuming no regression 😄

PR #376, opened 2022-11-03
Fix/load viewer webworker

toloudis:
Problem
=======
simularium website was not able to load webworkers from simularium-viewer

Solution
========
rework the build for simularium-viewer, updating some of the tooling along the way, and then update build tooling for simularium-website.
This is a companion PR to https://github.com/simularium/simularium-viewer/pull/290 and will require a version bump to simularium-viewer to incorporate it.   It may well work with existing viewer but that has not been tested.

## Type of change


toloudis:
TODO: reconcile with Megan's work to update nodejs version?

toloudis:
just quickly smoke tested with viewer 3.4.0 and code seems to be working fine, so this does not fully depend on the new viewer PR mentioned above.

meganrm (approved):
Looks good and I've been able to run this on my computer as well. I would like more of an explanation of the Globals.d.ts, just for my own understanding of why we need that

toloudis:
> Looks good and I've been able to run this on my computer as well. I would like more of an explanation of the Globals.d.ts, just for my own understanding of why we need that

I think it was key for allowing `import` for css files instead of `require`, so that typescript's type checker can treat them as modules and reason about how the contents would be accessed in other ts code.  A byproduct of the toolchain updates, I believe...

PR #380, opened 2022-11-30
push builds to new buckets

meganrm:
Problem
=======
we want to have a build with both our dev version of the viewer (what's on main), and our more resent stable release. 

Solution
========
We created two additional aws buckets to publish the app the same way we do production and staging. 
I changed the gh-pages page to link to all 4 version of the app we have: 
https://simularium.github.io/simularium-website/

I ran all the actions manually from my computer to have the first set of releases uploaded

## Type of change

Steps to Verify:
----------------
1. go [the gh-page](https://simularium.github.io/simularium-website/)
2. click on the different links

### DIFF
--- a/.github/workflows/nightly-new-viewer.yml
+++ b/.github/workflows/nightly-new-viewer.yml
@@ -1,5 +1,6 @@
 name: Nightly builds
-
+env: 
+  NODE_VERSION: '16.x'
 on:
   schedule:
     - cron: "0 0 * * *"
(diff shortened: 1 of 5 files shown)


gh-pages/index.html
```diff
@@ -9,10 +9,16 @@
 <body>
     <ul>
         <li>
-            <a href="dev">App with viewer at development head</a>
+            <a href="https://nightly-dev.simularium.allencell.org"">Nightly Dev: App with viewer at development head</a>
```
toloudis, line 12: double quotes

.github/workflows/nightly-new-viewer.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.x"
+
+      - name: Cache dependencies
```
toloudis, line 20: can you explain what this does for us?  this might be a good devchat topic to understand how it works

meganrm: I'm not sure if I was the one who originally added this, but here is the github article about it. It just speeds up the run time as I understand it. 

https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

toloudis (approved):

ascibisz (approved):

PR #383, opened 2022-12-20
Make staging and prod configs more similar

toloudis:
remove hashedmoduleids plugin to see if it fixes drag n drop in prod.  Staging does not have the issue.

Also add some config output.
Also don't show webpack dev server overlay for warnings.

### DIFF
--- a/src/Globals.d.ts
+++ b/src/Globals.d.ts
@@ -4,3 +4,6 @@ declare module "*.css" {
     const classes: { readonly [key: string]: string };
     export default classes;
 }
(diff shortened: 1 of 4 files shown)


toloudis:
I would say this PR is worth merging and the only question is whether to reinstate the HashedModuleIdsPlugin (and make it the same config for staging and prod).   It seemed to not make a difference in the buggy behavior.  This is now a companion PR to https://github.com/simularium/simularium-viewer/pull/297

blairlyons (approved):

PR #392, opened 2023-02-04
File upload modal rework and styling

### LINKED ISSUE #385: New global styling for modals
## Use Case

Part of Implement production autoconversion UI #346

## Acceptance Criteria

Models look like they do in the linked design 

## Details

[Link to final designs](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/63079c1e-f6f9-44d8-a235-39e10d40acbe)
Design task https://github.com/simularium/simularium-website/issues/355

### LINKED ISSUE #388: Simularium file menu option: Implement new modal experience where tabs toggle local vs URL inputs
[link to design](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/42683ba6-2c6e-4133-80e1-17db2d8f8af6)

frasercl:
Problem
=======
Two modal design issues for autoconversion:
- #385: autoconversion design [specifies](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/63079c1e-f6f9-44d8-a235-39e10d40acbe/) new styling for modals to improve contrast
- #388: the local file loading option should be moved out of the menu and into a new tabbed interface in the existing url loading modal, making room for the "import other file type" option in the future, per [specification](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/42683ba6-2c6e-4133-80e1-17db2d8f8af6/)

Solution
========
### Changes to resolve #385:
- Where possible, make these styling changes in ant-vars.less
- For global modal styling that this file can't capture, create new `CustomModal` component. This wraps ant's `Modal` component and adds a little extra styling, to keep the look of modals consistent as the app gains more of them.

### Changes to resolve #388:
- Rename `UrlUploadModal` to `FileUploadModal`, give it tabs, style the tabs
- Split the form it used to contain into its own subcomponent ("`UrlUploadForm`") for clarity. Because the "Load" button is located within the modal's footer and is shared between both tabs, the parent modal remains responsible for submitting this form via the `useForm` hook.
- Rework the `LocalFileUpload` component. This component is no longer responsible for initiating requests (I moved request code to `FileUploadModal`, though there may be a more permanent and appropriate place it can go) but contains UI for previewing a list of files before upload. This component should now be prepared for reuse in autoconversion's [file upload step](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/24ac925a-ed87-496f-93e0-bfdcf19bebe5/).
- Remove option to upload files from "Load model" dropdown and rename remaining options as specified.

### Other changes
- Convert `LoadFileMenu`'s dropdown menu from deprecated component API to preferred JSON API. (See [antd's website](https://4x.ant.design/components/menu/#Usage-upgrade-after-4.20.0))

## Type of change

Screenshots:
-------------
image
image
image
image


src/components/FileUploadModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setIsModalVisible,
+    clearSimulariumFile,
+    loadLocalFile,
+    setViewerStatus,
+    setError,
+}) => {
+    const [openTab, setOpenTab] = useState("device");
+    const [noUrlInput, setNoUrlInput] = useState(true);
+    const [fileList, setFileList] = useState<RcFile[]>([]);
+    const [urlForm] = Form.useForm();
```
meganrm, line 40: I haven't used this before, it seems neat

src/components/FileUploadModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        setIsModalVisible(false);
+    };
+
+    const onUrlInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setNoUrlInput(event.target.value.length === 0);
+    };
+
+    const disableLoad =
+        openTab === "device" ? fileList.length === 0 : noUrlInput;
+
```
meganrm, line 52: this was immediately obvious to me, ie why when we're loading a local file we'd disable the loading when the file list is zero? Can you add a comment here to explain the logic?

meganrm: also, to make sure you're not getting out of sync, shouldn't you be accessing the openTab through `useEffect`?

src/components/FileUploadModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const onLoadClick = () => {
+        if (openTab === "device") {
```
meganrm, line 54: i would make this an enum and type these as the only two options

src/components/FileUploadModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const [noUrlInput, setNoUrlInput] = useState(true);
+    const [fileList, setFileList] = useState<RcFile[]>([]);
+    const [urlForm] = Form.useForm();
+
+    const closeModal = () => {
+        setIsModalVisible(false);
+    };
+
+    const onUrlInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setNoUrlInput(event.target.value.length === 0);
```
meganrm, line 47: so the form itself takes care of the value here? you're just keeping track of wether something has been entered?

src/components/FileUploadModal/index.tsx
frasercl: I also just learned about this. Seems worth keeping in mind for the rest of autoconversion work, which may require a similar pattern of splitting a single logical form across multiple child components.

src/components/FileUploadModal/index.tsx
frasercl: I believe the body of the render function is a safe place to access state, it's just handlers where that state has been changed that can give inconsistent values.

src/components/FileUploadModal/index.tsx
frasercl: Right, the only thing we care about at this level is whether or not any input is present, to decide whether or not to disable the Load button. This component could instead be directly responsible for the code that loads from a url (and perhaps should be for consistency, since loading from device is now handled at this level and not in a child), in which case it would want to store the whole string. But there's so little actual code involved that I don't think it's a huge deal either way.

src/components/FileUploadModal/upload-local-files.ts
```diff
@@ -15,104 +15,100 @@ import { VIEWER_ERROR } from "../../state/viewer/constants";
...
                 "Trajectory file was not found; please make sure it has a .simularium extension."
             );
         }
-        Promise.all<string | ISimulariumFile>(
-            filesArr.map((element, index) => {
+        const parsedFiles = await Promise.all<string | ISimulariumFile>(
+            selectedFiles.map((element, index) => {
                 if (index !== simulariumFileIndex) {
                     // is async call
                     return element.text();
```
ascibisz, line 77: This is kinda out of scope of this ticket, but do you know if there are limitations to how the max file size that `.text()` can handle? We do a similar thing in the simularium-viewer call and anecdotally it seemed to fail loading the file somewhere in the ~100MB range, which is unfortunately within the size range of what input files may be. I think it'd be out of scope for this ticket to address that, but it probably is within scope to ensure we're handling any errors that may come out of this

src/components/FileUploadModal/upload-local-files.ts
```diff
@@ -15,104 +15,100 @@ import { VIEWER_ERROR } from "../../state/viewer/constants";
...
             // We're at the last customRequest for this batch of files,
             // so reset counter
-            numCustomRequests = 0;
+            numRequests = 0;
         }
         return;
     }
 
     if (selectedFiles.length === 1) {
         // numCustomRequests and selectedFiles.length are both 1, so reset
```
ascibisz, line 53: nit: update this comment say `numRequests` instead of `numCustomRequests`

src/components/FileUploadModal/upload-local-files.ts
```diff
@@ -15,104 +15,100 @@ import { VIEWER_ERROR } from "../../state/viewer/constants";
...
     clearSimulariumFile({ newFile: true });
 
-    const filesArr: File[] = Array.from(selectedFiles) as File[];
-
     try {
         // Try to identify the simularium file.
         // Put all the other files as text based geoAssets.
-        const simulariumFileIndex = findIndex(filesArr, (file) =>
+        const simulariumFileIndex = findIndex(selectedFiles, (file) =>
             file.name.includes(".simularium")
```
frasercl, line 66: Should this be `endsWith`? Are there instances where the file extension would be something other than .simularium, or where a file would contain .simularium in its name before the extension?

src/components/FileUploadModal/upload-local-files.ts
frasercl: I didn't make a lot of substantive change to this function, so I don't think I considered this. It looks like the `try`/`catch` this is enclosed in ought to handle the case where the conversion errors out, though that doesn't keep it safe from freezes/crashes. I wasn't able to turn up a specified reason on Google for why the memory ceiling would be so low - for instance, strings [ought to work up to ~2GB](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#description) on reasonable systems.

src/components/FileUploadModal/upload-local-files.ts
meganrm: yeah it could be endsWith

ascibisz (approved):

PR #397, opened 2023-02-17
add a default collapsed

meganrm:
Problem
=======
On mobile the app is pretty unusable because the panels are open by default and overlap each other. 

Blair has QR codes on her poster 

Solution
========
The `collapsed` state on the side panels is initialized based on the window size. This means that it's only set on load, it's not set as a listeners. So if a user did the strange thing of loading the app on their computer, and then make the browser window tiny, it won't auto close the panels unless they re-load at the smaller size

## Type of change

Steps to Verify:
----------------
1.`npm start`
2. make the screen mobile sized and reload
3. the panels should initialize to closed 
4. make the screen normal screen sized and reload
5. the panels should initialize to open

Screenshots (optional):
-----------------------
Screen Shot 2023-02-16 at 4 12 25 PM

### DIFF
--- a/src/components/SideBar/index.tsx
+++ b/src/components/SideBar/index.tsx
@@ -9,6 +9,7 @@ const { Sider } = Layout;
 import styles from "./style.css";
 
 interface SiderProps {
+    defaultCollapsed: boolean;
     type: string;
     onCollapse: (open: boolean) => void;
 }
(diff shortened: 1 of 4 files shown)


blairlyons (approved):

meganrm:
Feb-16-2023 17-08-21

toloudis (approved):
LGTM

PR #403, opened 2023-03-28
Feature/add conversion form overlay

### LINKED ISSUE #395: [Auto-conversion]-Create initial conversion form fields
3 steps: choose engine, choose file, enter optional data
[Process starts here](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/8449daa7-a755-4ab9-8a9d-8a547efdff60/)

- [x] Set up a ConversionFormOverlay component that is similar to ViewerOverlayTarget in that it returns null if the viewer isn't importing anything. Use similar styling to make it full screen. 
- [x] Use the VIEWER_IMPORTING state to conditionally make the appropriate  buttons in the header disabled
- [x] Use antd components to add the file import and select components

**Acceptance criteria:**
Clicking "import other file" opens a full screen overlay that takes the user through 3 steps. 
Close button does not yet need to work 
Submit button does not yet need to work 
The user can select an engine from the drop down menu. When they do it requests the required template from simulariumio 
The user can select a file to convert
The enter optional data form does not need to be functional, just the title

interim17:
Problem
=======
Continuing work on autoconversion UI feature.

Closes #395 

Solution
========
with @meganrm 

Change summary:
---------------
* ConversionForm component has been added
* Component has antd Select and Upload components for choosing file type and grabbing a local file
* Component renders through new import route
* Change in viewer state disables button in app header
* Light theme has been added to components folder and redundant styling was removed from modal specific css files
* Added Redux Logic to request template from simulariumio when simulation engine is selected
* Added selected engine template and selected file to state

Steps to Verify:
----------------
1. npm start
2. click Load Model button in header
3. Select 'import other file type'
4. Behold select and upload components

-----------------------
1. changes to a few files
2. to work with new import route changes were made to src/index.tsx , would appreciate extra scrutiny here to make sure i am not causing side effects


src/containers/Simularium/index.tsx
```diff
@@ -217,10 +218,11 @@ class App extends React.Component<AppProps, AppState> {
                             setViewerStatus={setViewerStatus}
                             setError={setError}
                         />
-                        <SideBar
-                            onCollapse={this.onPanelCollapse}
-                            type="left"
-                        >
+                        <ConversionFormOverlay
+                            className="conversionform"
+                            isLoading={false}
```
meganrm, line 223: this should be `isImporting` and should be using the `VIEWER_IMPORTING` value in state

src/components/LoadFileMenu/index.tsx
```diff
@@ -62,6 +69,11 @@ const LoadFileMenu = ({
         }
     };
 
+    let isDisabled = false;
+    if (viewerStatus == VIEWER_IMPORTING || isBuffering) {
+        isDisabled = true;
+    }
```
meganrm, line 75:
```suggestion
    const isDisabled = (viewerStatus === VIEWER_IMPORTING || isBuffering) 

```

src/containers/Simularium/index.tsx
meganrm: or it could just be called `active` too for simplicity

src/components/ConversionFormOverlay/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    );
+
+    if (showTarget) {
+        return conversionFormOverlay;
+    } else if (isLoading) {
+        // return loadingOverlay;
+        return null;
+    } else {
+        return null;
+    }
```
meganrm, line 140:
```suggestion
    if (active) {
        return conversionFormOverlay;
    } else {
        return null;
    }
```

meganrm: or `if (isImporting) `

src/components/ConversionFormOverlay/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <div className={styles.container}>
+            <h3 className={styles.title} style={{ fontSize: 30 }}>
+                Import a non-native file type
+            </h3>
+            <h3>
+                Convert and import a non-simularium file by providing the
+                following information
+            </h3>
+            <h3> Provide file information (required) </h3>
+            <h3 style={{ paddingBottom: 0 }}>Simulation Engine</h3>
```
meganrm, line 67: avoid using inline style as much as possible. use a className here and apply the style in the style sheet

src/containers/Simularium/index.tsx
interim17: yes, i left this static for ease during development, will change

src/components/CustomModal/light-theme.css
meganrm, line 1: this file shouldn't be here. Try it in `styles/` but make sure it still works

src/components/CustomModal/light-theme.css
```diff
@@ -0,0 +1,19 @@
...
+.light-theme :global(.ant-btn-default) {
+    background-color: var(--light-button-default-background);
+    border-color: var(--light-theme-modal-btn-default-color);
+    color: var(--light-button-default-color);
+}
+
+.light-theme :global(.ant-btn-primary) {
+    background-color: var(--light-theme-btn-primary-bg);
+    border: none;
+    color: #d8d8d8;
```
meganrm, line 11: these still need to be updated with the variables instead of hard coded

src/components/CustomModal/style.css
```diff
@@ -1,8 +1,8 @@
 /* Default and primary button colors differ within modals */
 .modal :global(.ant-btn-default) {
-    background-color: transparent;
+    background-color: var(--light-button-default-background);
     border-color: var(--light-theme-modal-btn-default-color);
-    color: var(--light-theme-modal-btn-default-color);
+    color: var(--light-button-default-color);
```
meganrm, line 5: these button styles are now redundant

src/components/LoadFileMenu/index.tsx
```diff
@@ -11,22 +11,29 @@ import {
...
     RequestNetworkFileAction,
 } from "../../state/trajectory/types";
+import { ViewerStatus } from "../../state/viewer/types";
+import { getStatus } from "../../state/viewer/selectors";
+import { State } from "../../state/types";
 import { TrajectoryDisplayData } from "../../constants/interfaces";
 import { VIEWER_PATHNAME } from "../../routes";
-import FileUploadModal from "../FileUploadModal";
+import LocalFileUpload from "../LocalFileUpload";
+// import UrlUploadModal from "../UrlUploadModal";
```
meganrm, line 20: remove unused imports

src/components/LoadFileMenu/index.tsx
```diff
@@ -60,44 +69,62 @@ const LoadFileMenu = ({
...
+                    From a URL
+                </Button>
+            </Menu.Item>
+            <Menu.Item key="local-file-upload">
+                {/* <LocalFileUpload
+                    clearSimulariumFile={clearSimulariumFile}
+                    loadLocalFile={loadLocalFile}
+                    setViewerStatus={setViewerStatus}
+                    setError={setError}
+                /> */}
```
meganrm, line 111: remove commented out code

src/components/CustomModal/style.css
interim17: so we should import the light theme into the CustomModal component and delete modal button stylings entirely

meganrm: yeah, but not the other styling that's specific to modals

src/components/ConversionFormOverlay/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+                <h3 className={styles.title}>Import a non-native file type</h3>
+                <h3>
+                    Convert and import a non-simularium file by providing the
+                    following information
+                </h3>
+                <h3 className={styles.provide}>
+                    {" "}
+                    Provide file information (required){" "}
+                </h3>
+                <h3 className={styles.selecttitle}>Simulation Engine</h3>
```
meganrm, line 35: nit: selecttitle should be camel case, ie styles.selectTitle

src/components/ConversionFormOverlay/style.css
```diff
@@ -0,0 +1,56 @@
...
+
+.upload-container {
+    display: flex;
+    flex-direction: row;
+}
+
+.selector-box {
+    width: 200px;
+    margin-right: 20px;
+}
```
meganrm, line 28: this class doesn't seem to be used?

src/components/ConversionFormOverlay/style.css
```diff
@@ -0,0 +1,56 @@
...
+
+.container button {
+    margin-right: 20px;
+    min-width: 80px;
+}
+
+.container span svg path {
+    fill: black
+}
```
meganrm, line 37: add a note that this is for the dropdown arrow since this is a little odd otherwise. It should also be our theme dark text color, not just hardcoded black

src/components/ConversionFormOverlay/style.css
interim17: it sizes the Select element for choosing simulation engine

src/components/ConversionFormOverlay/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    {" "}
+                    Provide file information (required){" "}
+                </h3>
+                <h3 className={styles.selectTitle}>Simulation Engine</h3>
+                <div className={styles.uploadContainer}>
+                    <Select
+                        className={styles.selectorBox}
+                        bordered={true}
+                        defaultValue="Select"
+                        options={selectOptions}
```
meganrm, line 41: we are going to set this in redux state in order to request the appropriate template

src/components/LoadFileMenu/index.tsx
```diff
@@ -89,24 +93,32 @@ const LoadFileMenu = ({
...
             key: "file-convert",
             label: (
-                <Button type="ghost" onClick={initializeFileConversion}>
+                <Link
+                    type="ghost"
+                    to={{
+                        pathname: IMPORT_PATHNAME,
+                    }}
+                    onClick={() => {
+                        initializeFileConversion();
```
toloudis, line 107: naming: `initializeFileConversion` sounds like it's about to convert something. Consider maybe `initializeFileConversionUI`  or `initializeFileConversionForm`?  kinda nitpicky not a huge deal.

src/constants/index.ts
```diff
@@ -29,6 +33,10 @@ export const MCELL_URL = "https://mcell.org/";
 export const SMOLDYN_URL = "http://www.smoldyn.org/";
 export const SPRINGSALAD_URL = "https://vcell.org/ssalad";
 export const MEDYAN_URL = "http://medyan.org/";
+export const UI_TEMPLATE_URL_ROOT =
+    "https://api.github.com/repos/simularium/simulariumio/contents/ui-templates";
+export const UI_TEMPLATE_DOWNLOAD_URL_ROOT =
+    "https://raw.githubusercontent.com/simularium/simulariumio/main/ui-templates";
```
toloudis, line 39: is github considered the production location for this data?  that seems slightly sketchy. I wonder if we could pre-fetch them as part of the build process here so they live with the app?  It's ok if this is out of scope for this PR but worth thinking about.

src/containers/ConversionForm/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+import customRequest from "./custom-request";
+import { SetErrorAction } from "../../state/viewer/types";
+import { UploadFile } from "antd/lib/upload";
+
+interface ConversionProps {
+    setConversionEngine: ActionCreator<SetConversionEngineAction>;
+    conversionProcessingData: {
+        template: Template;
+        templateMap: TemplateMap;
+        preConvertedFile: string;
```
toloudis, line 27: consider renaming to `fileToConvert` to match the `receiveFiletoConvert` below.

src/state/trajectory/conversion-data-types.ts
```diff
@@ -0,0 +1,40 @@
+export enum AvailableEngines {
+    Cytosim = "cytosim",
+    CellPack = "cellPACK",
+    Smoldyn = "Smoldyn",
+    SpringSalad = "SpringSaLaD",
+}
+
+export interface CustomType {
+    "python::module": string;
+    "python::object": string;
```
toloudis, line 10: I can't remember why these should be named `python::module` when they could be a more js friendly `pythonModule` or `pythonModuleName`

src/state/trajectory/logics.ts
```diff
@@ -334,21 +343,96 @@ const loadFileViaUrl = createLogic({
...
+            action,
+            uiTemplateUrlRoot,
+            uiBaseTypes,
+            uiCustomTypes,
+            uiTemplateDownloadUrlRoot,
+        } = deps;
+        const baseTypes = await httpClient
+            .get(`${uiTemplateDownloadUrlRoot}/${uiBaseTypes}`)
+            .then((baseTypesReturn: AxiosResponse) => {
+                console.log(baseTypesReturn);
```
toloudis, line 374: are these console logs needed?

src/state/trajectory/logics.ts
```diff
@@ -334,21 +343,96 @@ const loadFileViaUrl = createLogic({
...
+        return {
+            engineType: action.payload,
+            template: engineTemplate[templateName],
+            templateMap: typeMap,
+        };
+    },
+    processOptions: {
+        dispatchReturn: true,
+        successType: SET_CONVERSION_TEMPLATE,
+        // failType: FETCH_POLLS_FAILED, // dispatch this failed action type
```
toloudis, line 425: can this line be removed or maybe add a comment about why it should stay if so

src/styles/colors.css
```diff
@@ -68,7 +70,15 @@
...
     --light-theme-btn-default-disabled-bg: var(--heather);
+
+    --light-theme-btn-primary-color: var(--white-three);
+    --light-theme-btn-primary-disabled-color: var(--true-gray);
+    --light-theme-btn-primary-bg: var(--dark-four);
+    --light-theme-btn-primary-disabled-bg: var(--heather);
+    
+    
 }
+/* background color and color for deafult primary and primary disabled*/
```
toloudis, line 83: `deafult` --> `default`

src/constants/index.ts
interim17: Definitely something I could look into, I'll leave those changes for a different PR but follow up.

src/state/trajectory/conversion-data-types.ts
interim17: I'm not sure, but this is the convention throughout the ui-templates in simulariumio. It seems like we would want our typing to match the templates, whatever we go with.

src/state/trajectory/logics.ts
interim17: the ones we added no, removing them now. there is one console log from a different commit in a catch block that i will leave because it seems like it might be important.

src/state/trajectory/logics.ts
interim17: not sure about this one. @meganrm?

src/containers/ConversionForm/index.tsx
interim17: point taken its a bit vague, this object is both file and template though, can try to think of a more semantic name

src/components/LoadFileMenu/index.tsx
interim17: got it, went with initializeFileConversionUI

src/constants/index.ts
meganrm: yeah, that's been the plan, so they can stay in sync with the file format on simulariumio. I do like the idea of fetching them in the build process though

src/state/trajectory/logics.ts
meganrm: yeah. Oh yeah, I think we have a generic fail type you can set here

src/state/trajectory/logics.ts
```diff
@@ -334,21 +346,93 @@ const loadFileViaUrl = createLogic({
...
+                                .get(ref.download_url)
+                                .then((file) => file.data)
+                    )
+                )
+            );
+
+        const initTypeMap: TemplateMap = {};
+
+        const typeMap: TemplateMap = reduce(
+            customTypes,
```
frasercl, line 400: Might prefer native `map`/`reduce` here, unless there's a general preference in Simularium to use lodash. (The method syntax could also help to unwind the formatting of the `map` on 387 a bit)

src/components/theme/light-theme.css
```diff
@@ -0,0 +1,19 @@
+/* Default and primary button colors differ within modals */
+.light-theme :global(.ant-btn-default) {
+    background-color: var(--light-button-default-background);
+    border-color: var(--light-theme-modal-btn-default-color);
+    color: var(--light-button-default-color);
+}
```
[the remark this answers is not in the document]
meganrm: Yeah, so our app is mostly dark themed. So these are overrides of the default theme settings to get things like the modals and overlays to have consistent styling instead of having to style each component every time we want a light component. Do you have a suggestion for a better name?

src/containers/ConversionForm/style.css
```diff
@@ -0,0 +1,56 @@
+.container{
```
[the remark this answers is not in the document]
meganrm: I told him to do do: it has been our general convention for the top level selector in a file

src/components/LoadFileMenu/style.css
```diff
@@ -30,4 +30,4 @@
     width: fit-content;
     right: 300px !important;
     left: auto !important;
```
meganrm, line 32: it would be good if you could make this selector more specific so you can avoid using the important flags

src/components/LoadFileMenu/index.tsx
```diff
@@ -89,24 +93,32 @@ const LoadFileMenu = ({
...
                 </Button>
-            )
+            ),
         },
         {
             key: "file-convert",
             label: (
-                <Button type="ghost" onClick={initializeFileConversion}>
+                <Link
+                    type="ghost"
```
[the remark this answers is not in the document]
interim17: good catch, was redundant

src/components/LoadFileMenu/index.tsx
```diff
@@ -89,24 +93,32 @@ const LoadFileMenu = ({
...
+        },
     ];
 
     return (
         <>
             <Dropdown
                 menu={{ items, theme: "dark", className: styles.menu }}
                 placement="bottomRight"
-                disabled={isBuffering}
+                disabled={isBuffering || viewerStatus === VIEWER_IMPORTING}
```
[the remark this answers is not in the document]
interim17: thank you. so i fixed this just now by adding this code 
`else if (location.pathname === VIEWER_PATHNAME) {
            batch(() => {
                dispatch(setIsPlaying(false));
                dispatch(setStatus({ status: VIEWER_EMPTY }));
            });`
in src/index.tsx where the BrowserRouter stuff goes down. Posting here to check if my intuition is right that isPlaying should be false when the viewer is open but empty. This bug only seemed to happen when navigating to /viewer without a simulation loaded

src/components/LoadFileMenu/index.tsx
```diff
@@ -89,24 +93,32 @@ const LoadFileMenu = ({
                 <Button type="ghost" onClick={showModal}>
                     Simularium file
                 </Button>
```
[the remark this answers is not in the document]
interim17: yes that looks off

src/containers/ConversionForm/style.css
```diff
@@ -0,0 +1,56 @@
...
+
+.container button {
+    margin-right: 20px;
+    min-width: 80px;
+}
+
+.container span svg path {
+    /* this makes the down arrow in the select box visible  */
+    fill: var(--light-theme-btn-default-color);
+}
```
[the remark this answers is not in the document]
interim17: thank you, swapped it out, not urgent but curious why you prefer color to fill

src/containers/ConversionForm/style.css
```diff
@@ -0,0 +1,56 @@
...
+    font-weight: 600;
+    letter-spacing: .05rem;
+}
+
+.provide {
+    padding-top: 30px;
+    padding-bottom: 15px;
+    font-weight: 600;
+    font-size: 20px;
+}
```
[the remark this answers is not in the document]
interim17: good call, made the change

src/containers/ConversionForm/index.tsx
toloudis: I only meant `preConvertedFile` --> `fileContentsToConvert` or `fileToConvert` or something like that.

src/components/LoadFileMenu/style.css
interim17: I'm not sure what these style rules are even doing. When I remove them I can't tell that anything has changed about the pop up menu.

Might delete?

src/containers/ConversionForm/index.tsx
interim17: oops got it, thanks for clarifying, will change to fileToConvert

src/state/trajectory/logics.ts
meganrm: I think these are used here because the data coming in is actually an object instead of array

src/containers/ConversionForm/style.css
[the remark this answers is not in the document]
frasercl: ah gotcha

src/containers/ConversionForm/style.css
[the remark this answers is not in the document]
frasercl: In general, if an svg is acting as an icon or image, I'd consider it best practice to treat it as a static image or character rather than modifying its internal structure if you have the choice. In this case the path in the svg has the attribute `fill="currentColor"`, which explicitly defers to the element's CSS text color.

src/state/trajectory/logics.ts
frasercl: ah, that makes sense - I don't use lodash enough to consistently remember that it lets you do that

src/containers/ConversionForm/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        showDownloadIcon: false,
+                        showRemoveIcon: true,
+                    }}
+                    onChange={({ file }) => {
+                        setFileToConvert(file);
+                    }}
+                    customRequest={(options) =>
+                        customRequest(fileToConvert, receiveFileToConvert, setError, options)
+                    }
+                >
```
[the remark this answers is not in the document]
interim17: After discussing with @meganrm I think the plan is to leave this they way it is for now, but I will definitely look for opportunities to implement this component in the future, and we should pair on other stuff you created for this repo so I can build on what you've made.

src/components/LoadFileMenu/style.css
interim17: Removed.

src/components/theme/light-theme.css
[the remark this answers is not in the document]
frasercl: "theme" to me suggests a global, swappable color preset - e.g. the way vscode uses "light theme" vs "dark theme." But I can't really think of a better word for this kind of distinction between component styles, so if it's been around in Simularium a while anyways I don't have a problem leaving it.

src/components/LoadFileMenu/index.tsx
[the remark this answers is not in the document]
frasercl: Unconditionally setting the state to empty does make me a bit nervous, but it looks like that gets immediately overridden when the URL gives us something to load, and I'm confident enough that that will catch cases when the viewer shouldn't actually be empty.

PR #412, opened 2023-04-21
Draft for comments/feedback

meganrm:


src/components/ConversionFormFromViewer/BaseInput.tsx
meganrm, line 1: you should just call this `ConversionForm`

src/components/ConversionFormFromViewer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+} from "../../state/trajectory/conversion-data-types";
+
+const { Panel } = Collapse;
+
+interface InputFormProps {
+    // template: { [key: string]: any };
+    // templateData: { [key: string]: any };
+    // type: string;
+    // submitFile: (data: any) => void;
+    // onReturned: () => void;
```
meganrm, line 24: just delete these

src/components/ConversionFormFromViewer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+        //         } else {
+        //             thisValue = { ...currentState };
+        //             currentState = currentState[array[i + 1]] || {};
+        //         }
+        //         tempObject = tempObject[nestedKey] = thisValue;
+        //     });
+        // }
+        // this.setState(newState);
+        //calling to remove linting error
+        this.handleChange(path, key, value);
```
meganrm, line 69: this looks like it would create an infinite loop?

src/components/ConversionFormFromViewer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+                        {map(
+                            parameters,
+                            (
+                                parameter: CustomParameter,
+                                key: string
+                            ): JSX.Element | null => {
+                                const dataType = parameter.data_type;
+                                if (
+                                    dataType === "file" ||
+                                    dataType === "meta_data"
```
meganrm, line 111: why did you do this?

src/components/ConversionFormFromViewer/style.css
```diff
@@ -0,0 +1,106 @@
...
+}
+.ant-collapse-content > .ant-collapse-content-box {
+    padding: 0px;
+}
+
+:global(.ant-form-item) {
+    padding: 3 3 3 3;
+}
+
+:global(.ant-form-horizontal) {
```
meganrm, line 31: anything with a :global selector should be wrapped inside another class selector. Because this will style all forms, which is not what we want. The idea of using css modules is so you know the style you write in the single document will only effect the components you import the file into

src/style.css
```diff
@@ -1,5 +1,14 @@
 /* global styles
 css modules not applied */
+html,
+body {
+    height: 100%;
+    width: 100%;
+    margin: 0;
+    padding: 0;
+    overflow: hidden;
```
meganrm, line 9: why did you make this change?

meganrm: This makes it so the main page can't scroll

src/components/ConversionFormFromViewer/style.css
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.form-div :global(.ant-select) {
+    width: 200px;
+}
+
+:global(.ant-collapse) {
+    background-color: var(--overlay-purple);
+}
+.ant-collapse-content > .ant-collapse-content-box {
```
meganrm, line 23: this won't work without the :global selector

src/components/ConversionFormFromViewer/style.css
```diff
@@ -0,0 +1,106 @@
...
+    padding-left: 0px !important;
+}
+
+.collapse {
+    width: 82vw;
+    max-width: 1000px;
+    border: none;
+}
+
+.formz {
```
meganrm, line 98: typo?

src/containers/ConversionForm/style.css
```diff
@@ -37,20 +43,60 @@
...
 .select-title {
     margin-bottom: 0px;
+    font-size: 16px;
+}
+
+:global(.ant-input) {
+    width: 175px;
+}
+
+.my-form {
```
meganrm, line 72: use a clearer name

src/components/ConversionFormFromViewer/style.css
```diff
@@ -0,0 +1,106 @@
...
+:global(.ant-form-item) {
+    padding: 3 3 3 3;
+}
+
+:global(.ant-form-horizontal) {
+    padding: 3 3 3 3;
+}
+
+input,
+select {
```
meganrm, line 36: don't use selectors like this. Should always be using class selectors to wrap elements if you need to style base elements

src/components/ConversionFormFromViewer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    handleSubmit(event: any) {
+        // event.preventDefault();
+        // const payload = {
+        //     ...this.state,
+        // };
+        // console.log("submitting", payload);
+        // this.props.submitFile(payload)
+        //calling to remove linting error
+        this.handleSubmit(event);
```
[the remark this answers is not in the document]
interim17: ok got it, yes was just trying to avoid pushing anything with linter erors

src/components/ConversionFormFromViewer/index.tsx
interim17: not sure which part you are referring to. 

expandIcon stuff is just to style the arrows on the collapse menu

excluding "file and "meta_data" was just to try and work around template issues, can remove as per message earlier saying I should assume template willl be fixed

src/style.css
interim17: dang, ok. i was trying to solve a few issues: 
multiple overflow scroll bars appearing, unsightly and hard to navigate
buttons on the bottom of the page being cut off
a white border appearing at bottom and right of overlay

src/containers/ConversionForm/index.tsx
```diff
@@ -80,18 +91,71 @@ const ConversionForm = ({
...
+                            header=""
+                            showArrow={false}
+                            collapsible="disabled"
+                        ></Panel>
+                    </Collapse>
+                </div>
+            ) : null}
+            {conversionProcessingData.templateMap ? (
+                <InputForm
+                    conversionProcessingData={conversionProcessingData}
```
meganrm, line 153: you already have conditional rendering in this form. And there are multiple conditional renders in here which makes the logic harder to follow. I'm going to move all the things that only should be rendering after you upload a file into the form component

src/containers/ConversionForm/index.tsx
```diff
@@ -80,18 +91,71 @@ const ConversionForm = ({
...
+                    collapsible="disabled"
+                ></Panel>
+            </Collapse>
+            {fileToConvert ? (
+                <div>
+                    <h3 className={styles.provide}>
+                        Provide display information (optional)
+                    </h3>
+                    <h3 className={styles.convertText}>
+                        You can import your model now with defaults, or specify
```
meganrm, line 125: you shouldn't use <h3> for all different text sizes. Use the semantically appropriate tag

src/style.css
meganrm: interesting. I haven't seem the multiple overflow scroll bars, but that would be concerning, I'm curious to see what it looks like on your computer

src/components/ConversionFormFromViewer/index.tsx
meganrm: yeah ok, that's what I figured

PR #417, opened 2023-04-26
looping playback and tooltip styling

interim17:
Problem:
=======
Users want to be able to loop simulation playback, tooltips across simularium have different styling.

Closes https://github.com/simularium/simularium-website/issues/298

NOTE: Initial feature branch was mistakenly branched from auto-conversion feature branch instead of main, leading to git mess. Deleted that and recreated this branch so commit history is compressed into one commit.

Solution:
=======
New feature (non-breaking change which adds functionality)

Gives users the ability to loop simulation playback, as well as setting primary global styling for tooltips.

Looping button was added to UI, clicking the button toggles a new piece of state in the viewer branch. Playback functions, icons, and tooltips respond to that state.

Styles were added/removed to make tooltips across the application have matching styling.

Steps to Verify:
=======
Load a simulation, find and click the looping icon in the playback controls, run a simulation and see if its starts over when it reaches the last frame.

Check styling on tooltips by hovering cursor over left panel checkboxes, playback controls, and home icon. Should match [design.](https://xd.adobe.com/view/75b6a3c8-f650-443b-bccc-d1fcd134a35e-20e5/screen/324d327c-ade5-4765-8c66-aa61db1feb4f/specs/)

Keyfiles:
=======
- UI elements added to PlaybackControls/index.tsx
- Looping state changes the standard suite of Redux files.
- ViewerPanel/index.tsx holds function toggleLooping to toggle looping state, and function receiveTimeChange which has the logic for looping playback.

### DIFF
--- a/src/components/CheckBoxTree/style.css
+++ b/src/components/CheckBoxTree/style.css
@@ -89,13 +89,3 @@
 .container :global(.ant-typography-ellipsis) {
     color: var(--dark-theme-sidebar-text)
 }
-
-/* Tooltip styling  */
-/* Has to be global because it's at the bottom of the dom */
-:global(.ant-tooltip-inner){
-    background-color: var(--side-panel-tooltip);
-}
-
-:global(.ant-tooltip-arrow-content){
-    background-color:  var(--side-panel-tooltip);
-}
--- a/src/containers/ViewerPanel/index.tsx
+++ b/src/containers/ViewerPanel/index.tsx
@@ -67,6 +67,7 @@ interface ViewerPanelProps {
     displayTimes: DisplayTimes;
     timeUnits: TimeUnits;
     isPlaying: boolean;
+    isLooping: boolean;
     fileIsDraggedOver: boolean;
     status: string;
     numFrames: number;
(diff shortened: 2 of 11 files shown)


src/components/CheckBoxTree/style.css
interim17, line 1: now redundant, removed

interim17:
This feature works pretty well as it is, but I discovered a bug today related to the playback slider. Bug is in production branch but has side effects on the looping feature.

Looking into it now and hoping I can resolve easily and fold into this PR, may make a second issue if need be.

Error results from selecting points on the playback slider during playback.

interim17:
The bug seems to be mostly on the backend.

When frames are selected during playback (either by using the slider, or when looping back to the first frame) there are issues with data formatting and floats arriving when integers are expecting. Not something I can really fix on my own, so I'm going to assume we will make a separate issue/ticket for this.

My question: since this bug affects this feature, should this PR continue to go through review or maybe be changed into a draft and postpone review until the bug is ironed out?

src/containers/ViewerPanel/index.tsx
```diff
@@ -239,6 +242,15 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         setIsPlaying(false);
     }
 
+    public toggleLooping() {
+        const { isLooping, setIsLooping } = this.props;
+        if (isLooping) {
+            setIsLooping(false);
+        } else {
+            setIsLooping(true);
+        }
```
frasercl, line 251: This if could just become `setIsLooping(!isLooping)`

src/containers/ViewerPanel/index.tsx
```diff
@@ -270,6 +282,8 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
             timeStep,
             receiveTrajectory,
             setBuffering,
+            setIsPlaying,
```
frasercl, line 285: Linter says this value is unused and can be deleted

frasercl (approved):
LGTM apart from some minor polish

src/containers/ViewerPanel/index.tsx
interim17: Good catch, ty

Issue #295, opened 2022-01-05
Implement Export Data (Download)


lynwilhelm:
## Use Case
Users need ability to download a trajectory

## Acceptance Criteria
Add download "button" to UI
User agent handles the download

## Details
Subsequent downloads filename will end in (1), (2), (3), etc (not sure if this is standard behavior across all browsers, but this is important so users can save differentiate versions as they apply changes)

Design Link: https://xd.adobe.com/view/75b6a3c8-f650-443b-bccc-d1fcd134a35e-20e5/

More details about how to implement with the backend functionality TBD [here](https://github.com/simularium/octopus/issues/27)

blairlyons:
@lynwilhelm is there a design link ready for this one?

lynwilhelm:
Design link added. Thanks, @blairlyons

interim17:
What user changes should be captured in the downloaded trajectory?

interim17:
@lynwilhelm @blairlyons 
I have a working V1 of this but my hiccup is understanding the bit about capturing user changes, and file locations for fetching networked files. Even if my current approach works it will need updating in relation to [Alli's PR ](https://github.com/simularium/octopus/pull/35)when we switch to Octopus.

In my current version a local Simularium file or a networked file stored in our AWS data bucket can be downloaded. Blair, some gaps in my understanding about user-changes and file-locations might be suited for a Simularium walkthrough part two?

PR #422, opened 2023-06-01
Feature/download trajectory

interim17:
Problem
=======
Users need to be able to download their trajectories.

Closes [#295](https://github.com/simularium/simularium-website/issues/295)

Solution
========
Basic UI button and functionality is here to download networked files from our data bucket, as well as local files. 

Ability to download local files is redundant now, but will become relevant in future as we capture user changes in evolving Simularium file type. Recent updates to viewer have provided a method for getting blobs from JSON files, matching that functionality for binary files is forthcoming.

with @meganrm and @toloudis 

## Type of change

Steps to Verify:
----------------
1. Locate download button in top right of header.
2. Load a networked trajectory from examples, download it.
3. Load a local file, maybe the one you just downloaded, download it.

### DIFF
--- a/src/components/DownloadTrajectoryMenu/index.tsx
+++ b/src/components/DownloadTrajectoryMenu/index.tsx
@@ -1,47 +1,21 @@
-import React, { useState } from "react";
-import { Link, useLocation, useHistory } from "react-router-dom";
-import { ActionCreator } from "redux";
-import { Dropdown, Button, MenuProps, Tooltip } from "antd";
+import React from "react";
+import { Button, Tooltip } from "antd";
 
-import TRAJECTORIES from "../../constants/networked-trajectories";
-import { TOOLTIP_COLOR, URL_PARAM_KEY_FILE_NAME } from "../../constants";
-import {
-    ClearSimFileDataAction,
-    NetworkedSimFile,
-    RequestLocalFileAction,
-    RequestNetworkFileAction,
-} from "../../state/trajectory/types";
-import { TrajectoryDisplayData } from "../../constants/interfaces";
-import { VIEWER_PATHNAME } from "../../routes";
-import FileUploadModal from "../FileUploadModal";
-import { DownArrow, Download } from "../Icons";
-import {
-    SetErrorAction,
-    SetViewerStatusAction,
-} from "../../state/viewer/types";
+import { DATA_BUCKET_URL, TOOLTIP_COLOR } from "../../constants";
+import { NetworkedSimFile, LocalSimFile } from "../../state/trajectory/types";
+import { Download } from "../Icons";
 
 import styles from "./style.css";
-import { LocalSimFile } from "../../state/trajectory/types";
 
 interface DownloadTrajectoryMenuProps {
     isBuffering: boolean;
-    simulariumFile: LocalSimFile | NetworkedSimFile; // NetworkedSimFile? i think this would only work with local file, but parent component will accept either
-    // selectFile: ActionCreator<RequestNetworkFileAction>;
-    // clearSimulariumFile: ActionCreator<ClearSimFileDataAction>;
-    // loadLocalFile: ActionCreator<RequestLocalFileAction>;
-    // setViewerStatus: ActionCreator<SetViewerStatusAction>;
-    // setError: ActionCreator<SetErrorAction>;
+    simulariumFile: LocalSimFile | NetworkedSimFile;
 }
 
 const DownloadTrajectoryMenu = ({
     isBuffering,
     simulariumFile,
-}: // clearSimulariumFile,
-// loadLocalFile,
-// selectFile,
-// setViewerStatus,
-// setError,
-DownloadTrajectoryMenuProps): JSX.Element => {
+}: DownloadTrajectoryMenuProps): JSX.Element => {
     const isLocalSimFile = (
         file: LocalSimFile | NetworkedSimFile
     ): file is LocalSimFile => {
@@ -51,14 +25,13 @@ DownloadTrajectoryMenuProps): JSX.Element => {
     const fetchFile = async (fileName: string): Promise<string> => {
         try {
             const response = await fetch(
-                //JOE QUESTION what URL to use
-                // `URLUNKOWN/download?file={${fileName}.simularium}`
-                `https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/BloodPlasma.simularium`
-                // `https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/{fileName}`
+                // `https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/BloodPlasma.simularium`
+                `${DATA_BUCKET_URL}/trajectory/${fileName}`
             );
             const data = await response.text();
             return data;
         } catch {
+            //QUESTION:should i be passing down and using the setError action?
             console.log("error fetching file");
             return "";
         }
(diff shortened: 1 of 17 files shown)


src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+        if (!fileIsLoaded()) {
+            return "";
+        }
+        if (isNetworkSimFileInterface(simulariumFile)) {
+            return `${DATA_BUCKET_URL}/trajectory/${simulariumFile.name}`;
+        } else {
+            const data: ISimulariumFile = simulariumFile.data;
+            // won't work if they loaded a binary file
+            // TODO: need to wait for simulariumFile.data.getAsBlob() to be created on viewer;
+            const blob = data.getAsBlob();
```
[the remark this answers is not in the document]
interim17: done

meganrm:
I tried to verify this but I couldn't get any of the remote trajectories to load

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+            const data: ISimulariumFile = simulariumFile.data;
+            // won't work if they loaded a binary file
+            const blob = data.getAsBlob();
+            return URL.createObjectURL(blob);
+        }
+    };
+
+    const downloadFile = async (fileName: string): Promise<void> => {
+        const downloadLink = document.createElement("a");
+        setIsDownloading(true);
```
meganrm, line 43: have you seen the button switch to "Downloading..."? Because none of these are written as async functions, it seems like the time between `setIsDownloading(true)` and `setIsDownloading(false)` is always going to be imperceivable

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        document.body.removeChild(downloadLink);
+        URL.revokeObjectURL(downloadLink.href);
+        setIsDownloading(false);
+    };
+
+    const onClick = () => {
+        if (!fileIsLoaded()) {
+            return;
+        }
+        downloadFile(simulariumFile.name);
```
meganrm, line 58:
```suggestion
        setIsDownloading(true);
        await downloadFile(simulariumFile.name);
        setIsDownloading(false);
```

and remove them from the other function

meganrm: actually scratch this, I don't think it's ever possible to see the change in the tooltip, and that seems like the only reason for the `isDownloading` flag? When I click 'download' I never see it change it "Downloading..."

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+    return (
+        <div className={styles.container}>
+            <Tooltip
+                placement="left"
+                title={
+                    isDownloading
+                        ? "Downloading..."
+                        : simulariumFile.name
+                        ? "Download trajectory"
+                        : "Load a trajectory to perform this action"
```
meganrm, line 70:
```suggestion
                    simulariumFile.name
                        ? "Download trajectory"
                        : "Load a trajectory to perform this action"
```

meganrm: if you were going to keep the `isDowloading` I would suggest you change this to a "getTooltip" function with if else statements for readability

src/components/Icons/index.tsx
```diff
@@ -21,12 +22,13 @@ export const Loading = <LoadingOutlined style={{ fontSize: 40 }} spin />;
 export const Play = <CaretRightOutlined />;
 export const Pause = <PauseOutlined />;
 export const UploadFile = <ImportOutlined />;
-export const DownArrow = <DownOutlined />;
+export const DownArrow = <DownOutlined size={32} />;
```
meganrm, line 25: why was this change needed?

src/components/Icons/index.tsx
```diff
@@ -21,12 +22,13 @@ export const Loading = <LoadingOutlined style={{ fontSize: 40 }} spin />;
...
 export const CaretRight = <CaretRightOutlined />;
 export const GoBack = <ArrowLeftOutlined />;
 export const Reset = <HomeOutlined />;
 export const ZoomIn = <PlusOutlined />;
 export const ZoomOut = <MinusOutlined />;
+export const Download = <DownloadOutlined size={32} />;
```
meganrm, line 31: this seems redundant with `DownArrow`?

interim17: They are different icons, the download one has an arrow pointing into a tray.

src/components/DownloadTrajectoryMenu/index.tsx
interim17: Ok, taking your suggestion.

src/components/DownloadTrajectoryMenu/index.tsx
interim17: @meganrm when you download a networked file do you get anything at all, or do you get a file that doesn't load properly in the viewer?

meganrm: I wasn't able to load a networked file. I just loaded a local file and then downloaded it again, which did work

src/components/DownloadTrajectoryMenu/index.tsx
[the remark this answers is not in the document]
meganrm: getAsBlob works for either binary or JSON, so you don't need the other comment either

src/components/DownloadTrajectoryMenu/index.tsx
meganrm: unfortunately this seems to be a bug with the viewer version 3.6.0. I revered it back (and removed the `getAsBlob`) and the networked files loaded. @toloudis or @ascibisz do you know why this is? Here are the changes in the new release: https://github.com/simularium/simularium-viewer/releases/tag/v3.6.0, and here is the error I'm getting 
``` 
[Debug] [netconnection] – "WS Connection Request Sent: " – "ws://staging-node1-agentviz-backend.cellexplore.net:9002/" (vendor.d1aa0b7eabb278739d2c.js, line 54046)

[Error] WebSocket connection to 'ws://staging-node1-agentviz-backend.cellexplore.net:9002/' failed: The network connection was lost.
```

toloudis: I suspect it's this one: https://github.com/simularium/simularium-viewer/pull/316 and maybe we need a patch release?

interim17: yeah that's it. this issue wasn't showing up on my machine because i manually fixed the ws/wss issue in node modules yesterday after working with Alli.

meganrm: yeah, I tested it with main and it works. I also got this other error
```
Uncaught TypeError: Cannot read properties of undefined (reading 'sendWebSocketRequest')
```
So I added [a PR to fix it](https://github.com/simularium/simularium-viewer/pull/317), but I'm not sure why the websocketClient is undefined

meganrm: This has now all been merged in and I published a new version. You should update the viewer to 6.0.1 and then it should be good to go

meganrm (approved):
Approve, but don't forget to remove the binary comment first

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        if (!fileIsLoaded()) {
+            return;
+        }
+        downloadFile(simulariumFile.name);
+    };
+
+    return (
+        <div className={styles.container}>
+            <Tooltip
+                placement="left"
```
frasercl, line 60: Tooltip covers the buttons to the left. Should it go below instead?

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const downloadFile = async (fileName: string): Promise<void> => {
+        const downloadLink = document.createElement("a");
+        downloadLink.download = fileName;
+        downloadLink.style.display = "none";
+        downloadLink.href = getHref();
+        document.body.appendChild(downloadLink);
+        downloadLink.click();
+        document.body.removeChild(downloadLink);
+        URL.revokeObjectURL(downloadLink.href);
+    };
```
frasercl, line 48: This function doesn't need to be `async` unless it contains an `await` (I'm pretty sure).

src/components/DownloadTrajectoryMenu/index.tsx
interim17: Reset it to "bottomLeft" which seems to obscure less

src/components/DownloadTrajectoryMenu/index.tsx
interim17: You are correct

frasercl (approved):

Issue #418, opened 2023-05-11
Preserve time in URL so user can share


lynwilhelm:
## Use Case

As a user, I want to share a trajectory at a specific time point with another user via a URL so that I can point out something about the state of the simulation at a specific time.

## Acceptance Criteria

- Time can be preserved in URL string
- UI design and interaction design implemented as specified in designs
- Time units used in string must be vetted with team

## Details

[Link to Designs](https://www.figma.com/file/9i8Dphcfp3t7e7TouwNheB/Simularium-share-time?type=design&node-id=96%3A1941&t=QsNFwDZNp1jK6BLw-1)

lynwilhelm:
@interim17 Yes on tooltip with aligned styling and I've updated the URL above to be the developer link instead of review mode link. Thanks so much!

interim17:
@lynwilhelm i'm going to make the input box where users specify their desired time stamp larger than shown in the design because i'm noticing that that are some large numbers as possible values (5+ digits) and they are cut off at the current width of 51 pixels

blairlyons:
duplicates https://github.com/simularium/simularium-planning/issues/113

Issue #410, opened 2023-04-07
Implement save time in URL


blairlyons:
## Use Case
As a user, I want to share a trajectory at a specific time point with another user via a URL so that I can point out something about the state of the simulation at a specific time.

## Acceptance Criteria
Specifications TBD in https://github.com/simularium/simularium-planning/issues/139

PR #424, opened 2023-06-09
Feature/share time url

interim17:
Problem
=======
Users would like to share trajectories at particular timestamps, so we are storing and accessing those timestamps with URL params. 

Closes:
[#418](https://github.com/simularium/simularium-website/issues/418) and [#410 ](https://github.com/simularium/simularium-website/issues/410)

with @meganrm 

Solution
========
This PR should address:
1) the ability to preserve a timepoint in a simulation via a URL param
2) the ability to share a URL with such a timestamp
3) and lay the groundwork/utilities for addressing future URL params (camera position, etc.)

## Type of change

Change summary:
---------------
Added UI components and functions for handling user input, Redux logics for setting trajectory state via incoming params, and utilities for checking, clearing and resetting URL params.

Steps to Verify:
----------------
1. Verify that the Share button in the top right is disabled when no trajectory is loaded.
2. Load a trajectory, select a time stamp, use  the share button.
3. Try toggling back and forth from default share time to custom.
4. Copy provided URL string into browser and verify that trajectory loads at specified time.


src/state/trajectory/logics.ts
```diff
@@ -332,10 +332,35 @@ const loadFileViaUrl = createLogic({
     type: LOAD_FILE_VIA_URL,
 });
```
interim17, line 334: This logic currently handles the time parameter, but has been written to anticipate handling future URL parameters.

src/state/trajectory/reducer.ts
```diff
@@ -73,6 +75,13 @@ const actionToConfigMap: TypeToDescriptionMap = {
             simulariumFile: initialState.simulariumFile,
         }),
     },
```
interim17, line 77: At this point the action does not replace anything in state and is just a vehicle to run the logic.

src/util/index.ts
```diff
@@ -73,6 +75,33 @@ export const wrapText = (
     };
 };
```
interim17, line 77: A util function to check if a URL has params.

src/util/index.ts
```diff
@@ -73,6 +75,33 @@ export const wrapText = (
...
+    const parsed = queryString.parse(location.search);
+    for (let index = 0; index < urlSettings.length; index++) {
+        const paramName = urlSettings[index];
+        if (parsed[paramName] !== undefined) {
+            return true;
+        }
+    }
+    return false;
+};
+
```
interim17, line 89: Bumped this from Modal component into a utility function in case future params also need to be reset, without it I was accidentally concatenating multiple time params when sharing a new timestamp from a URL that already had a param on it.

src/state/trajectory/reducer.ts
```diff
@@ -73,6 +75,13 @@ const actionToConfigMap: TypeToDescriptionMap = {
...
+    [SET_URL_PARAMS]: {
```
meganrm, line 78: you don't need a switch in the reducer if you're not saving anything in state

src/components/ShareTrajectoryModal/style.css
```diff
@@ -0,0 +1,106 @@
...
+.upload-modal :global(.ant-modal-title) {
+    font-size: 14px;
+}
+
+.upload-modal :global(.ant-modal-close-x) {
+    height: 40px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
```
meganrm, line 29: this seems like a lot of additional styling. I haven't looked at the designs in a while but are there really that many differences between our already created modals and this one? My thinking is that all our modals should be styled consistently so things like the checkbox and header should be in the customModals and not just for this modal

src/components/ShareTrajectoryModal/style.css
```diff
@@ -0,0 +1,106 @@
...
+
+.upload-modal :global(.ant-checkbox-inner) {
+    background-color: transparent;
+    border: 1px solid var(--dark-three) !important;
+}
+
+input {
+    border-radius: 10px;
```
meganrm, line 37: this is going to style every input with a border radius. If we want this, it should be in the general style sheet. Otherwise here it needs to be wrapped in a class

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+                  </Button>
+              ),
+          };
+
+    return (
+        <div className={styles.container}>
+            <CustomModal
+                className={styles.uploadModal}
+                title="Share Trajectory"
+                width={isLocalFile ? 611 : 550}
```
meganrm, line 163: why is this happening? where do these numbers come from?

src/components/ShareTrajectoryModal/style.css
```diff
@@ -0,0 +1,106 @@
...
+    text-align: right;
+    direction: rtl;
+    border: 1px solid var(--heather);
+    border-radius: 3px;
+}
+
+.number-input[disabled] {
+    width: 75px;
+    border: 1px solid var(--heather);
+    border-radius: 3px;
```
meganrm, line 71: this has a bunch of repeated values with `.number-input`, don't you only need the things that changed when it was disabled?

src/components/ShareTrajectoryModal/style.css
```diff
@@ -0,0 +1,106 @@
...
+    border-radius: 10px;
+}
+
+.blue-text {
+    color: var(--blue);
+    display: inline;
+}
+
+.body-text {
+    font-size: 14px;
```
meganrm, line 46: what's the default text size? I'm surprised how many font size settings there are in here. You should be able to use <h> tags to do some of the font adjustment so it's semantically styled

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+                            isLocalFile={isLocalFile}
+                            closeModal={handleShare}
+                        />
+                    </div>
+                ) : null}
+                <Button
+                    className={styles.shareButton}
+                    onClick={handleShare}
+                    type="primary"
+                    disabled={!simulariumFile.name || isBuffering || isSharing}
```
meganrm, line 58: why does this component care if the trajectory is buffering? I can't remember what this is for, but I think it was just to keep the user from pressing the play button while it was requesting more frames. But I don't think it should effect the ability to download since that's a totally different process?

src/components/ShareTrajectoryModal/index.tsx
interim17: they come from the UX mock up!

src/components/ShareTrajectoryButton/index.tsx
interim17: removed

src/components/ShareTrajectoryModal/style.css
interim17: ok, so i've added some global style rules for heading levels and removed local font-size styling, as well as changing html to be more semantic/accessible

src/components/ShareTrajectoryModal/style.css
interim17: consolidated

src/components/ShareTrajectoryModal/style.css
interim17: ok, i've moved a lot of these rules to CustomModal and light-theme

src/util/index.ts
meganrm: these new functions should have a few tests (there's a test file for this file in the same folder

src/containers/ViewerPanel/index.tsx
```diff
@@ -260,7 +264,7 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         });
     }
 
-    public receiveTimeChange(timeData: TimeData) {
+    public receiveTimeChange(timeData: TimeData): any {
```
meganrm, line 275: this return should probably be void? but not any

src/components/ShareTrajectoryModal/style.css
```diff
@@ -0,0 +1,37 @@
...
+.time-input[disabled] {
+    width: 75px;
+}
+
+.time-input[disabled] {
+    background: transparent;
+    color: var(--medium-grey);
+}
+
+.second-line {
```
meganrm, line 27: consider name like "timeInputContainer"

interim17: made this change

src/containers/ViewerPanel/index.tsx
interim17: done

meganrm (approved):

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const handleShare = () => {
+        setIsSharing(!isSharing);
+    };
+
+    const handleTooltipOpen = () => {
+        if (isSharing || !simulariumFile.name) {
+            return false;
+        } else return;
+    };
```
ShrimpCryptid, line 34: Should this return true in the else branch?

Could also probably compress this to a one-liner, i.e.
```
const handleTooltipOpen = () => {
   return !isSharing && simulariumFile.name
}
```

ShrimpCryptid: Also, is there any reason to have this as a separate function instead of inline on line 42? Since line 56 does inline too.

interim17: I like both your suggestions and wish they worked because they are cleaner, but the under-the-hood mechanics of this antd prop make it so that the one liner function, and passing the values directly like on 56, cause the tooltip to be open all the time, rather than just when hovering over the button. Would love to streamline, but need the behavior to still be responsive somehow. 
56 works I think because the disabling is not responsive, its meant to be always active.

ShrimpCryptid: Messed around with this, you can also just remove the `open` prop altogether and the hovering behavior will still work! It looks like the `DownloadTrajectoryMenu` button next to this one in doesn't include the `open` prop.

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+                </Button>
+            ),
+        },
+    };
+
+    return (
+        <CustomModal
+            className={classNames(styles.uploadModal, theme.lightTheme)}
+            title="Share Trajectory"
+            width={isLocalFile ? 611 : 550}
```
ShrimpCryptid, line 180: I think Megan left a comment on this too but it got marked as outdated, did UX weigh in on this? Feels odd to have magic numbers here

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+                            disabled
+                        />
+                        <Button type="text" onClick={copyToClipboard}>
+                            Copy {Link}
+                        </Button>
+                    </div>
+                    <div className={styles.timeInputContainer}>
+                        {" "}
+                        <Checkbox onChange={handleAllowUserInput}></Checkbox>
+                        <p>Start at</p>
```
ShrimpCryptid, line 153: image
The "Start at" text looks misaligned on my side?

src/components/ShareTrajectoryButton/index.tsx
interim17: When I remove the open prop I do get the desired behavior re: hover, but I'm also able to see the tooltip in an empty viewer, and when the sharing modal is open (also stays open when hovering over the transparent overlay of the modal), which I don't think is the desired behavior. I can check in with UX to confirm.

src/components/ShareTrajectoryModal/index.tsx
interim17: I can check in with @lynwilhelm but yes I pulled these widths from the UX designs. Are you thinking they should be more responsive? In general we have a lot of fixed pixel sizes throughout the CSS in simularium website that should maybe be reviewed down the line in favor or more responsive design.

ShrimpCryptid: Got it. If this is convention for the project then that's fine :)

ShrimpCryptid (approved):
Looks good! Left a few comments for you. FYI, there's also some code coverage warnings in `logics.ts` and `actions.ts`? I think most of those are for code you didn't write but not sure if those would be worth addressing.

src/components/ShareTrajectoryModal/index.tsx
interim17: Ack, yes, you're right. Throwing this CSS rule on makes it look ok, but I want a better way. I thought align-items: center was handling it.

.start-time-text {
    padding-top: 8px;
}

src/components/ShareTrajectoryButton/index.tsx
interim17: Well actually I can delete that second return statement and it works, as long as i leave the rest, and it's cleaner for sure, thanks!
```
    const handleTooltipOpen = () => {
        if (isSharing || !simulariumFile.name) return false;
    };
```

src/components/ShareTrajectoryModal/index.tsx
ShrimpCryptid: Tweaked this a bit too, it looks like you could also change the `p` element for "Start at" to `div` and it will line up correctly, the same way the total time label on the right does it. But I'm not sure what other consequences that has for things like accessibility...

PR #428, opened 2023-07-10
Fix/css cleanup

meganrm:
Problem
=======
I was confused why you needed the 
```
.container .cancel-button {
  margin-right: 0px;
```
and so I started digging into the css organization

Solution
========
I consolidated some of the classes and made sure things were scoped well. I'll make comments below about the changes in more detail

### DIFF
--- a/src/components/ConversionProcessingOverlay/index.tsx
+++ b/src/components/ConversionProcessingOverlay/index.tsx
@@ -29,40 +29,39 @@ const ConversionProcessingOverlay = ({
                     cancelProcessing={toggleProcessing}
                 />
             ) : null}
-            <h2 className={styles.titleText}> File conversion in progress </h2>
-            <p className={styles.centerText}>
+            <h2 className={styles.title}> File conversion in progress </h2>
+            <div className={styles.content}>
                 {" "}
                 {fileName} is being converted and will load when complete.{" "}
                 <br></br>
                 Processing time will vary depending on file size.
-            </p>
-            <a
-                className={styles.goBackButton}
-                onClick={() => setCancelModalOpen(true)}
-            >
-                {" "}
-                {LeftArrow} Stop and go back to form{" "}
-            </a>
-            <div className={styles.spin}>
-                <Spin size={"large"}> </Spin>
+                <a
+                    className={styles.goBackButton}
+                    onClick={() => setCancelModalOpen(true)}
+                >
+                    {" "}
+                    {LeftArrow} Stop and go back to form{" "}
+                </a>
+                <div className={styles.processingIndicator}>
+                    <div className={styles.spinContainer}>
+                        <Spin size={"large"}> </Spin>
+                    </div>
+                    <div>Processing...</div>
+                </div>
+                <h3>Open another instance of Simularium </h3>
+                <Button
+                    className={styles.button}
+                    href={`https://simularium.allencell.org/viewer`}
+                    target="_blank"
+                >
+                    Open in new tab {UpRightArrow}
+                </Button>
             </div>
-            <p className={styles.processingMargin}> Processing... </p>
-            <h3 className={styles.mediumText}>
-                {" "}
-                Open another instance of Simularium{" "}
-            </h3>
-            <Button
-                className={styles.tabButton}
-                href={`https://simularium.allencell.org/viewer`}
-                target="_blank"
-            >
-                Open in new tab {UpRightArrow}
-            </Button>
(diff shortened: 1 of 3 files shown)


src/components/ConversionProcessingOverlay/index.tsx
```diff
@@ -29,40 +29,39 @@ const ConversionProcessingOverlay = ({
                     cancelProcessing={toggleProcessing}
                 />
             ) : null}
-            <h2 className={styles.titleText}> File conversion in progress </h2>
-            <p className={styles.centerText}>
+            <h2 className={styles.title}> File conversion in progress </h2>
+            <div className={styles.content}>
```
meganrm, line 33: `title` is as clear as `titleText`, `content` is less confusing than `centerText` (which seems like you're trying to center the text)

meganrm: I grouped everything within this container because conceptually they go together

src/components/ConversionProcessingOverlay/index.tsx
```diff
@@ -29,40 +29,39 @@ const ConversionProcessingOverlay = ({
...
+                </a>
+                <div className={styles.processingIndicator}>
+                    <div className={styles.spinContainer}>
+                        <Spin size={"large"}> </Spin>
+                    </div>
+                    <div>Processing...</div>
+                </div>
+                <h3>Open another instance of Simularium </h3>
+                <Button
+                    className={styles.button}
```
meganrm, line 53: the buttons had the same style, so I removed the specific names

src/components/ConversionProcessingOverlay/style.css
```diff
@@ -10,53 +10,46 @@
     justify-content: center;
 }
 
-.container h2,p,h3 {
-  color: var(--dark);
+.container h2,
+.container h3 {
```
meganrm, line 14: if you don't have the parent group after the `,` it will style every `<p>` as with the dark color

src/components/ConversionProcessingOverlay/style.css
```diff
@@ -10,53 +10,46 @@
...
+    color: var(--dark);
 }
 
-.spin {
-  margin: 20px 0;
-  padding: 30px 50px 10px;
-  text-align: center;
-  border-radius: 4px;
+.title {
+    font-weight: 700;
```
meganrm, line 19: I put the css doc in order of the jsx

src/components/ConversionProcessingOverlay/style.css
```diff
@@ -10,53 +10,46 @@
...
 
-.container .center-text {
-  text-align: center;
-  margin-bottom: 0px;
+.go-back-button {
+    margin-top: 10px;
 }
 
-.go-back-button svg {
-  fill: var(--blue);
```
meganrm, line 34: I tested it and this wasn't necessary

src/components/ConversionProcessingOverlay/style.css
```diff
@@ -10,53 +10,46 @@
...
 }
-
-.container .tab-button {
-  margin-bottom: 20px;
-  border: 1.75px solid var(--dark-four);
-}
-
-.container .cancel-button {
-  margin-right: 0px;
-  margin-top: 20px;
```
meganrm, line 60: this was caused by styling in the conversion form

src/containers/ConversionForm/style.css
```diff
@@ -4,19 +4,28 @@
     background-color: rgba(231, 228, 242, 1);
```
meganrm, line 4: I ordered this doc to align with the jsx

src/containers/ConversionForm/style.css
```diff
@@ -29,36 +38,27 @@
...
+    background-color: var(--overlay-purple);
 }
 
-.title {
-    font-size: 30px;
-    padding-bottom: 0px;
-    margin-bottom: 0em;
-    font-weight: 600;
-    letter-spacing: .05rem;
+.form-content button {
```
meganrm, line 51: in general, I would be careful about ever styling a pure html element. Since this was `.container button` it was styling everything within the conversion form (which the processing overlay was a child of) so the margin right was getting applied to your buttons

src/components/ConversionProcessingOverlay/style.css
interim17: this is so strange, it's working on your branch but when i changed it before it turned the arrow black.

interim17 (approved):

Issue #294, opened 2022-01-05
Implement Color Picker


lynwilhelm:
## Use Case

Users would like to quickly and easily specify custom colors for agents

## Acceptance Criteria
1. Ability to override colors by:
- Selecting from the system default palette
- Selecting a custom color from palette (up to 18 uniquely specified, ordered by most recent)
- Selecting a custom color from color canvas and/or hue slider
- Manually specifying a hex value
2. Ability to click onto "current" color window in picker to cancel current changes
3. Ability to see specified color before committing the update
4. Increase the hit target (width of color bar) in Agent Type row
5. Ability to see recent colors populated in color picker if/when specified in the code before loading in (ordered by agent's appearance top to bottom)
6. Ability to view hex codes of colors in palettes upon hover with tooltip (including "current" and "new" color windows)
7. Provide dynamic dependency between hue, picker canvas, hex code and "new" color box as user interacts, reflecting current color specified

## Details
Specs/Design Link:
https://xd.adobe.com/view/df166a3a-05be-4c19-abdc-9d3d123c6391-8a45/

Design Brief:
https://docs.google.com/document/d/1LL5_9avSMUkO4He1BfNwEnGcipGZV7rs679Zp2QUl_Y/edit

PR #431, opened 2023-09-22
Feature/color picker

interim17:
Problem
=======
Closes [#294](https://github.com/simularium/simularium-website/issues/294)

Solution
========
Implemented the Viewer code for changing agent colors in the website.

In Redux, we update selectors to account for new SelectionStateInfo that includes colorChanges, and have added actions for setting colorChanges and storing recently selected colors.

In components: ColorPicker and ColorPickerPopover are new, rendered by ColorSwatch. I was unsure here, as I need these  color selections to make it in and out of state, perhaps a higher level component or container should be interacting with the store and drilling things down, but the component chain seems long to be drilling a lot down (ModelPanel -> CheckBoxTree -> ColorSwatch -> ColorPIckerPopover, ColorPicker).

with @meganrm

## Type of change

Change summary:
---------------

Steps to Verify:
----------------

A visual of the popover, with color picker, and swatches showing new size and expected behavior if sub-agents are assigned different colors:
colorpicker


src/components/Checkbox/style.css
interim17, line 1: Checkboxes were overlapping with color swatches, preventing clicks from opening the color picker.

PR #432, opened 2023-09-25
Feature/color picker

interim17:
Problem
=======
Closes [#294](https://github.com/simularium/simularium-website/issues/294) - Users want to choose colors for their agents.

Solution
========
Agents can be selected by clicking on the color swatches in the side panel and interacting with a color picker component. Will allow users to change colors for agents and subagents, and keep track of recent color selections.

**Redux:** Actions, types, constants, reducer functions and initial state are all updated to handle setting `ColorChange` and `RecentColors`. As with most changes to redux, this touches a bunch of files but is pretty straightforward. 

**Selectors:** Updated in selection state branch, and ViewerPanel, to account for new typing of `SelectionStateInfo` that includes `ColorChange`.

**UI Components:** ColorPicker is new. We are implementing the react-colorful library to handle some of the functionality and then styling their components inside an antd popover.  There is a fair amount of CSS included some `!important` tags and various nested antd classes. Will happly remove/clean up rules if I missed opportunities to simplify. 

Changes to CheckBox and CheckBox tree are for passing down props to ColorPicker and resizing the swatches and containers to allow click behavior. Additionally, we had to swap the order of the `star, checkbox, color` to `star, color swatch, checkbox`, which may seem trivial, but because they're shared checkboxes they're being rendered as groups. 

**Note:** confirmed with UX team that shifting hue selection bar from vertical to horizontal orientation was acceptable.

with @meganrm

## Type of change

Steps to Verify:
----------------

- npm run start
- select agents and sub agents, confirm that the colors change in the viewer and in the side panel
- if sub agents are different colors, the parent color swatch should be blank with a white border
- recently selected colors should display in the color picker when selecting different agents

Screenshot 2023-12-07 at 4 31 06 PM
Screenshot 2023-12-07 at 4 31 14 PM


src/components/CheckBoxTree/index.tsx
```diff
@@ -99,6 +99,18 @@ class CheckBoxTree extends React.Component<CheckBoxTreeProps> {
         }
     };
 
+    getAgentTags = (agentName: string): string[] => {
+        const { treeData } = this.props;
+        const tags: string[] = [""];
+        const agent = treeData.find((agent) => agent.title === agentName);
+        if (agent) {
+            agent?.children.forEach((child) => {
```
meganrm, line 107: this `?` doesn't seem necessary if you're also checking for `if (agent)`

src/components/Checkbox/style.css
```diff
@@ -0,0 +1,3 @@
+.container :global(.ant-checkbox-wrapper) {
+width: 12px !important;
```
meganrm, line 2: is this !important needed?

interim17: pretty sure, I'll double check but some of these ant components will override styling unless you use the !important tag

src/components/CheckBoxTree/index.tsx
interim17: you're right

src/state/selection/selectors/index.ts
```diff
@@ -1,12 +1,19 @@
 import { createSelector } from "reselect";
 import { reduce } from "lodash";
 import { UIDisplayData } from "@aics/simularium-viewer/type-declarations";
-import { SelectionEntry } from "@aics/simularium-viewer/type-declarations/simularium/SelectionInterface";
+import {
+    ColorChanges,
+    SelectionEntry,
+} from "@aics/simularium-viewer/type-declarations/simularium/SelectionInterface";
```
meganrm, line 7: you shouldn't be importing things from the file structure within the viewer. If you need these to be available to the front end, you should export them the same as `UIDisplayData` is above

interim17: I'm a little confused by the way the type-declarations are generated and exported.

The interfaces for ColorChanges and UIDisplayData are declared in the same file (SelectionInterface), but the build scripts result in UIDisplayData being exported from @aics/simularium-viewer/type-declarations, and ColorChanges being exported from @aics/simularium-viewer/type-declarations/simularium.

I think that addresses your concern from above, as nothing is being imported directly from SelectionInterface, but I'm confused why the exports don't end up in the same place and Im not 100% sure that addresses your concern.

meganrm: did you add the export type to the viewer the same way it's done for `UIDisplayData`?

meganrm: you need to add it to the index file: 
```export type {
    SelectionStateInfo,
    UIDisplayData,
    VisDataFrame,
} from "./simularium";

File: src/components/ColorSwatch/index.tsx [hunk 4]
```diff
@@ -1,14 +1,59 @@
...
-const ColorSwatch = ({ color }: ColorSwatchProps): JSX.Element => {
+const ColorSwatch = ({
+    childrenHaveDifferentColors,
+    color,
+    agentName,
+    tags,
+}: ColorSwatchProps): JSX.Element => {
+    const [isColorPickerVisible, setColorPickerVisible] = useState(false);
+    const [initialColor, setInitialColor] = useState(color);
+
```
interim17, line 21: This state variable seems redundant, but we need it to maintain visibility of the initialColor in the picker.

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+const mapStateToProps = (state: any) => ({
+    recentColors: selectionStateBranch.selectors.getRecentColors(state),
+});
+
+const dispatchToPropsMap = {
+    setColorChanges: selectionStateBranch.actions.setColorChanges,
+    setRecentColors: selectionStateBranch.actions.setRecentColors,
+};
+
+export default connect(mapStateToProps, dispatchToPropsMap)(ColorPicker);
```
meganrm, line 157: this shouldn't be connected to state if it's a component (ie in the component folder). I thought we had discussed a way to send down this information without needing this connection

src/components/ColorSwatch/index.tsx
meganrm: would `currentColor` be a more accurate name here?

src/components/ColorPicker/index.tsx
ShrimpCryptid, line 1: Encountered a strange bug with the color picker where it seems to only show colors from the recent tab? (or choose colors that are out of date.)
http://localhost:9001/viewer?trajFileName=smoldyn_min1.simularium
Repro:
1. Click the color picker for `MinD-ATP (surface)`.
2. Click further along the color gradient to choose a bright green. The molecule will turn a turquoise color.
3. Click again, further to the right, to select a blue. The molecule turns pink.
4. Continuing to click other colors causes the molecule to be colored by whatever is in the fourth slot in the display of recent colors.

ShrimpCryptid (changes requested):
Going to submit the comments and bug reports I have now, though I'm still going through the rest of the files!

src/components/Checkbox/style.css
```diff
@@ -0,0 +1,3 @@
...
+}
```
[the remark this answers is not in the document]
interim17: grabbed these with prettier, thanks

src/components/ColorSwatch/styles.css
```diff
@@ -1,5 +1,5 @@
 .container {
-    width: 5px;
+    width: 12px;
     height: 16px;
     display: inline-block;
 }
```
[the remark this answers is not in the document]
interim17: I like this, added!

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        {" "}
+                    </div>
+                    <p> NEW </p>
+                </div>
+                <div className={styles.oldColorContainer}>
+                    <HexColorInput
+                        className={styles.hexInput}
+                        color={color}
+                        onChange={setColor}
+                    />
```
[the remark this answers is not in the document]
interim17: That could be nice, I think I feel neutral about it. Maybe defer to UX... @lynwilhelm thoughts on this font selection?

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+}: ColorPickerProps) => {
+    const [color, setColor] = useState(initialColor);
+
+    const handleColorChange = (color: string) => {
+        const colorChanges: ColorChangesMap = {
+            agents: { [agentName]: tags },
+            color: color,
+        };
+        setColorChanges(colorChanges);
+        updateRecentColors(color);
```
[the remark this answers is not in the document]
interim17: I agree that it's silly to capture all those micro color changes, my instinct is to let users store colors from within a session, rather than when the popup is closed, but again I'm fairly neutral. Will definitely implement one or the other change.

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    styles.agentColorsSwatches,
+                ])}
+            >
+                {AGENT_COLORS.map((color) => (
+                    <Tooltip
+                        key={color}
+                        align={{ offset: [2, 5] }}
+                        color="#69738A"
+                        overlayClassName={styles.tooltip}
+                        title={color.slice(1)}
```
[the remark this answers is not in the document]
interim17: I'm not sure I understand the comment?

src/components/ColorPicker/style.css
```diff
@@ -0,0 +1,124 @@
...
+    border-radius: 0px;
+    height: 115px;
+    width: 196px;
+}
+
+.container :global(.react-colorful__hue) {
+    border-radius: 0px;
+    width: 196px;
+    height: 12px;
+}
```
ShrimpCryptid, line 124:
```suggestion
}

```
Nit: Whitespace at end of file (This happens in a few places in this change. I have VSCode set up to automatically format files when I save, which usually catches this? Not sure if you already have that set up but it changed my life when I found out about it)

src/components/ColorSwatch/index.tsx
```diff
@@ -1,14 +1,57 @@
...
+            />
+            {isColorPickerVisible ? (
+                <ColorPicker
+                    agentName={agentName}
+                    tags={tags}
+                    initialColor={initialColor}
+                    isOpen={isColorPickerVisible}
+                    closePopover={closePopover}
+                />
+            ) : null}
```
ShrimpCryptid, line 53: Is this check redundant with the `isOpen` flag in `ColorPicker?`

src/components/ColorSwatch/index.tsx
```diff
@@ -1,14 +1,57 @@
...
+    const openPopover = () => {
+        setInitialColor(color);
+        setColorPickerVisible(true);
+    };
+
+    // could move these rules into the css sheet, but to get color from prop
+    // and apply rules conditionlly means there will be some inline styling
+    // figured its more clean to have it all visible here rather than some here and some in stylesheet
+    const style = childrenHaveDifferentColors
+        ? { border: "1px solid #d3d3d3" }
```
ShrimpCryptid, line 35: Should this color be a constant somewhere?

src/components/ColorSwatch/styles.css
ShrimpCryptid, line 1: Nit: Maybe trivial enough to inline in `ColorSwatch`? Not an important change though

src/components/CheckBoxTree/index.tsx
```diff
@@ -99,6 +99,18 @@ class CheckBoxTree extends React.Component<CheckBoxTreeProps> {
...
     };
 
+    getAgentTags = (agentName: string): string[] => {
+        const { treeData } = this.props;
+        const tags: string[] = [""];
+        const agent = treeData.find((agent) => agent.title === agentName);
+        if (agent) {
+            agent.children.forEach((child) => {
+                tags.push(child.value as string);
+            });
```
ShrimpCryptid, line 109: Nit: Could also do a map operation here to save you a few lines!
```suggestion
            return agent.children.map((child) => child.value as string);
```

src/components/CheckBoxTree/index.tsx
```diff
@@ -220,6 +232,10 @@ class CheckBoxTree extends React.Component<CheckBoxTreeProps> {
                 </Row>
                 <TreeNode headerContent={this.renderCheckAllButton()} />
                 {treeData.map((nodeData) => {
+                    const childrenHaveDifferentColors =
+                        !nodeData.children.every(
+                            (el) => el.color === nodeData.children[0].color
+                        );
```
ShrimpCryptid, line 238: Bug: This check fails if children have differently-capitalized hex strings.

Repro:
http://localhost:9001/viewer?trajFileName=pc4covid19.simularium
1. Open the simularium file.
2. Open any object with children.
3. Paste `A384EE` into the first child's color picker. The parent should show the empty box.
4. Paste `a384EE` into the second child's color picker. The parent should still show the empty box.
5. Paste `A384EE` into the second child's color picker. Now, the parent will update with the correct color.

ShrimpCryptid (changes requested):

src/components/ColorPicker/index.tsx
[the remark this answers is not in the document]
ShrimpCryptid: Ah, on second look I misunderstood this. I thought `color` was an array for some reason and missed that it was a string for a hex code, and that you were removing the `#`.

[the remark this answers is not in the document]
interim17: Megan clarified, this is just stripping the # off the color string. Swapped slice for replace to make it obvious.

src/components/ColorPicker/index.tsx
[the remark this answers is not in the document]
interim17: Pushed a commit that implements debounce, thanks for the suggestion, less of a party when you drag the pointer around, but much more reasonable.

toloudis:
There are merge conflicts

interim17:
> There are merge conflicts

Sorry about that I was having git confusions, Megan bailed me out. Converted back to draft anyway to deal with bug and some needed refactoring.

src/components/CheckBoxTree/index.tsx
meganrm: We also need the "" in the array, so it ends up still needing a `concat` or a `push`

ShrimpCryptid:
Bug: Changing the parent color changes the children's color in the main viewport but does not change the children's color pickers. 
image

Repro:
- Open the `pc4covid19` simularium file.
- Expand the Macrophage dropdown.
- Change the parent color for Macrophage.

src/components/CheckBoxTree/style.css
```diff
@@ -1,91 +1,102 @@
...
-    font-size: 10px;
-    text-transform: uppercase;
+.sub-menu {
+    flex-flow: row;
 }
 
-.container :global(.ant-checkbox-inner) {
-    border-color: var(--warm-gray);
+.sub-menu :global(.ant-col-offset-3) {
+    margin-left: 13.4%;
```
ShrimpCryptid, line 61: Is this an arbitrary value?

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+    tags,
+    recentColors,
+    setColorChange,
+    setRecentColors,
+    selectedColor,
+    childrenHaveDifferentColors,
+}: ColorPickerProps) => {
+    const [color, setColor] = useState(selectedColor);
+    const [debouncedColor] = useDebounce(color, 250);
+    const isInitialRender = useRef(true);
```
ShrimpCryptid, line 38: Remove? It doesn't seem to be in use anymore

meganrm:
Update: after talking with Lyndsay we decided it made more sense to call the swatch that reverts the color change "previous" instead of "current"

src/components/ColorPicker/index.tsx
meganrm: it is used, but it's accessed as isInitialRender.current

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+                        style={{ background: color }}
+                        onClick={() => setColor(color)}
+                    />
+                ))}
+            </div>
+        </div>
+    );
+    const style = childrenHaveDifferentColors
+        ? { border: "1px solid #d3d3d3" }
+        : { backgroundColor: lastSelectedColor };
```
ShrimpCryptid: Should this be `color` and not `lastSelectedColor`?

src/components/ColorPicker/index.tsx
ShrimpCryptid: Oh I see, I skipped over how it was being used in the `if` check below. My bad!

src/components/ColorPicker/index.tsx
```diff
@@ -145,7 +145,7 @@ const ColorPicker = ({
     );
     const style = childrenHaveDifferentColors
         ? { border: "1px solid #d3d3d3" }
-        : { backgroundColor: color };
+        : { backgroundColor: lastSelectedColor };
```
ShrimpCryptid: I actually like the previous behavior where this was `color` instead of `lastSelectedColor`, because the agents in the viewport already update when you change the color. Maybe a question for Lyndsay, but I feel like it makes sense to me for this box to show the current color.

ShrimpCryptid (approved):
Looks good! Works when testing with a local build, thank you for fixing those bugs.

src/components/CheckBoxTree/style.css
meganrm: it is, but it's actually not new. I think this was a way for me to do consistent columns that were between ant's 3 and 4.

ShrimpCryptid:
I think this wasn't a change introduced by this PR, but the cancel button is very difficult to see against the background of the page when hovered or active...
image

meganrm:
> I think this wasn't a change introduced by this PR, but the cancel button is very difficult to see against the background of the page when hovered or active...
> image

was this for the other PR?

ShrimpCryptid:
> > I think this wasn't a change introduced by this PR, but the cancel button is very difficult to see against the background of the page when hovered or active...
> > image
> 
> was this for the other PR?

oops yes :X

src/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }, [initialColor]);
+
+    const updateRecentColors = (color: string) => {
+        if (recentColors.includes(color)) {
+            return;
+        }
+        const newRecentColors = [color, ...recentColors];
+        if (newRecentColors.length > 18) {
+            newRecentColors.pop();
+        }
```
frasercl, line 79: nit: even though something truly terrible would have to happen to make it run more than once, I'd still use `while` here for extra safety.

src/components/CheckBoxTree/index.tsx
```diff
@@ -195,41 +209,74 @@ class CheckBoxTree extends React.Component<CheckBoxTreeProps> {
...
+            <ColorPicker
+                selectedColor={value.color || nodeData.color}
+                agentName={nodeData.title}
+                tags={[value.value as string]}
+                recentColors={recentColors}
+                setColorChange={setColorChange}
+                setRecentColors={setRecentColors}
+            />
+        );
+    };
```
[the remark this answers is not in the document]
meganrm: yeah, that was the thinking I want through, and landed on the conditional checks being uglier than two functions. but it does bother me too

src/components/ColorPicker/style.css
```diff
@@ -0,0 +1,156 @@
...
+
+.swatch:hover {
+    transform: scale(1.29);
+    box-shadow: 0px 0px 10px #000000;
+}
+
+.color-picker .tooltip {
+    box-shadow: 0px 3px 4px #000000;
+    height: 20px !important;
+}
```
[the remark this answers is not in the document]
meganrm: yeah you're right. It wasn't needed at all

src/components/ColorPicker/style.css
```diff
@@ -0,0 +1,156 @@
...
+    color: var(--dark-theme-popover-text)
+}
+
+.tooltip :global(.ant-tooltip-inner) {
+    font-size: 9px;
+    letter-spacing: 0px;
+    min-height: 0px;
+    padding: 5px 10px 16px 10px;
+    text-transform: uppercase;
+    height: 20px !important;
```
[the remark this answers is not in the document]
meganrm: I was able to fix this by adjusting the padding, so the heights not needed at all

frasercl (approved):
Looks good!

src/components/CheckBoxTree/index.tsx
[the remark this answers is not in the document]
meganrm: I agree with this but I don't want to mess up the diff with a rename, so I'll do it after you look at the other changes

Issue #387, opened 2023-01-10
[Auto conversion]-Implement new error messaging content when user tries to load a non-convertible file


meganrm:
[Link to design](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/a75ea569-5e31-49f9-8a32-dc2c45823ba1/specs/)

PR #434, opened 2023-09-26
Feature/file type error

interim17:
Problem
=======
Closes [#387](https://github.com/simularium/simularium-website/issues/387)

Solution
========
This is an MVP implementation that checks file extensions and fires up the modal if there is not an allowed extension. Valid extensions for the various engines aren't defined in detail yet.

More checks can be added here or on the back end.

There is a new modal component for this error, and some functions for handling extension check and modal behavior in the ConversionForm container.

Steps to Verify:
----------------
Any uploaded file with an invalid extension should trigger the modal after clicking 'Next' on the import page.

fileerror

### DIFF
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -1,0 +1,51 @@
+import { Button } from "antd";
+import React from "react";
+
+import CustomModal from "../CustomModal";
+
+import styles from "./style.css";
+import { Exclamation } from "../Icons";
+
+interface ConversionFileErrorModalProps {
+    closeModal: () => void;
+}
+
+const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
+    closeModal,
+}) => {
+    const footerButtons = (
+        <>
+            <Button type="primary" onClick={closeModal}>
+                Ok
+            </Button>
+        </>
+    );
+
+    return (
+        <CustomModal
+            className={styles.errorModal}
+            title="File import cannot be completed"
+            open
+            footer={footerButtons}
+            width={425}
+            centered
+            onCancel={closeModal}
+        >
+            <div>
+                <div className={styles.redText}>
+                    {" "}
+                    {Exclamation}{" "}
+                    {`We're sorry, there was a problem importing your file.`}
+                </div>{" "}
+                You may want to double check that the file you selected is a
+                Smoldyn file and try again. For further assistance, please visit
+                <a href="https://forum.allencell.org/">
+                    {" "}
+                    The Allen Cell Discussion Forum.{" "}
+                </a>
+            </div>
+        </CustomModal>
+    );
+};
+
+export default ConversionFileErrorModal;
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -5,13 +5,16 @@ import CustomModal from "../CustomModal";
 import { Exclamation } from "../Icons";
 
 import styles from "./style.css";
+import { AvailableEngines } from "../../state/trajectory/conversion-data-types";
 
 interface ConversionFileErrorModalProps {
     closeModal: () => void;
+    engineType: AvailableEngines;
 }
 
 const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
     closeModal,
+    engineType,
 }) => {
     const footerButton = (
         <Button type="primary" onClick={closeModal}>
@@ -36,7 +39,8 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
                     {`We're sorry, there was a problem importing your file.`}
                 </div>{" "}
                 You may want to double check that the file you selected is a
-                Smoldyn file and try again. For further assistance, please visit
+                valid {engineType} file and try again. For further assistance,
+                please visit
                 <a
                     href="https://forum.allencell.org/"
                     target="_blank"
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -35,8 +35,7 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             <div>
                 <div className={styles.redText}>
                     {" "}
-                    {Exclamation}{" "}
-                    {`We're sorry, there was a problem importing your file.`}
+                    {"We're sorry, there was a problem importing your file."}
                 </div>{" "}
                 You may want to double check that the file you selected is a
                 valid {engineType} file and try again. For further assistance,
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -2,9 +2,9 @@ import { Button } from "antd";
 import React from "react";
 
 import CustomModal from "../CustomModal";
+import { Exclamation } from "../Icons";
 
 import styles from "./style.css";
-import { Exclamation } from "../Icons";
 
 interface ConversionFileErrorModalProps {
     closeModal: () => void;
@@ -13,12 +13,10 @@ interface ConversionFileErrorModalProps {
 const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
     closeModal,
 }) => {
-    const footerButtons = (
-        <>
-            <Button type="primary" onClick={closeModal}>
-                Ok
-            </Button>
-        </>
+    const footerButton = (
+        <Button type="primary" onClick={closeModal}>
+            OK
+        </Button>
     );
 
     return (
@@ -26,8 +24,8 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             className={styles.errorModal}
             title="File import cannot be completed"
             open
-            footer={footerButtons}
-            width={425}
+            footer={footerButton}
+            width={426}
             centered
             onCancel={closeModal}
         >
@@ -39,7 +37,11 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
                 </div>{" "}
                 You may want to double check that the file you selected is a
                 Smoldyn file and try again. For further assistance, please visit
-                <a href="https://forum.allencell.org/">
+                <a
+                    href="https://forum.allencell.org/"
+                    target="_blank"
+                    rel="noreferrer"
+                >
                     {" "}
                     The Allen Cell Discussion Forum.{" "}
                 </a>
--- a/src/containers/ConversionForm/index.tsx
+++ b/src/containers/ConversionForm/index.tsx
@@ -66,9 +66,6 @@ const ConversionForm = ({
     };
 
     const validateFileType = () => {
-        // currently only checking file extension, and allowing smoldyn
-        // valid files could have txt extention, this isnt a good check
-        // how extensive of a validation do we want to do on the front end?
         if (fileToConvert) {
             const fileExtension =
                 fileToConvert.name.split(".").pop()?.toLowerCase() || "";
--- a/src/containers/ConversionForm/index.tsx
+++ b/src/containers/ConversionForm/index.tsx
@@ -97,7 +97,10 @@ const ConversionForm = ({
     const conversionForm = (
         <div className={classNames(styles.container, theme.lightTheme)}>
             {fileTypeErrorModalOpen ? (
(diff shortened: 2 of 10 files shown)


src/containers/ConversionForm/index.tsx
```diff
@@ -55,15 +56,35 @@ const ConversionForm = ({
...
+    };
     const toggleProcessing = () => {
         setIsProcessing(!isProcessing);
     };
 
+    const validateFileType = () => {
+        if (fileToConvert) {
+            const fileExtension =
+                fileToConvert.name.split(".").pop()?.toLowerCase() || "";
+            const validExtensions = ["smoldyn"];
```
meganrm, line 72: I'm pretty sure smoldyn files are `.txt` not `.smoldyn`. @blairlyons?

meganrm: yeah some of the other engines have their own extensions. But we definitely want people to be able to convert smoldyn files, which are txt files, so that's what this needs to be. I don't think there are any files that are `.smoldyn`

src/components/ConversionFileErrorModal/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+            onCancel={closeModal}
+        >
+            <div>
+                <div className={styles.redText}>
+                    {" "}
+                    {Exclamation}{" "}
+                    {`We're sorry, there was a problem importing your file.`}
+                </div>{" "}
+                You may want to double check that the file you selected is a
+                Smoldyn file and try again. For further assistance, please visit
```
meganrm, line 39: the `smoldyn` here should be a variable based on what engine they chose

interim17: Ok, fixed this and it led to some other changes. 

Instead of an array of file extensions we should have a map, and I defined a SET_CONVERSION_ENGINE block in the reducer. Actually working with conversion engines is out of scope for this PR, but this sets us up to be dynamic when we get to that issue.

src/containers/ConversionForm/index.tsx
```diff
@@ -36,6 +37,12 @@ interface ConversionProps {
     setError: ActionCreator<SetErrorAction>;
 }
 
+interface ExtensionMap {
+    [key: string]: string;
+}
+
```
interim17, line 43: Can adjust this if there are multiple possible extensions for some engines...

src/containers/ConversionForm/index.tsx
```diff
@@ -55,15 +62,46 @@ const ConversionForm = ({
...
     const [engineSelected, setEngineSelected] = useState<boolean>(false);
     const [isProcessing, setIsProcessing] = useState<boolean>(false);
+    const [fileTypeErrorModalOpen, setFileTypeErrorModalOpen] = useState(false);
 
+    const toggleModal = () => {
+        setFileTypeErrorModalOpen(!fileTypeErrorModalOpen);
+    };
     const toggleProcessing = () => {
         setIsProcessing(!isProcessing);
     };
```
interim17, line 73: New handler sets engine in store, then validator checks file extension against store. I think this is more robust and we'll be storing that engine anyway.

src/containers/ConversionForm/index.tsx
```diff
@@ -55,15 +62,46 @@ const ConversionForm = ({
...
+                validFileExtensions[conversionProcessingData.engineType] ===
+                fileExtension
+            ) {
+                setIsProcessing(!isProcessing);
+                return;
+            }
+        }
+        setFileTypeErrorModalOpen(true);
+    };
+
```
interim17, line 94: New handler sets engine in store, and validateFileType checks extension against that. More robust this way and no new state needed.

src/state/trajectory/reducer.ts
```diff
@@ -97,6 +99,20 @@ const actionToConfigMap: TypeToDescriptionMap = {
...
+            state: TrajectoryStateBranch,
+            action: SetConversionEngineAction
+        ) => ({
+            ...state,
+            processingData: {
+                ...state.processingData,
+                engineType: action.payload,
+            },
+        }),
+    },
```
interim17, line 115: slightly out of scope for this PR, but we can flesh it out adjust later when we work with the engines?

src/containers/ConversionForm/index.tsx
interim17, line 94: New handler sets engine in store, and validateFileType checks extension against that. More robust this way and no new state needed.

src/containers/ConversionForm/index.tsx
```diff
@@ -36,6 +37,12 @@ interface ConversionProps {
     setError: ActionCreator<SetErrorAction>;
 }
 
+interface ExtensionMap {
+    [key: string]: string;
```
meganrm, line 41: this should be typed as `[key: AvailableEngines]`

meganrm: also this interface should be in `state/trajectory/conversion-engine-types`

src/containers/ConversionForm/index.tsx
```diff
@@ -55,15 +62,46 @@ const ConversionForm = ({
...
+    const handleEngineChange = (selectedValue: string) => {
+        const selectedEngine = selectedValue as AvailableEngines;
+        setConversionEngine(selectedEngine);
+        setEngineSelected(true);
+    };
+
+    const validateFileType = () => {
+        if (fileToConvert) {
+            const fileExtension =
+                fileToConvert.name.split(".").pop()?.toLowerCase() || "";
```
meganrm, line 83: why is the pop uncertain here? split always returns an array

interim17: ok will remove

src/containers/ConversionForm/index.tsx
interim17: Ok , moving to proper folder.

Typescript doesn't like `[key: AvailableEngines]`, doesn't an object key have to be a string?

Tried this `[key in AvailableEngines]: string` , but also getting errors.

Will keep poking around.

src/containers/ConversionForm/index.tsx
interim17: it's giving `Object is possibly undefined ` without the `?`

Can either leave the `?` or refactor ir

meganrm: is that because `name` might be undefined ? I would probably send in `fileToConvert.name` as an argument in this function ie. 
```const validateFileType = (fileName) => {
              if (fileName) {
                    const fileExtension = fileName.split(".").pop()?.toLowerCase() || "";
```

src/containers/ConversionForm/index.tsx
meganrm: example: 
```TypeScript
export const ENGINE_TO_TEMPLATE_MAP: { [key in AvailableEngines]: string } = {
    [AvailableEngines.Smoldyn]: "smoldyn_data",
    [AvailableEngines.Cytosim]: "cytosim_data",
    [AvailableEngines.CellPack]: "cellpack_data",
    [AvailableEngines.SpringSalad]: "springsalad_data",
};
```

meganrm: what's the error you're getting?

meganrm:
```TypeScript 
export type ExtensionMap = {
    [key in AvailableEngines]: string;
}
```

src/containers/ConversionForm/index.tsx
```diff
@@ -55,15 +59,46 @@ const ConversionForm = ({
...
+            }
+        }
+        setFileTypeErrorModalOpen(true);
+    };
+
     // TODO: use conversion template data to render the form
     console.log("conversion form data", conversionProcessingData);
     const conversionForm = (
         <div className={classNames(styles.container, theme.lightTheme)}>
+            {fileTypeErrorModalOpen ? (
```
meganrm, line 96:
```suggestion
            {fileTypeErrorModalOpen && (
```
and remove the `:null`. It's shorthand for what you have here. You could also do that for line 102

src/containers/ConversionForm/index.tsx
toloudis: pop returns undefined if the array is empty

interim17:
```
    const validateFileType = (fileName: string) => {
        if (fileName) {
            const fileExtension = fileName.split(".").pop().toLowerCase();
```

gives the same error which is too bad, i would figure the `if (fileName)` where serve as a check.

src/containers/ConversionForm/index.tsx
interim17: Ok

src/state/trajectory/conversion-data-types.ts
```diff
@@ -36,5 +36,8 @@ export type TemplateMap = {
...
 
 export interface Template {
-    [key: string]: CustomType
-}
\ No newline at end of file
+    [key: string]: CustomType;
+}
+export interface ExtensionMap {
+    [key: string]: string;
+}
```
meganrm, line 43: once you change this to 
```TypeScript
export type ExtensionMap = {
    [key in AvailableEngines]: string;
}
```
this looks good to me

src/containers/ConversionForm/index.tsx
```diff
@@ -55,21 +59,54 @@ const ConversionForm = ({
...
+        setConversionEngine(selectedEngine);
+        setEngineSelected(true);
+    };
+
+    const validateFileType = () => {
+        if (fileToConvert) {
+            const fileExtension = fileToConvert.name
+                .split(".")
+                .pop()
+                ?.toLowerCase();
```
meganrm, line 82: I would prefer the reworking I suggested (it's a little lost because the code has changed) to avoid the ambiguity of what is going to be undefined here

src/state/trajectory/conversion-data-types.ts
interim17: Done! Had to add "txt" to validFileExtensions for all four engines, I don't actually know what file extensions they each use.

src/containers/ConversionForm/index.tsx
```diff
@@ -36,6 +38,13 @@ interface ConversionProps {
     setError: ActionCreator<SetErrorAction>;
 }
 
+const validFileExtensions: ExtensionMap = {
+    Smoldyn: "txt",
+    cytosim: "txt",
+    cellPACK: "txt",
+    SpringSaLaD: "txt",
```
meganrm, line 45: put a TODO here to update these with their correct format once we support them, so that we know this isn't necessarily the correct data

meganrm (approved):
Looks good, I suggested you put a comment where we know we have future work to do.

src/components/ConversionFileErrorModal/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+            footer={footerButton}
+            width={426}
+            centered
+            onCancel={closeModal}
+        >
+            <div>
+                <div className={styles.redText}>
+                    {" "}
+                    {Exclamation}{" "}
+                    {`We're sorry, there was a problem importing your file.`}
```
ShrimpCryptid, line 39: Nit: Is there a reason for these being backticks instead of quotations?

src/containers/ConversionForm/index.tsx
```diff
@@ -123,8 +164,12 @@ const ConversionForm = ({
...
                 <Button
                     type="primary"
-                    disabled={!fileToConvert || !engineSelected}
-                    onClick={() => setIsProcessing(!isProcessing)}
+                    disabled={!readyToConvert}
+                    onClick={() =>
+                        readyToConvert
+                            ? validateFileType(fileToConvert.name)
+                            : null
+                    } // this will be unclickable anyway, but typescript doesn't' know that
```
ShrimpCryptid, line 172:
```suggestion
                        readyToConvert && validateFileType(fileToConvert.name)
                    } // this will be unclickable anyway, but typescript doesn't know that
```

Nit: Could use the same `&&` condition execution instead of conditional ternary (`?`)

src/containers/ConversionForm/index.tsx
```diff
@@ -36,6 +38,14 @@ interface ConversionProps {
...
 
+const validFileExtensions: ExtensionMap = {
+    // TODO: update this with correct extensions
+    Smoldyn: "txt",
+    cytosim: "txt",
+    cellPACK: "txt",
+    SpringSaLaD: "txt",
+};
+
 const selectOptions = Object.keys(AvailableEngines).map(
```
ShrimpCryptid, line 49: Could be nice to add an extension hint (.txt) after the engine names in the future?

ShrimpCryptid (approved):
Left some small notes/suggestions but otherwise looks good!

PR #448, opened 2023-11-20
Fix/share download styling

interim17:
Problem
=======
@lynwilhelm pointed out some styling mistakes on the`Share` and`Download` buttons, and added a couple tweaks not in the original designs.
Designs [here](https://www.figma.com/file/9i8Dphcfp3t7e7TouwNheB/Simularium-share-time?type=design&node-id=96-1941&mode=design&t=Of50s0v0xSpiAGTZ-0) and [here](https://xd.adobe.com/view/75b6a3c8-f650-443b-bccc-d1fcd134a35e-20e5/screen/d97dd2b2-08b4-4fe1-ae09-1dddb764240b)

This PR should fix those problems and I added a couple readability changes. I went over the visual changes with @lynwilhelm in a huddle so feel confident this branch reflects design intentions.

Styling changes:
- remove pipe divider
- swap button order
- icons slightly larger
- active button hover state should be lighter color
- tooltips should be closer to buttons and more centered
- new tooltip text

Note about selectors: `.container :global(.ant-tooltip-open.ant-btn)`
The hover state and tooltip open behavior were not in sync, so this rule is used of the `:hover` pseudo-class.

@lynwilhelm  also noticed a bug that I could not reproduce where a tooltip would open automatically when a trajectory is loaded, if the user had hovered over the share button BEFORE loading the trajectory. I'm crossing my fingers that some of these changes will resolve that issue, but please let me know if you can reproduce it on this branch.

Steps to Verify:
----------------
1. Run website
2. Check button styling before and after loading trajectory
3. From main page: hover over share button, then load trajectory
4. See if tooltip opens erroneously

Screenshots:
----------------

Before, disabled:
incorrectdisabled

Before, hover:
incorrecthover

incorrectdownload

After, disabled:
correctdisabled

After, hover:
correcthover

correcthovershare


meganrm:
can you add before and after screen shots?

interim17:
@meganrm  Added some screenshots of before and after for disabled and hover states. Let me know if you'd like to see anything else.

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -53,22 +53,33 @@ const DownloadTrajectoryMenu = ({
         downloadFile(simulariumFile.name);
     };
 
+    const isDisabled = () => {
```
ascibisz, line 56: nit: couldn't this just be a regular boolean const, instead of a function? like:

`const isDisabled = !fileIsLoaded() || isBuffering;`

Just feels a little more concise, since it's not taking in any arguments and is just returning a bool

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -53,22 +53,33 @@ const DownloadTrajectoryMenu = ({
...
+        return !fileIsLoaded() || isBuffering;
+    };
+
+    const getTooltipOffset = () => {
```
ascibisz, line 60: similarly to above, this could just be

`const tooltipOffset = isDisabled ? [25, -30] : [40, -18];`

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -24,31 +26,41 @@ const ShareTrajectoryButton = ({
...
-    const handleTooltipOpen = () => {
-        if (isSharing || !simulariumFile.name) return false;
+    const isDisabled = () => {
+        return !simulariumFile.name || isSharing || isBuffering;
+    };
+
+    const getTooltipOffset = () => {
+        if (isDisabled()) {
+            return [0, -30];
+        } else return [0, -18];
```
ascibisz, line 36: again, maybe it's just personal preference, but I feel like just having boolean constants here would nice

```
const isDisabled = !simulariumFile.name || isSharing || isBuffering;
const tooltipOffset = isDisabled ? [0, -30] : [0, -18];
```

ascibisz (approved):
My ability to judge CSS files is extremely limited, but this looks good for me! I tested it locally too and it looks good for me

src/components/ShareTrajectoryButton/style.css
```diff
@@ -2,11 +2,20 @@
...
+.container :global(.ant-tooltip-open.ant-btn) {
+    color: var(--baby-purple-two);
+}
+
+.container .disabled :global(.ant-btn){
+    color: var(--dim-gray-two);
+}
+
+.container :global(.anticon) {
+    font-size: 16px;
```
meganrm, line 18: since these are shared across the two components I would put these higher. You may be able to change the disabled color in `ant-vars.less` (you have to restart the server to see the changes). but make sure the other disabled buttons in the light modal are still the way they should be. 

For the other two colors: are all our tooltips this color? We have a number of colors in the `colors.css` file called `tooltip` something, are those not this color? If these are different colors than the other tooltips I would add a classname that indicates that and move this styling up so it applies to both buttons, like in the AppHeader. Same with the anticon thing: is this a more general style across the header? again could use it's own classname if not

meganrm (approved):
thanks for the changes to the css organization, looks good!

PR #452, opened 2023-12-07
Feature/conversion check server

### LINKED ISSUE #389: [Auto-conversion]-Implement server health check and error message modal
After the user selects a file to be auto converted we ping the server and make sure it's healthy before continuing. if the ping fails, show a message to the user. 

[link to design](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/0449cf8a-549f-4b1e-a4bf-45397ce51846/specs/)

interim17:
Problem
=======
We want to ping the Octopus server to see if it is healthy before letting the user proceed too far along the auto-conversion workflow.

Closes #389 

Solution
========

This is the front end follow up to [Alli's work](https://github.com/simularium/simularium-viewer/pull/349), which set up the health check communication between the the Viewer and Octopus.

Reviewing this PR requires a dev build of the viewer, and a local instance of Octopus running, fun! A glimpse into the future.

Style things
===
I did a walkthrough of the modal styling with @lynwilhelm and have her approval on it. If you look at [these designs ](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/a75ea569-5e31-49f9-8a32-dc2c45823ba1/specs/)you may notice that the auto-conversion designs are old and out of sync with the most current global style guidance. 

For devs in review, if something on your machine looks/behaves dramatically wrong, or is vastly different than the screen shot, please let me know, but don't be too concerned about the design brief, Lyndsay and I are in touch about it all. 

Screenshot 2023-12-07 at 1 28 26 PM
Hover state: 
Untitled 2

Code things
==

To check the server health we need to implement some logic to account for the fact that WebSocket requests don't have an automatic response if the request fails: 
- first we now store `conversionStatus` in state: we were already storing `status` for the viewer in the Viewer Branch, I'm cloning that idea in the trajectory branch to track conversion status. I don't think `VIEWER_IMPORTING` was sufficient for this case, and I anticipate it will be useful to have a little granularity here going forward.

`configureFileConversionLogic` is where the action happens:
- check for a properly configured controller and set one up if needed
- then we define `performHealthChecks` a recursive function that will send a defined number of health check requests to Octopus, the health checks include a handler for when a positive response comes back, and a net connection config.
- a timeout waits a certain number of seconds before assuming the last check failed to get a response (would appreciate feedback on how long to wait here, thoughts in code comments)
- each handler goes out with a `requestId` and each `requestId` is stored along with a three second `timeoutId` in `healthCheckTimeouts`
- if a handler fires (meaning the server responded positively) we clear that set of ids from `healthCheckTimeouts` set `healthCheckSuccesful` to true and update `conversionStatus` to `CONVERSION_SERVER_LIVE`
- if a request times out with no response and `healthCheckSuccesful` is still false we try again until we reach `MAX_ATTEMPTS` and then set `conversionStatus` to `CONVERSION_NO_SERVER`
- this logic is triggered by the `receiveFileToConvert` action, which can be called with a `null` payload if we just want to do a health check

- ConversionForm sends a health check on load, and then another when a file is uploaded. If the `conversionStatus` is negative when we click the "Next" button, the modal should be raised.

Steps to Verify:
----------------
1) You need a dev build of the viewer installed in the website. Building from `main` will work. `npm run build` and copy the path of your viewer repo into package.json
2) You need an Octopus instance running locally at server ip "0.0.0.0" and port 8765, I can try to help you set this up if needed, @ascibisz is a also a wiz at this.
3) When your server is running, `npm run start` the website and go to `http://localhost:9001/import` or choose "import other file type" from the Load model menu.
4) Select Smoldyn for your conversion engine and upload any file that ends in .txt
5) Click next. If your server is running you should proceed to the next page, and also see a log of "conversion-server-live" in your console.
6) To test failure, repeat steps 3-6 without running your local Octopus instance. When you click Next you should see the error modal.
7) Similarly, if you turn octopus on between page load and file upload step, you should be able to advance.
8) Vice versa, if octopus is not running when you load the page, but then you turn it on, and then you upload a file, it should hopefully let you advance!


src/state/trajectory/logics.ts
```diff
@@ -345,6 +353,102 @@ const loadFileViaUrl = createLogic({
...
+            }, 3000);
+
+            // store the time out id
+            healthCheckTimeouts[requestId] = timeoutId;
+        };
+
+        // Start the first health check
+        performHealthCheck(attempts);
+    },
+    type: CONFIGURE_FILE_CONVERSION,
```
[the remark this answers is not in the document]
interim17: or do we want to just run `CONVERT_FILE` with no payload whenever we health check? that's less actions/logics but also less semantic to me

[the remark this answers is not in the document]
interim17: when we go to the `/import` route to show the form we just trigger these:
```
                dispatch(setIsPlaying(false));
                dispatch(setStatus({ status: VIEWER_IMPORTING }));
 ```
 
 the `convertFile()` action isn't called anywhere yet, but I assumed that's what we would use to send the uploaded file to octopus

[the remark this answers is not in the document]
Comment (interim17) [reply src/state/trajectory/logics.ts]: Ok yes, I used `RECEIVE_FILE_TO_CONVERT` for readability since it doesn't have it's own redux logic.

Added some safeguards in reducer so that we can call with `null` payload on page load, but if I'm just being stubborn and you really only want this logic done as a side effect of other actions I'll remove it.

File: src/components/ConversionFileErrorModal/style.css [hunk 2]
```diff
@@ -7,7 +7,7 @@
     padding: 15px 15px 0px 15px;
 }
 
-.error-modal .red-text {
+.error-modal .red-warning-text {
     color: var(--brick-red);
```
meganrm, line 11:
```suggestion
    color: var(--warning-text-color);
```

src/components/ConversionServerErrorModal/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+            open
+            footer={
+                <Button type="primary" onClick={closeModal}>
+                    OK
+                </Button>
+            }
+            centered
+            onCancel={closeModal}
+        >
+            <p className={styles.redText}>
```
meganrm, line 29: nit: but I think `warningText` is a better class name

src/state/trajectory/logics.ts
```diff
@@ -345,6 +353,102 @@ const loadFileViaUrl = createLogic({
...
+            secureConnection: false,
+        };
+        // check if a controller exists and has the right configuration
+        let controller = getSimulariumController(getState());
+        if (!controller || !controller.remoteWebsocketClient) {
+            dispatch(
+                setConversionStatus({
+                    status: CONVERSION_NO_SERVER,
+                })
+            );
```
meganrm, line 381: why do you do this here? aren't you going to be checking the server below anyway?

src/state/trajectory/logics.ts
```diff
@@ -345,6 +353,102 @@ const loadFileViaUrl = createLogic({
...
+        // until we switch to Octopus, make sure it matches your local instance.
+        const netConnectionConfig: NetConnectionParams = {
+            serverIp: "0.0.0.0",
+            serverPort: 8765,
+            useOctopus: true,
+            secureConnection: false,
+        };
+        // check if a controller exists and has the right configuration
+        let controller = getSimulariumController(getState());
+        if (!controller || !controller.remoteWebsocketClient) {
```
meganrm, line 376: do you need to create a new controller to set the network settings? If the controller exists can't you just update it?

src/state/trajectory/reducer.ts
```diff
@@ -119,13 +134,22 @@ const actionToConfigMap: TypeToDescriptionMap = {
...
-            ...state,
-            processingData: {
-                ...state.processingData,
-                fileToConvert: action.payload,
-            },
-        }),
+        ) => {
+            // prevent side effects if called with no payload
+            // (e.g. when configuring server)
+            if (action.payload === null || action.payload === undefined) {
```
meganrm, line 140: I don't understand how this would happen

src/components/ConversionFileErrorModal/index.tsx
```diff
@@ -33,7 +32,7 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             onCancel={closeModal}
         >
             <div>
-                <div className={styles.redText}>
+                <div className={styles.redWarningText}>
                     {" "}
```
meganrm, line 36: I know this isn't part of the code changes, but why are these spaces here?

src/state/trajectory/logics.ts
interim17: yes, but there could be a case where the server was previously healthy, and now is not, and the attempt to set the controller below could fail, so it makes sense in the meanwhile to change status to `CONVERSION_NO_SERVER` just in case. that was my thinking at least

src/state/trajectory/logics.ts
interim17: yes, i think we could just update it if it exists. i can add that conditional.

interim17:
```
   if (!controller) {
            dispatch(
                setConversionStatus({
                    status: CONVERSION_NO_SERVER,
                })
            );
            controller = new SimulariumController({
                netConnectionSettings: netConnectionConfig,
            });
            dispatch(setSimulariumController(controller));
        } 
        else if (!controller.remoteWebsocketClient) {
            controller.configureNetwork(netConnectionConfig);
        }

File: src/components/ConversionFileErrorModal/index.tsx (hunk 7 above)
Comment (interim17) [reply src/components/ConversionFileErrorModal/index.tsx:36]: i think those are artifacts of using prettier to format the document

File: src/state/trajectory/reducer.ts (hunk 6 above)
Comment (interim17) [reply src/state/trajectory/reducer.ts:140]: So I went with an approach we we trigger this configure/servercheck logic on page load just to make sure that its all squared away before the user proceeds too far in the conversion process, and so that the modal is raised instantly if they reach that file upload break point. That's why I gave it its own action in the first place.

I took your suggestion of piggybacking that logic on the the receive file action instead, but that means that we could trigger that action without a payload and potentially wipe out some data that was in state by having the payload be null or undefined. It's unlikely given the order of operations but I'm just trying to account for that edge case.

I could just remove the useEffect and not run the server check on page load, or just remove this check and assume this never happens.

File: src/containers/ConversionForm/index.tsx [hunk 8]
```diff
@@ -1,31 +1,33 @@
...
-import { UploadFile } from "antd/lib/upload";
+import ConversionServerErrorModal from "../../components/ConversionServerErrorModal";
 import ConversionProcessingOverlay from "../../components/ConversionProcessingOverlay";
 import ConversionFileErrorModal from "../../components/ConversionFileErrorModal";
+import { CONVERSION_NO_SERVER } from "../../state/trajectory/constants";
+import customRequest from "./custom-request";
 
+import theme from "../../components/theme/light-theme.css";
+import styles from "./style.css";
 interface ConversionProps {
```
ShrimpCryptid, line 31:
```suggestion
import styles from "./style.css";

interface ConversionProps {
```
Nit: spacing

src/containers/ConversionForm/index.tsx
```diff
@@ -61,15 +64,35 @@ const ConversionForm = ({
...
+    const [serverDownModalOpen, setServerIsDownModalOpen] =
+        useState<boolean>(false);
     const [isProcessing, setIsProcessing] = useState<boolean>(false);
     const [fileTypeErrorModalOpen, setFileTypeErrorModalOpen] = useState(false);
 
-    const toggleModal = () => {
+    // On load, configure controller and check if server is healthy
+    useEffect(() => {
+        // calling this with null payload configures the controller and checks server health
+        receiveFileToConvert(null);
```
ShrimpCryptid, line 79: I kind of like the idea of having a separate `configureController` prop that gets passed in? It could still be the same base method getting passed in (`configureController={() => receiveFileToConvert(null)}` in the parent) but I feel like this behavior should either be explicitly documented in the arguments or it should just be a separate prop.

src/state/trajectory/logics.ts
meganrm: But isn't the point of this status to show the error modal? If you want it to default to that, you could just have it be the default in the reducer. But here it seems like an unnecessary setting of state

src/state/trajectory/reducer.ts
meganrm: yeah this isn't needed. We only need to call it once. If you want it to happen at the very beginning of the process you could use the `convert file` action instead, which could be renamed, it's really `initializeFileConversion`. Probably makes the most sense

src/components/ConversionServerErrorModal/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+            footer={
+                <Button type="primary" onClick={closeModal}>
+                    OK
+                </Button>
+            }
+            centered
+            onCancel={closeModal}
+        >
+            <p className={styles.warningText}>
+                {" "}
```
meganrm, line 30: Can you delete these and try to figure out how to keep them from happening? Are you adding spaces while you're typing that then get converted?

src/components/ConversionServerErrorModal/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+                {Exclamation} We&apos;re sorry, the server is currently
+                experiencing an issue.
+            </p>
+            <p>
+                {" "}
+                Please try again at a later time. For further assistance, please
+                visit{" "}
```
meganrm, line 37: This one is the only one that I think you need?

src/containers/ConversionForm/index.tsx
meganrm: This isn't a good way to use logics/actions. If you find yourself doing this, you should either change the action that trigger the logic, or write a new action just to trigger the logic. This is breaking the point of having all these very specifically named actions that just do one thing.

src/state/trajectory/logics.ts
interim17: It is the default state in the reducer. The case I'm trying o to anticipate is: we have had done a successful check, and set state to `CONVERSION_SERVER_LIVE` but then something happened to the network or the controller in the meanwhile. I'm trying to use `conversionState` as a way to track the whole process, get in front of bugs and and make the user facing interactions like modals easy to time and control.

But I think it's making review for this PR more complicated than it needs to be. I'll just remove the useEffect check on pageload, and this status setting.

If there are bugs in these kind of cases when octopus is actually live we can figure it out then. It will probably work fine this way.

meganrm: yeah, that's overkill. The overall goal is to just check it once, and if it's up assume we're good to go.

src/state/trajectory/logics.ts
```diff
@@ -345,6 +353,104 @@ const loadFileViaUrl = createLogic({
...
+        const { getState } = deps;
+
+        // TODO: Most likely this will eventually replace the config up top
+        // but for development purposes it's here
+        // until we switch to Octopus, make sure it matches your local instance.
+        const netConnectionConfig: NetConnectionParams = {
+            serverIp: "0.0.0.0",
+            serverPort: 8765,
+            useOctopus: true,
+            secureConnection: false,
```
meganrm, line 372: Do these need to be set back afterwards to enable someone to load a networked trajectory?

src/state/trajectory/logics.ts
```diff
@@ -345,6 +353,104 @@ const loadFileViaUrl = createLogic({
...
+        }
+        // check the server health
+        // Originally thought to send checks every 15 seconds,
+        // currently sending a flurry of checks, 3 seconds apart, 15 seconds total.
+        // If any come back true we assume we're good for now... this timing is arbitrary
+        // I reasoned a quick set of requests that can be started on page load,
+        // that cumulatively have enough delay to let a connection be configured
+        // is a better idea than a minute long period of sending checks,
+        // during which things could change.
+        // We can retrigger this as often as necessary.
```
ShrimpCryptid, line 398: This is a good explanation, thank you! Maybe worth shortening this comment in-code though, since this seems like a comment for GitHub/PR

src/containers/ConversionForm/index.tsx
interim17: so after talking with Megan we're actually going to pull this back out into a separate action `CREATE_FILE` is going to become `INITIALIZE_FILE_CONVERSION` and we're going to do it there.

src/state/trajectory/types.ts
```diff
@@ -92,3 +106,18 @@ export interface TimeUnits {
...
+    [requestId: string]: NodeJS.Timeout;
+}
+
+type CONVERSION_NO_SERVER = typeof CONVERSION_NO_SERVER;
+type CONVERSION_SERVER_LIVE = typeof CONVERSION_SERVER_LIVE;
+type CONVERSION_ACTIVE = typeof CONVERSION_ACTIVE;
+type CONVERSION_COMPLETE = typeof CONVERSION_COMPLETE;
+
+export type ConversionStatus =
+    | CONVERSION_NO_SERVER
```
ShrimpCryptid, line 120: Does this first line need a `|`?

src/state/trajectory/types.ts
```diff
@@ -92,3 +106,18 @@ export interface TimeUnits {
...
 }
+
+export interface HealthCheckTimeout {
+    [requestId: string]: NodeJS.Timeout;
+}
+
+type CONVERSION_NO_SERVER = typeof CONVERSION_NO_SERVER;
+type CONVERSION_SERVER_LIVE = typeof CONVERSION_SERVER_LIVE;
+type CONVERSION_ACTIVE = typeof CONVERSION_ACTIVE;
+type CONVERSION_COMPLETE = typeof CONVERSION_COMPLETE;
```
ShrimpCryptid, line 117: This is really strange to me. Could this and line 119 be reduced? 

`ConversionStatus = typeof CONVERSION_NO_SERVER | typeof CONVERSION_SERVER_LIVE | ...`

package.json
ShrimpCryptid, line 1: `npm` is warning about some dependency issues. May want to run `npm audit fix`.

src/state/trajectory/logics.ts
interim17: I think in production we won't need to reset these because everything will use the same Octopus backend, and we will be able to remove this config and just use the same one for everything stored at the top of the file.

src/state/trajectory/logics.ts
interim17: Indeed! I'll put a TODO flag at the top to make sure to cut it down before merging.

src/components/ConversionServerErrorModal/index.tsx
interim17: I've removed these and made a note for myself to look through the repo for these things and keep them from showing up.

src/state/trajectory/types.ts
interim17: logically no but i hate the way it looks without it :(

src/state/trajectory/types.ts
interim17: This is a good observation, to be honest I just copied the way it was done in viewer status.

But your suggestion will improve both. @meganrm do you see any reason that would cause an issue?

I'm going to leave the initial pipe because it looks too weird to me without it, but I'll change it if I have too lol

```
export type ConversionStatus =
    | typeof CONVERSION_INACTIVE
    | typeof CONVERSION_NO_SERVER
    | typeof CONVERSION_SERVER_LIVE
    | typeof CONVERSION_ACTIVE
    | typeof CONVERSION_COMPLETE;
    
    export type ViewerStatus =
    | typeof VIEWER_LOADING
    | typeof VIEWER_ERROR
    | typeof VIEWER_EMPTY
    | typeof VIEWER_SUCCESS;

File: src/components/ConversionServerErrorModal/index.tsx (hunk 11 above)
Comment (interim17) [reply src/components/ConversionServerErrorModal/index.tsx]: i think i can remove them all and it looks fine

**Comment by interim17** [comment]
A couple updates:

We were updating the viewer `status` to`VIEWER_IMPORTING` to disable some buttons in the app header but it was not working well because any trigger of `VIEWER_ERROR` status would reactive the buttons. We don't need it now that we have `conversionStatus` so I removed it entirely and also updated both of the status types based on @ShrimpCryptid 's feedback.

The redux `initializeConversion()` action is now handling the configuration and conversion so we were able to thin a little code from a few other places and it should be more semantic now. Thanks for the ongoing feedback!

File: src/state/trajectory/logics.ts (hunk 4 above)
Comment (interim17) [reply src/state/trajectory/logics.ts]: following from our other discussion yesterday there is still a call on page load, but we've moved things into the action formerly known as `convertFile` and removed the `onclick` function from the load file menu

File: src/state/trajectory/reducer.ts (hunk 6 above)
Comment (interim17) [reply src/state/trajectory/reducer.ts]: done

File: src/components/LoadFileMenu/index.tsx [hunk 16]
```diff
@@ -117,7 +111,9 @@ const LoadFileMenu = ({
             <Dropdown
                 menu={{ items, theme: "dark", className: styles.menu }}
                 placement="bottomRight"
-                disabled={isBuffering || viewerStatus === VIEWER_IMPORTING}
+                disabled={
+                    isBuffering || conversionStatus !== CONVERSION_INACTIVE
+                }
```
ShrimpCryptid, line 116: Should this also be disabled when the status is `SERVER_LIVE`?

ShrimpCryptid:
Copied over from another PR I accidentally commented on:

image
The cancel button is really faint when hovered. 

image
Also, the delete button is invisible until hovering over the text filename.

src/state/viewer/types.ts
```diff
@@ -42,14 +41,8 @@ export interface ViewerError {
...
 export type ViewerStatus =
-    | VIEWER_LOADING
-    | VIEWER_ERROR
-    | VIEWER_EMPTY
-    | VIEWER_SUCCESS
-    | VIEWER_IMPORTING;
+    | typeof VIEWER_LOADING
+    | typeof VIEWER_ERROR
+    | typeof VIEWER_EMPTY
+    | typeof VIEWER_SUCCESS;
```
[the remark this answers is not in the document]
ShrimpCryptid: This or the constants

src/components/LoadFileMenu/index.tsx
interim17: yes, this should be disabled if we are in any `conersionStatus` other than inactive

ShrimpCryptid:
```suggestion
                disabled={
                    isBuffering || conversionStatus === CONVERSION_ACTIVE
                }
```
Is this what you meant here?

ShrimpCryptid: image

So currently this disables the load button if the user has the page open to the file conversion page, even if no file is selected... This is maybe because I don't have a local Octopus instance running, but the server not communicating probably shouldn't prevent users from being able to access the other import methods?

interim17: that's how ux wants it

src/state/viewer/types.ts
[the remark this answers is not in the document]
interim17: [new issue opened](https://github.com/simularium/simularium-website/issues/455), will address later

ShrimpCryptid (approved):
Approving this, though I know there are more design discussions happening with UX. Feel free to re-request review if you do any other major changes.

I was able to test this locally (without Octopus) without issues, though I made a few notes about behavior and styling in the comments. It's worth having someone else who has a working Octopus install test this too.

src/state/trajectory/logics.ts
meganrm: have you tested initiating the config and then trying to load a networked file afterwards?

interim17: i did although i realize now it needs a little more work.
what i tested that works: currently if you navigate to `/import` and then to a trajectory url like `/viewer?trajFileName=endocytosis.simularium` it works.

but i now realize that, navigating to that url with the back button errors out, and if you hit `/import` then quickly navigate away while the server check is till pending there are some status setting side effects (i.e. the load trajectory button gets disabled while on the homepage for example) so i think i need to make this a bit more sophisticated in the logic and the router.

src/components/ConversionFileErrorModal/index.tsx
```diff
@@ -33,10 +32,9 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             onCancel={closeModal}
         >
             <div>
-                <div className={styles.redText}>
-                    {" "}
+                <div className={styles.warningText}>
                     {"We're sorry, there was a problem importing your file."}
```
meganrm, line 36: nit: you don't need the `{"`

meganrm (approved):

PR #453, opened 2023-12-09
Fix/viewer imports

meganrm:
Problem
=======
I noticed we were importing types from the types file, even though we've been exporting types from index. 

Solution
========
I updated the viewer with exports for the 3 types we were missing, and then edited the import statements here.


interim17:
LGTM

interim17 (approved):

ShrimpCryptid (approved):

PR #456, opened 2023-12-14
Fix/cancel file icon

interim17:
Problem
=======
A previous PR of mine left the styling incorrect on the icon that removes/cancels the uploaded file: 
    It had the default trash icon and also only appeared during hover.

There are couple little tweaks to button and select styling here, as well as some css organization.

Antd button was hard to target and change icon of, so I put `display: none` on that class and implemented a simple function on a `div` to replace it.

Old/incorrect:
Screenshot 2023-12-14 at 2 48 04 PM
New:
Screenshot 2023-12-14 at 3 27 08 PM
Design:
Screenshot 2023-12-14 at 2 49 11 PM

Steps to Verify:
----------------
Should look like the [design](https://xd.adobe.com/view/5f448466-eebd-4ad2-bdd8-314a55133403-0a17/screen/4e72b109-7f4d-472e-99d5-2e0e1d53569d/specs/), and when you click the red icon, any uploaded file should be removed.


src/containers/ConversionForm/index.tsx
```diff
@@ -205,6 +212,14 @@ const ConversionForm = ({
                     >
                         <Button type="default">Select file</Button>
                     </Upload>
+                    {fileToConvert && (
+                        <div
+                            className={styles.removeFileIcon}
+                            onClick={handleRemoveFile}
+                        >
+                            {CancelIcon}
+                        </div>
```
ShrimpCryptid, line 221: Clickable `div` elements can cause accessibility issues since they don't always play nice with screen readers. Consider changing this to a button!

src/containers/ConversionForm/index.tsx
```diff
@@ -24,6 +24,7 @@ import {
 import ConversionServerErrorModal from "../../components/ConversionServerErrorModal";
 import ConversionProcessingOverlay from "../../components/ConversionProcessingOverlay";
 import ConversionFileErrorModal from "../../components/ConversionFileErrorModal";
+import { ZoomIn as CancelIcon, DownCaret } from "../../components/Icons";
```
ShrimpCryptid, line 27: This is a neat trick but I feel like it would be better to have a separate `CancelIcon` SVG file. It seems to me like we could change `ZoomIn`'s icon at any point, and any icon that is not a plus (like a magnifying glass a plus with a bubble in it) would break the cancel icon.

interim17: It looked redundant in the Icons file to export the same antd icon twice, but you're totally right, this is more flexible/maintainable.

Applying inline styling in the Icons file makes that difference explicit, but then the styling is spread in two places. Neither is ideal, I just opted to export the same antd Icon for both `ZoomIn` and `Cancel`.

src/containers/ConversionForm/index.tsx
interim17: Good call, made the selector more specific and added a couple style rules to get the button right.

src/containers/ConversionForm/index.tsx
ShrimpCryptid: Did you try using the `CloseOutlined` icon that Ant provides? If so, were there any issues with it that made the `ZoomIn` icon a better fit?

interim17: I did, it's so subtle a difference that I'm not sure it _really_ matters, but the `CloseOutlined` has a little bevel on the ends of the lines that doesn't match the design.

`CloseOutlined` doesn't show the bevel on the antd icons website when you preview it, i think the discrepancy may be because we're using `"@ant-design/icons": "^4.0.6"`?

ShrimpCryptid: Hmm... unless the design explicitly forbids the bevel I'd be in favor of using the actual `CloseOutlined` icon rather than having to do a rotation hack, since it makes the code simpler and reduces the chance of unexpected consequences happening in the future. 

Does anyone else have any thoughts? @meganrm @ascibisz ?

interim17: Maybe applying the rotation as inline styling in the icons file would be a happy medium? Preserve the design and make the difference explicit to anyone reading that file who might change an import, or wondering why the same antd icon is being exported twice?

ShrimpCryptid (commented):
Adding a comment so Github tracks it, will re-review once the styling for the cancel icon is moved!

interim17:
> Adding a comment so Github tracks it, will re-review once the styling for the cancel icon is moved!

@ShrimpCryptid what do you think of this pattern? FWIW, it matches a couple of pairs of icons in the same file.

src/containers/ConversionForm/index.tsx
```diff
@@ -205,6 +212,14 @@ const ConversionForm = ({
                     >
                         <Button type="default">Select file</Button>
                     </Upload>
+                    {fileToConvert && (
+                        <button
+                            className={styles.removeFileIcon}
+                            onClick={handleRemoveFile}
+                        >
+                            {Cancel}
+                        </button>
```
ShrimpCryptid, line 221:
```suggestion
                        <button
                            aria-label="Remove file"
                            className={styles.removeFileIcon}
                            onClick={handleRemoveFile}
                        >
                            {Cancel}
                        </button>
```

ShrimpCryptid (approved):
LGTM! Seems to be working when tested locally. Thank you for the accessibility improvements 👍

meganrm (approved):

PR #461, opened 2024-01-29
Fix/bump react

interim17:
### Problem - React 18
Bumping react to be in sync with viewer.

A couple other dependencies seemed to need come along, `react-dom, @types/react-dom, @types/react, react-waypoint`. Also `children` is no longer a default prop on some components so I added it to a couple interfaces.

The only change that seems more significant is:
In React 18 the legacy render method is replaced with `createRoot`. 

Using the legacy API will cause the app to run as if its in React 17. We should use the new API so we are running the version we think we're running. The container for root could possible be null and [these docs](https://react.dev/blog/2022/03/08/react-18-upgrade-guide) show this non-null assertion:
```
const root = createRoot(container); // createRoot(container!) if you use TypeScript
```

But that seems not safe and the linter doesn't like it so I used a ternary check, thoughts?
```
const container: HTMLElement | null = document.getElementById(APP_ID);
container !== null
    ? renderApp(container)
    : console.log("Could not find container element with id " + APP_ID);
    ```

### CODE REVIEW

File: src/index.tsx [hunk 1]
```diff
@@ -91,6 +91,13 @@ console.log("Welcome to Simularium " + SIMULARIUM_BUILD_ENVIRONMENT + " build");
...
+    const root = createRoot(container);
+    root.render(<App />);
+};
+
+const container: HTMLElement | null = document.getElementById(APP_ID);
 
-renderApp();
+container !== null
+    ? renderApp(container)
+    : console.log("Could not find container element with id " + APP_ID);
```
ShrimpCryptid, line 103: re: your comment about the check here, this to me suggests that there is a possibility of the container being null, when it should (theoretically) never be. I'd slightly prefer the `!` syntax because it seems like hitting an error will be more informative than the blank screen and a console message.

interim17: trying this instead:
```
const container: HTMLElement | null = document.getElementById(APP_ID);
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
const root = createRoot(container!);
root.render(<App />);
```

ShrimpCryptid (approved):

ascibisz (approved):
I tested it locally and it works for me! Reading through the changes, it all makes sense to me as well, although I don't necessarily trust my own judgement on assessing the changes in `src/index.tsx`

PR #462, opened 2024-02-02
Autoconversion Fixes

ascibisz:
## Problem
This is a somewhat large PR that solves three needs:
1) Need to send the file to Octopus to get it converted.
2) Receive the converted file from Octopus.
3) Front end errors arising from switching backends.

@interim17 did most of the work for this PR, @ascibisz helped with some debugging and happened to be the one to open the PR because she had the time to

 Solution
-------------------

### Convert the file
Redux action/logic  being triggered to send the file to Octopus. As always with redux there is much boiler plate, the action is in `convertFileLogic` which does the sending. 

### Receive the converted file
To receive the file we need a configured viewer at the ready, so Joe refactored the routes so that the conversion form is just a component that is just visible in front of the viewer when the user opens it. This made some of the communication with Octopus simpler. We're always at `/viewer` ready to receive a file, and we handle navigation in and out of the conversion form with `conversionState`.

### Errors
Alli then did some miscellaneous debugging, as well as switching the autoconversion feature branch to use octopus for everything, not just the autoconversion point. The infrastructure team is actively working on the octopus deployment, so rather than continuing to have a weird state of partially using octopus and  partially using simularium-engine, we think it makes sense to have this feature branch ready to fully embrace octopus.


src/components/Icons/index.tsx
```diff
@@ -48,6 +48,8 @@ export const StepForward = <img src={ClockwiseArrow} />;
 export const StepBack = <img src={CounterClockwiseArrow} />;
 export const BetaTag = <img src={Beta} style={{ width: "42px" }} />;
 export const UpRightArrow = <ArrowLeftOutlined rotate={135} />;
+export const OrthographicCamera = <img src={Orthographic} />;
+export const PerspectiveCamera = <img src={Perspective} />;
```
ascibisz, line 52: Somehow (probably during merge conflict / resolution), these lines are missing in the autoconversion feature branch, even though they've been merged into main. I added them back in here, but they have nothing to do with our overall change set

src/state/trajectory/logics.ts
```diff
@@ -547,6 +542,46 @@ const setConversionEngineLogic = createLogic({
...
+                engineType
+            )
+            .catch((err: Error) => {
+                console.error(err);
+            });
+        done();
+    },
+    type: CONVERT_FILE,
+});
+
```
interim17, line 584: This is the logic that actually sends the file to Octopus for conversion. Because we are already at `/viewer` we are ready for Octopus to receive the converted file as soon as its ready.

src/containers/Simularium/index.tsx
```diff
@@ -204,10 +213,14 @@ class App extends React.Component<AppProps, AppState> {
...
             clearSimulariumFile,
             setError,
+            conversionStatus,
         } = this.props;
         return (
             <Layout className={styles.container}>
                 <div ref={this.interactiveContent}>
+                    {conversionStatus !== CONVERSION_INACTIVE && (
+                        <ConversionForm />
+                    )}
```
interim17, line 223: Now that the conversion process doesn't have its own path, we are rendering the conversion form here in `Simularium` because the viewer itself doesn't really need to know anything about conversion state, it just needs to handle the converted trajectory. Styling is also much easier here above the `ViewerPanel` layout.

src/containers/ConversionForm/custom-request.ts
```diff
@@ -1,19 +1,21 @@
...
+        fileContents: string,
+        fileName: string
+    ) => ReceiveFileToConvertAction,
     setError: (error: ViewerError) => void,
     rcRequest?: UploadRequestOption
 ) => {
     try {
         const unpackedFileText = await fileToConvert.originFileObj.text();
-        receiveFileToConvert(unpackedFileText);
+        const fileName = fileToConvert.name.replace(/\.[^/.]+$/, "");
```
meganrm, line 17: why is this needed? can you add a comment about what this regex selects for. You can also always use a named variable for your regex, like `const whiteSpace = RegEX; string.replace(whiteSpace, "")`

src/containers/ConversionForm/index.tsx
```diff
@@ -70,21 +71,28 @@ const ConversionForm = ({
     receiveFileToConvert,
     initializeConversion,
     conversionStatus,
+    convertFile,
+    setConversionStatus,
 }: ConversionProps): JSX.Element => {
     const [fileToConvert, setFileToConvert] = useState<UploadFile | null>();
     const [engineSelected, setEngineSelected] = useState<boolean>(false);
+    const [isProcessing, setIsProcessing] = useState<boolean>(false);
     const [serverDownModalOpen, setServerIsDownModalOpen] =
```
frasercl, line 80: nit (and not even one in the diff!): inconsistent names: `serverDownModalOpen` vs `setServer*Is*DownModalOpen`

src/containers/ConversionForm/index.tsx
```diff
@@ -70,21 +71,28 @@ const ConversionForm = ({
...
     useEffect(() => {
+        // on page load assume server is down until we hear back from it
+        setConversionStatus({ status: CONVERSION_NO_SERVER });
         initializeConversion();
     }, []);
 
-    // TODO delete after development, useEffect to log a change in server health
     useEffect(() => {
-        console.log(conversionStatus);
+        // this it to account for the server going down while a conversion is in process
```
frasercl, line 91:
```suggestion
        // this is to account for the server going down while a conversion is in process
```

src/containers/ConversionForm/index.tsx
```diff
@@ -70,21 +71,28 @@ const ConversionForm = ({
     receiveFileToConvert,
     initializeConversion,
     conversionStatus,
+    convertFile,
+    setConversionStatus,
 }: ConversionProps): JSX.Element => {
     const [fileToConvert, setFileToConvert] = useState<UploadFile | null>();
     const [engineSelected, setEngineSelected] = useState<boolean>(false);
```
frasercl, line 78: Also not technically part of this PR (so feel free to ignore me!), but this looks like derived state to me. `engineSelected` tracks whether another component of state is set, and may become out of sync if not properly updated (e.g. `engineSelected: false` while `store.trajectory.processingData.engineType: "cytosim"`). This particular case is probably too simple to cause issues, but it still might be slightly better practice to select `engineType` out of the store and null-check it.

frasercl (commented):
Only very small comments from me. I don't have the full perspective of how the app and octopus interact, but from what I understand this looks good!

src/containers/ConversionForm/index.tsx
interim17: Good catch, I think I now prefer serverErrorModalOpen anyway as it matches the modal component better

src/containers/Simularium/index.tsx
```diff
@@ -77,6 +85,7 @@ class App extends React.Component<AppProps, AppState> {
     public simulariumController: SimulariumController | undefined;
     private interactiveContent = React.createRef<HTMLDivElement>();
     private endDragover = 0;
+    private conversionCheckInterval: NodeJS.Timeout | null = null;
```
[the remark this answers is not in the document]
interim17: Nope, needs removing

src/index.tsx
```diff
@@ -37,14 +33,8 @@ function useLocationChange() {
...
-        if (location.pathname === IMPORT_PATHNAME) {
+        if (location.pathname === VIEWER_PATHNAME) {
             batch(() => {
-                dispatch(setIsPlaying(false));
-                dispatch(setConversionStatus({ status: CONVERSION_NO_SERVER }));
-            });
-        } else if (location.pathname === VIEWER_PATHNAME) {
-            batch(() => {
-                dispatch(setIsPlaying(false));
                 dispatch(setStatus({ status: VIEWER_EMPTY }));
```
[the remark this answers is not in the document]
interim17: Nope!

src/containers/ConversionForm/index.tsx
interim17: Love it, seems like this works and can remove the state variable:
`const engineSelected = !!conversionProcessingData.engineType;`

meganrm (approved):

frasercl (approved):
Looks good!

PR #463, opened 2024-02-13
Feature/og tags

### LINKED ISSUE #335: Update landing page meta tags (image, title, description, etc.)
## What needs to happen?
Update the open graph (og) meta tags: 
`og:title`: Simularium
`og:description`: The Simularium Viewer makes it easy to share and interrogate interactive 3D visualizations of biological simulation trajectories and related plots directly in a web browser.
`og:image`: tbd, maybe this? 

Screen Shot 2022-04-04 at 2 50 18 PM

## Why should we do this?
On social media platforms (Twitter, Facebook, LinkedIn, Slack, etc.), when a person type the link "simularium.allencell.org", a social card will be generated automatically and displayed together with the link in the post/message.

We'd like to have relevant information displayed properly to encourage interactions.

Examples:
 

or 

During regular maintenance update; ideally before Communication initiates the Google Ads for Simularium at the end of April.

interim17:
### Use open graph meta tags to make homepage links load previews on social media

This is a branch that will close #335 

Open to feedback now but will probably wait until Thao is back to run things by her as well.

I used the text and image provided in the issue page.

I used this website to do initial previews: https://www.opengraph.xyz/ that looked good for facebook, twitter, linkedin. And then made an `ngrok` tunnel to my localhost so that I could see the meta tag previews in slack messages to myself.
Screenshot 2024-02-12 at 11 09 54 PM


ShrimpCryptid (approved):

webpack/index.template.html
```diff
@@ -13,7 +13,36 @@
...
+
+    <!-- HTML Meta Tags -->
     <title>Simularium</title>
+    <meta name="description"
+        content="The Simularium Viewer makes it easy to share and interrogate interactive 3D visualizations of biological simulation trajectories and related plots directly in a web browser.">
+    
+    <!-- Facebook Meta Tags -->
+    <meta property="og:title" content="Simularium" />
+    <meta property="og:url" content="https://simularium.allencell.org/">
+    <meta property="og:image" content="../src/assets/meta-tag.png">
```
frasercl, line 25: Does this relative URL get transformed at all in the build process, and/or does the image get put in the build directory? The image may not appear in the hosted site if it's dependent on navigating out of the build directory and into src.

interim17: That's a good question. When I was first looking at this I thought we should put the asset into s3 and use an absolute URL but @meganrm said a relative URL was the more modern way to handle it.

I think this is a good catch. I assumed it was working because my preview looking okay, but I'm looking for the `png` in the build folder and not seeing it.

I think with our build rules we will need to do an import to ensure that the `png` comes along for the ride, and the easiest way to do that will be to inject these tags progamatically. Going to poke around looking at options for that.

interim17: RE: doing this dynamically/in React: because web-crawlers don't execute javascript it seems likely we will either need a static URL directly in the HTML or have to implement server-side rendering, or get the back end more involved in this. For SSR there are some libraries but that seems like a lift for this feature.

I think just uploading the necessary images to S3 seems like the most straightforward way to handle this. Would you agree @frasercl ?

frasercl (approved):
If the next thing y'all want to do is try this out in production, that sounds like a good next step to me!

PR #467, opened 2024-02-16
React bump fixes

interim17:
Bumping React messed with a few small things in auto-conversion world, mostly little styling tweaks.

The only functional changes were in the antd `Upload` component.

1) `customRequest` and `handleFileSelection` had issues when selecting a file. Previously `customRequest` and `onChange` each received props and worked in sync. After bumping React, `customRequest` was receiving undefined arguments and throwing errors, I think because React 18 batches state updates and this introduced a delay. Moving the call of `customRequest` inside of `handleFileSelection` fixed it.
Screenshot 2024-02-19 at 10 28 33 AM

2) Also had to add the custom render function for the `fileList` because it started rendering an unwanted tooltip (that was weirdly full of the DOCUMENT html markup...).

Screenshot 2024-02-19 at 10 30 09 AM

Note: a lot of things (including this code) were piling up in [this PR](https://github.com/simularium/simularium-website/pull/462), talked with @ascibisz and RE: dev chat, we are breaking it up into at least two pieces that are more targeted. I think this is wise and will make review easier, the downside in this case is that we can't complete the auto-conversion process with current viewer/octopus on this branch. You need the rest of the changes to take a file through the whole conversion process.

This PR can be reviewed by 
- checking that your screen looks like the screenshots below
- going to `/import` and uploading a file, there should be no error
- hover the cursor over that file and seeing if there is an erroneous tooltip
- turn octopus off OR choose an incorrect file type to see the error modal stylings look ok

Screenshot 2024-02-16 at 3 11 22 PM
Screenshot 2024-02-16 at 3 12 09 PM
Screenshot 2024-02-16 at 3 12 19 PM
Screenshot 2024-02-16 at 3 12 35 PM

### DIFF
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -31,13 +31,15 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             centered
             onCancel={closeModal}
         >
-            <p>
-                <div className={styles.warningText}>
+            <div>
+                <p className={styles.warningText}>
                     {"We're sorry, there was a problem importing your file."}
-                </div>
-                You may want to double check that the file you selected is a
-                valid {engineType} file and try again. For further assistance,
-                please visit
+                </p>
+                <p>
+                    You may want to double check that the file you selected is a
+                    valid {engineType} file and try again. For further
+                    assistance, please visit
+                </p>
                 <a
                     href="https://forum.allencell.org/"
                     target="_blank"
@@ -45,7 +47,7 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
                 >
                     The Allen Cell Discussion Forum.
                 </a>
-            </p>
+            </div>
         </CustomModal>
     );
 };
--- a/src/components/ConversionFileErrorModal/index.tsx
+++ b/src/components/ConversionFileErrorModal/index.tsx
@@ -31,7 +31,7 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             centered
             onCancel={closeModal}
         >
-            <div>
+            <p>
                 <div className={styles.warningText}>
                     {"We're sorry, there was a problem importing your file."}
                 </div>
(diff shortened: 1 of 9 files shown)


src/components/ConversionFileErrorModal/index.tsx
```diff
@@ -31,7 +31,7 @@ const ConversionFileErrorModal: React.FC<ConversionFileErrorModalProps> = ({
             centered
             onCancel={closeModal}
         >
-            <div>
+            <p>
                 <div className={styles.warningText}>
                     {"We're sorry, there was a problem importing your file."}
                 </div>
```
ShrimpCryptid, line 37: Nit: Placing a `div` inside of a `p` element seems odd to me... Consider changing the inner div to a `span` element instead?

src/containers/ConversionForm/index.tsx
```diff
@@ -144,6 +145,16 @@ const ConversionForm = ({
...
+    const renderUploadFile = () => {
+        if (fileToConvert) {
+            return (
+                <span className={styles.renderedFileName}>
+                    {fileToConvert.name}
+                </span>
+            );
+        }
+    };
+
```
ShrimpCryptid, line 157: Should this have a return type if `fileToConvert` is null/undefined?

src/components/ConversionFileErrorModal/index.tsx
interim17: Ok fair point, however the span is inline and i want this to be a block element for the styling. I could change it back to being div? 

I'm realizing I didn't really think this through all the way, but my intention in changing it from a div to a p was that it is more accessible to have p tags where there is text

src/containers/ConversionForm/index.tsx
interim17: Changed to always return the span and use a ternary to grab the filename if there is a `fileToConvert`

src/containers/ConversionForm/index.tsx
```diff
@@ -224,7 +219,9 @@ const ConversionForm = ({
                         </button>
                     )}
                 </div>
-                <Divider orientation="right" orientationMargin={400}> </Divider>
+                <Divider orientation="right" orientationMargin={400}>
+                    {" "}
+                </Divider>
```
ShrimpCryptid, line 224: Nit: 
```suggestion
                <Divider orientation="right" orientationMargin={400}/>
```

ShrimpCryptid: Not sure if this requires child props but seems like it would be a little cleaner

ShrimpCryptid (approved):
LGTM! Thanks for addressing my feedback. :')

PR #469, opened 2024-02-22
Fix/nav tooltips

meganrm:
Problem
=======
On staging right now, it's hard to click the "share" button because the target is covered by the tooltip target 

Solution
========
Adjusted the tooltip positioning so they don't cover the button target 

## Type of change

Steps to Verify:
----------------
1. load a local simularium file 
2. click "share"
3. should be easer than it is on staging

### DIFF
--- a/src/components/DownloadTrajectoryMenu/index.tsx
+++ b/src/components/DownloadTrajectoryMenu/index.tsx
@@ -2,7 +2,12 @@ import React from "react";
 import { Button, Tooltip } from "antd";
 import { ISimulariumFile } from "@aics/simularium-viewer";
 
-import { DATA_BUCKET_URL, TOOLTIP_COLOR, TOOLTIP_DELAY } from "../../constants";
+import {
+    DATA_BUCKET_URL,
+    NAV_BAR_TOOLTIP_OFFSET,
+    TOOLTIP_COLOR,
+    TOOLTIP_DELAY,
+} from "../../constants";
 import { NetworkedSimFile, LocalSimFile } from "../../state/trajectory/types";
 import { Download } from "../Icons";
 
@@ -55,7 +60,7 @@ const DownloadTrajectoryMenu = ({
 
     const isDisabled = !fileIsLoaded() || isBuffering;
     // disabled buttons are wrapped in a span which changes the tooltip offset
-    const tooltipOffset = isDisabled ? [-30, -20] : [0, -8];
+    const tooltipOffset = isDisabled ? [-30, -20] : NAV_BAR_TOOLTIP_OFFSET;
 
     return (
         <div className={styles.container}>
--- a/src/components/DownloadTrajectoryMenu/index.tsx
+++ b/src/components/DownloadTrajectoryMenu/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Button, Tooltip } from "antd";
 import { ISimulariumFile } from "@aics/simularium-viewer";
 
-import { DATA_BUCKET_URL, TOOLTIP_COLOR } from "../../constants";
+import { DATA_BUCKET_URL, TOOLTIP_COLOR, TOOLTIP_DELAY } from "../../constants";
 import { NetworkedSimFile, LocalSimFile } from "../../state/trajectory/types";
 import { Download } from "../Icons";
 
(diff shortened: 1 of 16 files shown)


src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -54,7 +54,8 @@ const DownloadTrajectoryMenu = ({
     };
 
     const isDisabled = !fileIsLoaded() || isBuffering;
-    const tooltipOffset = isDisabled ? [25, -30] : [40, -18];
+    // disabled buttons are wrapped in a span which changes the tooltip offset
+    const tooltipOffset = isDisabled ? [-30, -20] : [0, -8];
```
meganrm, line 58: the fix was to lower the tooltip

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -41,19 +45,25 @@ const ShareTrajectoryButton = ({
...
-                    isDisabled
+                    !simulariumFile.name
                         ? "Load a model to perform this action"
                         : "Share trajectory"
                 }
                 placement="bottomLeft"
                 color={TOOLTIP_COLOR}
                 align={{ offset: tooltipOffset }}
+                mouseEnterDelay={0.5}
+                trigger={isSharing ? [] : ["hover"]}
```
ShrimpCryptid, line 56: Also consider adding "focus" to the list of triggers so that tab selection of buttons will show the tooltip as well

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -41,19 +45,25 @@ const ShareTrajectoryButton = ({
...
+                onOpenChange={(visible) => {
+                    setTooltipVisible(visible);
+                }}
```
ShrimpCryptid, line 59: How does `onOpenChange` work when you're also controlling visibility via `open`? Will it still call the callback when the user stops hovering on the tooltip?

src/components/DownloadTrajectoryMenu/index.tsx
toloudis: Since these numbers show up in two places, is there a common place to store them that is maybe related to the offset created by the span you reference in the comment?

toloudis (approved):
LGTM, just one DRY-type question

src/components/ShareTrajectoryButton/index.tsx
```diff
@@ -19,15 +19,19 @@ const ShareTrajectoryButton = ({
...
 }: ShareTrajectoryButtonProps): JSX.Element => {
     const [isSharing, setIsSharing] = React.useState(false);
+    const [tooltipVisible, setTooltipVisible] = React.useState(false);
 
     const trajectoryIsSharable = isOnlineTrajectory(location.href);
 
     const handleShare = () => {
         setIsSharing(!isSharing);
+        // this is to ensure the tooltip closes when the button is clicked
+        setTooltipVisible(false);
```
interim17, line 29: This doesn't work all the time, if you move the cursor on and click quickly `handleShare` runs during the `mouseOver` delay and then `onOpenChange` opens the tooltip while the modal is open. Suggestion below seems to fix it, probably other ways as well.
```suggestion
         if (isSharing) return;
        setTooltipVisible(false);
```

src/components/ShareTrajectoryButton/index.tsx
interim17: @ShrimpCryptid  The callback seems to be triggered by the mouse event and not the change in state at least when I played with it and logged it all out.

interim17, line 59: If you move the cursor on the button quickly and click right away `handleShare` runs during the `mouseOver` delay and then `onOpenChange` opens the tooltip while the modal is open. Suggestion below seems to fix it, probably other ways as well. Sorry if you get two notifications, put this comment/suggestion on the wrong line a moment ago.
```suggestion
                onOpenChange={(visible) => {
                     if (isSharing) return;
                    setTooltipVisible(visible);
                }}
```

interim17 (approved):
Left a suggestion about sometimes wonky tooltip behavior. Non-blocking. :)

src/components/ShareTrajectoryButton/index.tsx
meganrm: I'll do a change like this, but I actually haven't been able to get this to happen

interim17: Weird, can show a video if you like, maybe its a browser thing if you're on Safari? I'm on Chrome...

PR #474, opened 2024-03-07
Feature/embed

meganrm:
Time estimate or Size
=======
medium

Problem
=======
We need a version of the app to link to when someone wants to embed their trajectory in a website. This will still need to be updated with the design. 

closes https://github.com/simularium/simularium-planning/issues/188

Solution
========
Made an embed path that renders the app without the header and creates an overlay for future styling, overlay info. I made an assumption that like the notebook version of the app we'd get rid of the header and place anything we want (the title, maybe a share link) over the full screen view (the way that YouTube does). So I added a minimal overlay that can be styled and added to. For now I've rendered the title with no styling.

## Type of change

Steps to Verify:
----------------
1. `npm start`
2. navigate to `http://localhost:9001/embed?trajFileName=endocytosis.simularium`
3. Compare to `http://localhost:9001/viewer?trajFileName=endocytosis.simularium`

Screenshots (optional):
-----------------------
Screenshot 2024-03-07 at 1 02 26 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9323,9 +9323,9 @@
             "dev": true
         },
         "node_modules/fast-glob": {
-            "version": "3.2.12",
-            "resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.2.12.tgz",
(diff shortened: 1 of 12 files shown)


interim17:
Screenshot 2024-03-08 at 11 37 41 AM

PR #475, opened 2024-03-07
fix the html loader result

meganrm:
Time estimate or Size
=======
small

Problem
=======
the `html-loader` was exporting a module instead of just the png for the og tag image 

Solution
========
I added the `esModules: false`
While I was investigating this, I realized we already had `HtmlWebpackPlugin` which you can use to set metatags that don't need any processing done to them, so I also moved all the tags without paths into there. 

Verify
=======
I deployed it on gh-pages and you can see the preview: https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fsimularium.github.io%2Fsimularium-website

## Type of change

I was able to test this by momentarily overwriting a nightly build and the image loaded.

### DIFF
--- a/webpack/plugins.js
+++ b/webpack/plugins.js
@@ -34,6 +34,7 @@ const getBasePlugins = (dist, env) => {
                 ["og:description"]:
                     "The Simularium Viewer makes it easy to share and interrogate interactive 3D visualizations of biological simulation trajectories and related plots directly in a web browser.",
                 ["og:url"]: "https://simularium.allencell.org",
+                ["og:type"]: "website",
                 ["og:image:type"]: "image/png",
                 ["og:image:width"]: "2048",
                 ["og:image:height"]: "1115",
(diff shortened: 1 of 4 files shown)


webpack/webpack.config.js
```diff
@@ -134,6 +134,9 @@ module.exports = ({ analyze, env, dest = "dist" } = {}) => ({
             {
                 test: /\.html$/i,
                 loader: "html-loader",
+                options: {
+                        esModule: false,
+                },
```
meganrm, line 139: this is the fix

webpack/plugins.js
```diff
@@ -26,6 +26,26 @@ const getBasePlugins = (dist, env) => {
...
+                ["og:image:width"]: "2048",
+                ["og:image:height"]: "1115",
+                ["og:image:alt"]:
+                    "A view of the Simularium viewer showing a 3D scene of molecular interactions, with data plots on the right side of the screen, and a menu of the visible agents on the left side.",
+                ["twitter:card"]: "summary_large_image",
+                ["twitter:title"]: "Simularium",
+                ["twitter:description"]:
+                    "The Simularium Viewer makes it easy to share and interrogate interactive 3D visualizations of biological simulation trajectories and related plots directly in a web browser.",
+                ["twitter:url"]: "https://simularium.allencell.org/",
+            },
```
meganrm, line 48: these are all exactly the same as the tags were in the html template

ShrimpCryptid (approved):

interim17 (approved):

PR #476, opened 2024-03-08
 Fix/icomoon

interim17:
### Time to review:
_Small < 10 minutes_

### Problem:
Icomoon font glyphs had not been updated in a long time and they were written into the code base in hard to understand ways. New icons had been added in other ways, with conflicting design approaches.

### Solution:
Create new set of glyphs, define each glyph as a css variable, apply those variables via global style selectors.
I added some explanatory comments about how to use this styling pattern because I found it confusing at first, which led to me adding icons in suboptimal ways. Updating the font was easy! I just didn't realize...

_Part of preliminary changes/maintenance to fixing all styles on the playback and camera control buttons._

### Other changes:
Moved global style imports to `src/style.css` why were they in the `SideBar` component?
Removed some unused style rules, cleaned up an import or two

### To review:
- Confirm all the icons match current guidance in the [figma guide](https://www.figma.com/file/u4lZYuVwbsFpzBikgwAVRb/Master-components-Simularium?type=design&node-id=0-1&mode=design), the relevant icons are all in playback controls, camera controls, and the highlighting stars in the side panel. (Rotate and home/reset were wrong, orthographic/perspective were correct icon but way wrong size)
- Confirm that the styling/spacing is not egregiously worse.
_Note:_  The buttons and the icon positioning inside them _do not_ match the style guide on staging, production, or this branch... I have a change-set lined up for this, out of scope for this PR to fix it all. If we can get these icons organized and available for styling we can fix this the right way.
Screenshot 2024-03-08 at 9 52 13 AM

Screenshot 2024-03-08 at 9 52 06 AM

Screenshot 2024-03-08 at 9 51 59 AM


src/components/AgentTree/style.css
```diff
@@ -44,11 +44,6 @@
     height: 40px;
 }
 
-.star-icon::after {
-    content: "\e902";
-    color: var(--dark-theme-sidebar-text);
-}
-
```
interim17, line 51: This rule was unused as far as I could tell.

src/style.css
```diff
@@ -1,3 +1,8 @@
+
+@import "./styles/colors.css";
+@import "./styles/constants.css";
+@import "./styles/icons.css";
+
```
interim17, line 5: These imports were in a seemingly random component file, makes more sense here.

src/style.css
```diff
@@ -73,59 +78,81 @@ canvas:focus {
...
+If new glyphs are added/glyphs are changed: update the css vars 
+in /styles/icons.css, and define the appropriate::after selector 
+here, apply icon using "name-icon" class name in conjunction 
+with "icon-moon" class name.
+ */
+
+ /* Current in-use icons */
+.star-empty-icon::after {
+    content: var(--star-empty);
+    color: var(--warm-gray)
```
interim17, line 106: Existing rules defining global style rules were all unused with a comment that they were just for storage. Icons in use had their content  provided via code string  in stylesheets (i.e. content: "\e909") which gave no indication what the content was. This pattern defined here makes a lot of sense! I think we should use it all the time, not just for storing unused icons.

src/components/CameraControls/index.tsx
```diff
@@ -4,7 +4,7 @@ import classNames from "classnames";
 import { useHotkeys, useIsHotkeyPressed } from "react-hotkeys-hook";
 
 import { TOOLTIP_COLOR } from "../../constants/index";
-import Icons from "../Icons";
+import { ZoomIn, ZoomOut } from "../Icons";
```
interim17, line 7: This is how we do it in most parts of the repo, changing to match.

src/components/CameraControls/index.tsx
```diff
@@ -189,7 +189,7 @@ const CameraControls = ({
                                 className={classNames([
                                     "icon-moon",
                                     "anticon",
-                                    styles.rotate,
+                                    "rotate-icon",
```
interim17, line 192: This points to global style rule, which references css variable, rather than pointing at component style sheet that directly references glyph code in content property.

src/components/Icons/index.tsx
```diff
@@ -70,6 +60,4 @@ export default {
     Link,
     Download,
     LoopOutlined,
-    OrthographicCamera,
-    PerspectiveCamera,
 };
```
interim17, line 63: Removed icons are now in icomoon font.

src/components/CameraControls/index.tsx
```diff
@@ -151,7 +151,7 @@ const CameraControls = ({
                 >
                     <Button
                         className={styles.btn}
-                        icon={Icons.ZoomIn}
+                        icon={ZoomIn}
                         onClick={zoomIn}
```
ShrimpCryptid, line 155: Nit: Should these buttons have an `aria-label` or other readable text label, or is that included with the icon itself?

src/styles/icons.css
ShrimpCryptid, line 1: Love this change!

ShrimpCryptid (approved):
Looks good, left a comment about accessibility for the various buttons and icons.

src/components/CameraControls/index.tsx
interim17: Great point, fits better in the next set of changes, noted!

meganrm (approved):
This looks good

PR #477, opened 2024-03-11
fix/hide tooltips after click

### LINKED ISSUE #297: Resolve Issues with Tooltips (and labels)
## Use Case
1. Users cannot easily select/deselect stars and checkboxes because tooltips block ability to interact with the one directly above
2. Users need to be able to quickly identify what elements make up the agent panel and actions they can perform.
3. The star with text “SHOW TYPE” meant to serve as labels to columns below, do not read as labels to columns and cause confusion.

## Acceptance Criteria

1. Immediate timeout on tooltips when exiting star/checkbox
2. New tooltip does not show after a user clicks star/checkbox (only show on initial hover/focus)
3. Line with star icon, SHOW, TYPE is completely removed
4. "All" checkbox label is changed to "All Agent Types"

## Details

Design Link: https://xd.adobe.com/view/75b6a3c8-f650-443b-bccc-d1fcd134a35e-20e5/

interim17:
Time to review:
_Small: 5 minutes_

Problem
=======
Closes #297 specifially criteria 1 and 2
@meganrm already resolved criteria 3 and 4 in previous work.

Solution
========
We need to manage the tooltip open/closed state via a local state so we can keep it closed after a click event, and only update the title text while the tooltip is closed so that the new text doesn't flicker before the tooltip has a chance to hide.

Steps to Verify:
----------------
1. Try the different hide and highlight checkboxes, observe the tooltip behavior: after clicking it should hide until the mouse has left and re-entered.

https://github.com/simularium/simularium-website/assets/24981838/2efef448-aa93-417f-a6eb-67b71e0063e2

### DIFF
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -74,7 +74,7 @@ const Checkbox = (props: CheckboxTypeProps): JSX.Element => {
                 {...childProps}
                 onChange={(e) => {
                     setShowTooltip(false);
-                    childProps.onChange ? childProps.onChange(e) : null;
+                    childProps.onChange?.(e);
                 }}
                 onMouseEnter={handleMouseEnter}
                 onMouseLeave={handleMouseLeave}
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -1,4 +1,4 @@
-import * as React from "react";
+import React, { useState } from "react";
 import { Checkbox as AntdCheckbox, Tooltip } from "antd";
 import { CheckboxProps } from "antd/lib/checkbox";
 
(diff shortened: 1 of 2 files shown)


toloudis (approved):
LGTM

src/components/Checkbox/index.tsx
```diff
@@ -44,14 +61,24 @@ const Checkbox = (props: CheckboxTypeProps): JSX.Element => {
...
             color={TOOLTIP_COLOR}
+            open={showTooltip}
+            onOpenChange={updateTooltipTitle}
         >
-            <AntdCheckbox {...childProps} />
+            <AntdCheckbox
+                {...childProps}
+                onChange={(e) => {
+                    setShowTooltip(false);
+                    childProps.onChange ? childProps.onChange(e) : null;
```
frasercl, line 77: Can use TS's optional chaining here: `childProps.onChange?.(e);`

src/components/StarCheckbox/index.tsx
```diff
@@ -14,29 +14,53 @@ const StarCheckbox = ({
...
                         checked={checked}
                         type="checkbox"
-                        onChange={(e: any) =>
-                            onChange ? onChange(e as CheckboxChangeEvent) : null
-                        }
+                        onChange={(e: any) => {
+                            setShowTooltip(false);
+                            onChange
+                                ? onChange(e as CheckboxChangeEvent)
+                                : null;
```
frasercl, line 59: Here too

frasercl (approved):
I was tricked into thinking `tooltipTitle` was derived state for a sec, but it is actually capturing something significant. This looks nice!

src/components/StarCheckbox/index.tsx
interim17: Thank you!

PR #478, opened 2024-03-13
feature/viewport button component

interim17:
### Time to review:
_Medium_
_Design review with UX complete._

### Problem
Viewport buttons (camera and playback controls) do not match current style guidance.
The code for these buttons and their stylesheets is not DRY.

### Solution
Create `ViewportButton` component. Each `ViewportButton` wraps a button in a tooltip and takes a short list of custom props that get it styled correctly. Implementing components (playback controls, camera controls, forthcoming record movies component) are concise and readable.

Relevant re-organizing and re-naming is ongoing in `colors.css`. It's a pain, but if any reviewers would like to use it I have a python script that scans the repo for unused css vars and css properties that use variables that aren't actually defined (there were several!) that is making this somewhat less painful T_T

* Bug fix (non-breaking change which fixes an issue) (style issues)
* New feature (non-breaking change which adds functionality) (wrapper component)

Steps to Verify:
----------------
1. Confirm styling is correct for buttons in default, hover, and active states.

Screenshots (optional):
-----------------------
Old:
Screenshot 2024-03-13 at 12 06 29 PM
New:
Screenshot 2024-03-13 at 12 06 47 PM

### DIFF
--- a/src/components/ViewportButton/index.tsx
+++ b/src/components/ViewportButton/index.tsx
@@ -1,5 +1,5 @@
 import React, { ReactNode } from "react";
-import { Button, ButtonProps, Tooltip } from "antd";
+import { Button, ButtonProps, Tooltip, TooltipProps } from "antd";
 import classNames from "classnames";
 
 import { TOOLTIP_COLOR } from "../../constants";
@@ -7,7 +7,7 @@ import styles from "./style.css";
 
 interface ViewportButtonProps extends ButtonProps {
     tooltipText?: string;
-    tooltipPlacement?: "top" | "bottom" | "left" | "right";
+    tooltipPlacement?: TooltipProps["placement"];
     icon?: ReactNode | string; // When using an icomoon icon, pass the icon name as defined in selectors in src/styles.css
     radioGroupPosition?: "top" | "bottom";
     clickHandler?: () => void;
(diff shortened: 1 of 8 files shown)


src/components/ViewportButton/index.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { ReactNode } from "react";
+import { Button, ButtonProps, Tooltip } from "antd";
+import classNames from "classnames";
+
+import { TOOLTIP_COLOR } from "../../constants";
+import styles from "./style.css";
+
+interface ViewportButtonProps extends ButtonProps {
+    tooltipText?: string;
+    tooltipPlacement?: "top" | "bottom" | "left" | "right";
```
meganrm, line 10: is it possible to also extend the tooltip props instead of re-writing it here?

meganrm (commented):
this looks great, nice job getting rid of a bunch of code. just one question about you merging some prop interfaces

meganrm (approved):

toloudis (approved):
excellent DRYing.  Good use for a reusable component wrapper.

src/components/ViewportButton/index.tsx
interim17: yes it is! didn't realize how till you suggested it, thanks.

PR #483, opened 2024-03-26
Feature/record movie UI

### LINKED ISSUE #327: Implement Record Movies
- [ ] Firefox support
- [ ] messaging for Firefox users (UX)

## Use Case

Today, there is not a quick and easy way for users to record a movie to use in talks/presentations or to share with others.

## Acceptance Criteria

- Implement recording functionality
- Implement buttons w/ icons to accommodate recording (start and stop with tooltips)
- Style UI when recording is on as shown in mockup (includes a real time duration counter)
- Add confirmation modal with ability to download movie

## Details

[Design Link](https://xd.adobe.com/view/87f2942c-60fa-40c6-a5b8-774713e402a4-d2a1/)
[Pulse Animation Example](https://codepen.io/FlorinPop17/pen/drJJzK)
[Record Icon](https://drive.google.com/file/d/1RbNW3jYz7Zg5xcu3zIjOl4HLRZMToGWE/view?usp=sharing)
[Stop Icon](https://drive.google.com/file/d/1-9UOS6Nzso3n3h6hzkKk1mUHRiWGmd8k/view?usp=sharing)

interim17:
Time estimate or Size
=======
_Review time: medium_

Problem
=======
Closes #327 

Solution
========
`RecordMovieComponent` manages the modal and button, retrieves and formats the various display icons and text, defines the downloading function, and calls the viewer handlers it receives.

Modal is trim, no logic.

Minimized storing state in `ViewerPanel`: only holding the url for the blob.

Steps to Verify:
----------------
On supported browsers:
1. Load a trajectory, click record, confirm that recording indicator shows up next to button.
2. During playback the record button should pulse red unless hovering, when it should show a red square.
3. Click stop recording, and then confirm download via the modal.
4. Check that the downloaded movie is playable.

On unsupported browsers  (Firefox, and apparently sometimes Safari):
1. Confirm that the button has disabled styling and does nothing when clicked.

Screenshots (optional):
-----------------------
Screenshot 2024-03-26 at 10 41 52 AM
Screenshot 2024-03-26 at 10 41 46 AM
Screenshot 2024-03-26 at 10 41 40 AM
Screenshot 2024-03-26 at 10 41 34 AM


src/components/ViewportButton/style.css
```diff
@@ -36,3 +37,14 @@
...
+.viewport-button:global(.ant-btn) :global(.record-icon-circle),
+.viewport-button:global(.ant-btn) :global(.stop-record-icon) {
+    color: var(--dark-theme-record-button-red);
+}
+
+.viewport-button:global(.ant-btn).disabled :global(.record-icon-circle) {
+    color: var(--dark-theme-viewport-button-disabled-color);
+}
+
+
```
interim17, line 50: These are not general to any other viewport buttons but need this disabled state so they are here, see note above for my questions.

src/components/RecordMoviesComponent/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const startRecordIcon = (
+        <span className={styles.iconContainer}>
+            <span
+                className={classNames("icon-moon", "record-icon-circle")}
+            ></span>
+            <span
+                className={classNames("icon-moon", "record-icon-ring")}
+            ></span>
+        </span>
+    );
```
interim17, line 70: To do multicolor icons via icomoon requires stacking glyphs. This was the best I could come up with, but I'm open to feedback if there is a better way!

Debated defining this in `Icons.tsx` with inline styles and importing from there, but that file typically doesn't handle any thing related to icomoon icons... Style rules for this icon also got a bit spread out, some in primary stylesheet, some here, and some in viewport button. Not ideal? Could consolidate somewhat by most of it to ViewportButton stylesheet but it seems a bit case-specific for that, so I only put the things that relied on disabled state there. This works, but I'm definitely down to refactor/move things around.

src/components/RecordMoviesComponent/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                const secondsStr = secs.toString().padStart(2, "0");
+
+                setFormattedDuration(`${minutesStr}:${secondsStr}`);
+                video.src = "";
+                video.load();
+            };
+        }, [movieUrl]);
+
+        return formattedDuration;
+    };
```
interim17, line 114: `onloadedmetadata` is async hence the need for a hook, Promise or callback. Code was less complicated if this is handled up in `ViewerPanel` and store in state there but that file is already very cluttered with many concerns...

src/components/RecordMoviesComponent/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <div className={classNames([styles.redCircle, styles.animate])}></div>
+    );
+
+    const stopRecordingIcon = isHovering
+        ? "stop-record-icon"
+        : activeRecordingIcon;
+
+    const getIcon = () => {
+        return isRecording ? stopRecordingIcon : startRecordingIcon;
+    };
```
meganrm, line 85: nit: but these two things seem really similar, but one is just a constant and one is a function? unless there is a reason I would pick one setup and go with it for both

src/components/RecordMoviesComponent/index.tsx
meganrm: could you just move this out into its own file?

src/containers/ViewerPanel/index.tsx
```diff
@@ -99,6 +103,7 @@ interface ViewerPanelState {
     particleTypeIds: string[];
     height: number;
     width: number;
+    movieURL: string;
```
meganrm, line 106: since this is stored in local state and only sent down to one component, why not just have it stored by that one component?

src/containers/ViewerPanel/index.tsx
```diff
@@ -364,6 +370,33 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+    };
+
+    public onRecordedMovie = (videoBlob: Blob) => {
+        const url = URL.createObjectURL(videoBlob);
+        this.setState({
+            movieURL: url,
+        });
+    };
+
+    private getMovieTitle = () => {
```
meganrm, line 389: this seems like it could be a selector?

src/components/ViewportButton/index.tsx
```diff
@@ -53,6 +55,11 @@ const ViewportButton: React.FC<ViewportButtonProps> = ({
         }
     };
 
+    const getTooltip = () =>
+        !disabled || tooltipWhenDisabled ? tooltipText : "";
+
+    const getClickHandler = () => (!disabled ? clickHandler : undefined);
```
meganrm, line 61: I'm surprised typescript isn't upset about having `onClick={undefined}`. 
Just a suggestion: 
```suggestion
    onClick = () => {
       if (disabled) {return} else {clickHander()}
    }
    
    onClick={onClick}
    
```
the only difference being that it's now a function that returns nothing rather than setting it to undefined. but it's fine either way.

src/containers/ViewerPanel/index.tsx
interim17: The `onRecordedMovie` handler receives the blob of data from the viewer, so we need to do something with that blob in the component where the callback is defined and passed to the viewer.

We could pass the blob to the lower component as a prop and define the URL there, but I figured that passing the blob around, rather than just a URL, could potentially have performance implications?

meganrm: oh never mind, I see, you needed it for the onRecordMovie prop

src/containers/ViewerPanel/index.tsx
interim17: Yeah, good call

meganrm (approved):
you have a linting error with a now un-needed import, otherwise looks good

src/components/ViewportButton/index.tsx
frasercl: Returning `undefined` seems fine to me, but I might consider making these just plain constants.

frasercl (approved):
Looks good!

PR #484, opened 2024-03-27
Fix/header buttons

interim17:
Time estimate or Size
=======
_Review time: small/medium. Bigish diff, but lots of deleted code and 1:1 swaps of buttons for new custom components._

_Reading `NavButton` and `NavButtonWIthTooltip` and their stylesheet will tell you most of what you need to know._

Problem
=======
Buttons in the header are not styled correctly! They also aren't tab navigable, and could be more DRY.

Actions buttons need disabled state styling, primary and secondary need hover states, "Simularium Home" needs action button styling, and all buttons need consistent focus/hover behavior.

App header needed spacing tweaks on elements. The media queries are out of date with current design, responsive design update out of scope.

[Figma SSOT](https://www.figma.com/file/u4lZYuVwbsFpzBikgwAVRb/Master-components-Simularium?node-id=0%3A1&mode=design)

Solution
========
Solved: styling.

Improved but not solved: accessibility, DRY concerns.

### `NavButton` and `NavButtonWithTooltip`
All buttons in the header are now one of these.

`NavButton` is primarily a theming wrapper, styles are now correct for different button types (primary, seconday, action), and for hover/focus states which makes tab navigation visually apparent. Next steps for that: make the dropdowns work; this refactor takes us a step in the right direction. In looking at accessible drop-downs I think implementing this component will be useful in the dropdown menu items

`NavButtonWIthTooltip` takes care of tooltip logic.

I could make these all one component, but I think it's nice to have it separated so the props are clear, and there isn't conditional rendering of tooltip. It would be easy to consolidate if reviewers have feedback preferring consolidation.

### Miscellany:

Tooltip offsets: using the disabled prop on the buttons added a span that made it necessary to fuss with the tooltip offsets. I think it's easier to just disable the buttons by disabling pointer events with CSS and not use the antd prop at all. Then we can use one offset constant all the time.

The "Simularium Home" link was a react router `Link` but UX wants it styled identical to an action button, and with the same tab navigation behavior. So I did a quick shift of the `AppHeader` from a class based component to functional so that I could make use of `useHistory`. `AppHeader` had no lifecycle methods, the refactor is trivial.

Share and Download components no longer need classNames or stylesheets.

In Download component I refactored a bit to remove arguments and what I thought were redundant checks that are handled by `isDisabled` constant.

Steps to Verify:
----------------
1. Try all the buttons see if they work.
2. Share and Download should be disabled (and look like it) on home page.

Screenshots (optional):
-----------------------
Screenshot 2024-03-26 at 5 02 56 PM
Screenshot 2024-03-26 at 5 00 51 PM


src/containers/AppHeader/index.tsx
```diff
@@ -39,74 +40,76 @@ interface AppHeaderProps {
...
+    simulariumFile,
+    isBuffering,
+    changeToLocalSimulariumFile: loadLocalFile,
+    changeToNetworkedFile: loadNetworkFile,
+    setViewerStatus,
+    clearSimulariumFile,
+    setError,
+    isNetworkedFile,
+}) => {
+    const history = useHistory();
```
interim17, line 53: Don't believe the diff, this functional component is practically identical to the previous class based component. except for changing the `Link` to a `NavButton` that implements `useHistory` and removing the action button container.

src/components/DownloadTrajectoryMenu/index.tsx
```diff
@@ -24,12 +17,10 @@ const DownloadTrajectoryMenu = ({
     simulariumFile,
     isNetworkedFile,
 }: DownloadTrajectoryMenuProps): JSX.Element => {
-    const fileIsLoaded = () => !!simulariumFile.name;
+    const fileName = simulariumFile.name;
+    const isDisabled = !fileName || isBuffering;
```
interim17, line 21: `fileIsLoaded` just did a boolean conversion on the name, and then it was required as an argument in `downloadFile` might as well just save it once and it serves both purposes

src/components/NavButtonWithTooltip/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+} from "../../constants";
+import NavButton, { NavButtonProps } from "../NavButton";
+
+interface TooltipText {
+    default: string;
+    disabled?: string;
+}
+interface NavButtonWithTooltipProps extends NavButtonProps {
+    tooltipText?: TooltipText;
+    tooltipPlacement?: TooltipProps["placement"];
```
toloudis, line 17: how does this work? It looks clunky to use the brackets and string literal.  Does ts let you do something like `typeof TooltipProps.placement` or can you specify it more explicitly?

src/constants/index.ts
```diff
@@ -8,7 +8,7 @@ export const CHECKBOX_TYPE_STAR = "star";
 export type CHECKBOX_TYPE_STAR = typeof CHECKBOX_TYPE_STAR;
 export const TOOLTIP_COLOR = "#3B3649";
 export const TOOLTIP_DELAY = 1; // in seconds
-export const NAV_BAR_TOOLTIP_OFFSET = [0, -8];
+export const NAV_BAR_TOOLTIP_OFFSET = [0, 10];
```
toloudis, line 11: this seems like a big difference.. but I went back and read your description of why.

src/components/DownloadTrajectoryMenu/style.css
toloudis, line 1: this says "empty file"... can delete?

toloudis (approved):
A couple questions but not show stoppers for me.

src/components/NavButton/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    icon,
+    clickHandler,
+    isDisabled,
+    ...props
+}) => {
+    const buttonClassNames = classNames(
+        className,
+        styles.navButton,
+        buttonType && styles[buttonType],
+        isDisabled && styles.disabled
```
meganrm, line 28: just FYI, you could also write this as `{[styles.disabled]: isDisabled}`

src/components/NavButton/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    buttonType,
+    icon,
+    clickHandler,
+    isDisabled,
+    ...props
+}) => {
+    const buttonClassNames = classNames(
+        className,
+        styles.navButton,
+        buttonType && styles[buttonType],
```
meganrm, line 27: is there a default button type? you could set that so you don't need to have a conditional here

src/components/NavButton/style.css
```diff
@@ -0,0 +1,62 @@
...
+
+.nav-button.secondary:global(.ant-btn) {
+    background-color: var(--dark-theme-secondary-button-bg);
+    border-color: var(--dark-theme-secondary-button-border);
+    border-radius: 3px;
+    color: var(--dark-theme-secondary-button-color);
+}
+
+.nav-button.secondary:global(.ant-btn:hover):not(.disabled),
+.nav-button.secondary:global(.ant-btn:focus) {
```
meganrm, line 51: why don't you have the `:not(.disabled)` for the focus state as well?

src/components/NavButtonWithTooltip/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    titleText,
+    tooltipPlacement,
+    buttonType,
+    icon,
+    clickHandler,
+    isDisabled = false,
+    ...props
+}) => {
+    const getTooltipRenderText = () => {
+        return isDisabled && tooltipText.disabled
```
meganrm, line 31: nit: you could also have a default disabled tooltipText like "disabled" that would avoid the need for this double check. In general I would assume that if `isDisabled` is true it all has `tooltipText.disabled`

src/components/NavButtonWithTooltip/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const [tooltipRenderText, setTooltipRenderText] =
+        useState(getTooltipRenderText);
+    const [tooltipVisible, setTooltipVisible] = useState(false);
+
+    // The conditional below prevents a flicker when tooltip text changes faster than tooltip can hide
+    useEffect(() => {
+        if (!isDisabled || (!tooltipVisible && isDisabled)) {
+            setTooltipRenderText(getTooltipRenderText());
+        }
```
meganrm, line 44: this is personal preference, but as this is written now, I need to read that comment in order to parse what this conditional is doing. It also seems like you're checking `!isDiasbled/isDisabled` both here and in get the render text, and I think it's easier to read to just show in this function what you do when isDisabled is changed 

```suggestion
       if (tooltipVisible) {
           return
       }
        if (isDisabled)) {
            setTooltipRenderText(tooltipText.disabled || "");
        } else {
            setTooltipRenderText(tooltipText.default);
        }
```

meganrm: Note, I wouldn't commit this directly, I wrote it in the GitHub editor so it might break things

src/components/NavButtonWithTooltip/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    default: string;
+    disabled?: string;
+}
+interface NavButtonWithTooltipProps extends NavButtonProps {
+    tooltipText?: TooltipText;
+    tooltipPlacement?: TooltipProps["placement"];
+}
+
+const NavButtonWithTooltip: React.FC<NavButtonWithTooltipProps> = ({
+    tooltipText = { default: "" },
```
meganrm, line 21:
```suggestion
    tooltipText = { default: "", disabled: "" },
```

meganrm (commented):
Overall this looks really great, and I think it's a good improvement on drying out the code

src/components/NavButtonWithTooltip/index.tsx
interim17: Hmm, that `typeof` version doesn't work directly.

Antd actually exports `TooltipPlacement` as a type but not from their top level I have to import from `antd/lib/tooltip` which is maybe not the most maintainable?

Alternately can just redeclare it with all the string literals in our own types file and import it, that might be the best way imo

src/components/DownloadTrajectoryMenu/style.css
interim17: oops yes, sorry thought that was deleted

src/components/NavButton/index.tsx
interim17: Great idea, there is now!

src/components/NavButton/style.css
interim17: Hmm, may have been a holdover from a branch where I was working on the tab navigation, I think I can actually remove `:not(.disabled)` from all these selectors now

src/components/NavButtonWithTooltip/index.tsx
interim17: Oh this is great, with this and checking `if (tooltipVisible)` like you suggested we can actually get rid of most of the conditionals in both the `useEffect` and `getRenderTooltipText`

src/components/NavButtonWithTooltip/index.tsx
interim17: Only needed to fix one bracket, but with the default you suggested below we can actually get rid of most of this

meganrm (approved):

PR #486, opened 2024-03-28
autoconversion: Fix/conversion filename bug

interim17:
Time estimate or Size
=======
_Medium_

This should be working with current viewer patch.

Problem
=======
Converted trajectories were not downloadable, had no titles in the header, and were buggy when loaded after other trajectories.

Solution
========
The biggest part of this change is a new flow to receive converted files, first `handleIncomingConvertedFile` in the viewer panel and then `receiveConvertedFileLogic` in redux.

`handleIncomingConvertedFile` puts the name and title into state for use in the header, download menu, etc.

The `receiveConvertedFileLogic` calls `controller.changeFile` with the right arguments to keep the websocket connection we established during conversion, switching connections was causing bugs.

Some of this was fixed by recent viewer PR, and this change set brings the website into alignment with viewer in how it calls controller methods like `convertTrajectory` and `changeFile`.

Miscellaneous changes:
- `getIsNetworkedFile` selector was checking if files are in `TRAJECTORIES` constant, but converted files won't always be there in auto-conversion era, removed check
- in octopus era all downloads will point at new HTTP endpoint instead of s3, updated constant and added todo for launch, and adjusted download string to match in `getHref`

To review:
- convert a smoldyn file
- when it arrives, check for title in status bar
- download it with download button
- load the downloaded trajectory
- load a different trajectory and try the flow again


meganrm:
can you convert this to a draft until the tests have been fixed?

src/containers/ConversionForm/index.tsx
```diff
@@ -109,6 +113,8 @@ const ConversionForm = ({
     const cancelProcessing = () => {
         setIsProcessing(false);
         setConversionStatus({ status: CONVERSION_NO_SERVER });
+        // todo - keep old trajectory and timestamp when cancelling conversion request
+        clearSimulariumFile({ newFile: false });
```
interim17, line 117: There are some issues currently in Octopus with returning to the previous trajectory once conversion has started, and Blair has requested that for now, once a conversion request is initiated, we clear out the viewer.

src/containers/ViewerPanel/index.tsx
```diff
@@ -261,13 +264,25 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+        const { receiveConvertedFile, conversionProcessingData } = this.props;
+        const fileId = conversionProcessingData.fileId + ".simularium";
+        const hasTitle =
+            data.trajectoryTitle !== undefined &&
+            data.trajectoryTitle.length > 0;
+        const title = hasTitle ? data.trajectoryTitle : fileId;
+        receiveConvertedFile({
+            name: fileId,
+            title: title,
+        });
```
interim17, line 277: This incoming file _should_ have a title, but it will definitely have a fileId, so we can default to that. Having a title is how we have been checking that a file is networked and not local.

src/state/trajectory/logics.ts
```diff
@@ -575,6 +578,35 @@ const convertFileLogic = createLogic({
...
+        const currentState = getState();
+        const conversionStatus = getConversionStatus(currentState);
+        const simulariumController = getSimulariumController(currentState);
+        const simulariumFile = action.payload;
+
+        simulariumController.changeFile(
+            netConnectionSettings,
+            simulariumFile.name,
+            true
+        );
```
interim17, line 593: calling changeFile with this boolean flag ensure we use the same websocket connection from the conversion process

src/state/trajectory/logics.ts
```diff
@@ -575,6 +578,35 @@ const convertFileLogic = createLogic({
...
+        history.replaceState(
+            {},
+            "",
+            `${location.origin}${location.pathname}?${URL_PARAM_KEY_FILE_NAME}=${simulariumFile.name}`
+        );
+        if (conversionStatus !== CONVERSION_INACTIVE) {
+            dispatch(setConversionStatus({ status: CONVERSION_INACTIVE }));
+        }
+        simulariumController.gotoTime(0);
+        done();
```
interim17, line 605: goToTime is no longer called during the controller convertTrajectory method (formerly convertAndLoadTrajectory)

src/state/trajectory/selectors/index.ts
```diff
@@ -17,10 +14,7 @@ export const getIsNetworkedFile = createSelector(
         if (!simFile.name) {
             return false;
         }
-        return (
-            !!find(TRAJECTORIES, { id: simFile.name }) &&
-            isNetworkSimFileInterface(simFile)
-        );
+        return isNetworkSimFileInterface(simFile);
```
interim17, line 17: a converted file is a networked trajectory we can no longer count on having all networked files in in our list of trajectories

src/containers/ViewerPanel/index.tsx
```diff
@@ -261,13 +264,25 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         setIsLooping(!isLooping);
     }
 
+    public handleIncomingConvertedFile(data: TrajectoryFileInfo) {
+        const { receiveConvertedFile, conversionProcessingData } = this.props;
+        const fileId = conversionProcessingData.fileId + ".simularium";
```
meganrm, line 269: nit: 
```suggestion
        const { fileId } = conversionProcessingData;
```

src/containers/ViewerPanel/index.tsx
```diff
@@ -261,13 +264,25 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+        const hasTitle =
+            data.trajectoryTitle !== undefined &&
+            data.trajectoryTitle.length > 0;
+        const title = hasTitle ? data.trajectoryTitle : fileId;
```
meganrm, line 273:
```suggestion
        const title = hasTitle ? data.trajectoryTitle : `${fileId}.simularium`;
```

src/containers/ViewerPanel/index.tsx
meganrm: what are these two things used for? and how are they different? just curious why they're set to different things (although they may be both fieldId in some cases it seems

src/state/trajectory/constants.ts
```diff
@@ -35,6 +35,8 @@ export const RECEIVE_FILE_TO_CONVERT = makeTrajectoryConstant(
     "receive-file-to-convert"
 );
 export const CONVERT_FILE = makeTrajectoryConstant("convert-file");
+export const RECEIVE_CONVERTED_FILE =
+    makeTrajectoryConstant("receive-converted");
 
 export const SET_CONVERSION_STATUS = makeTrajectoryConstant("set-status");
 export const CONVERSION_INACTIVE = "conversion-inactive";
```
meganrm, line 42: why isn't this using the makeTrajectoryConstant function?

src/state/trajectory/logics.ts
```diff
@@ -546,26 +547,28 @@ const convertFileLogic = createLogic({
...
 
         const { engineType, fileToConvert, fileName } =
             getConversionProcessingData(getState());
         const fileContents: Record<string, any> = {
             fileContents: { fileContents: fileToConvert },
-            metaData: { trajectoryTitle: fileName },
+            trajectoryTitle: fileName,
         };
         const controller = getSimulariumController(getState());
+        const providedFileName = action.payload + ".simularium";
```
meganrm, line 559: nit: 
```suggestion
        const providedFileName = `${action.payload}.simularium`;
```

src/state/trajectory/logics.ts
meganrm: is changeFile an async function? how do you know it was successful/catch errors if it failed?

src/containers/ViewerPanel/index.tsx
```diff
@@ -261,13 +264,25 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+        });
+    }
+
     public onTrajectoryFileInfoChanged(data: TrajectoryFileInfo) {
-        const { conversionStatus, setConversionStatus } = this.props;
-        if (conversionStatus !== CONVERSION_INACTIVE) {
-            setConversionStatus({ status: CONVERSION_INACTIVE });
+        const { receiveTrajectory, simulariumController, conversionStatus } =
+            this.props;
+        if (conversionStatus === CONVERSION_ACTIVE) {
```
meganrm, line 283: I have a memory of these status's being an enum? Like the way we have for the `VIEWER_STATUS`?

src/containers/ViewerPanel/index.tsx
meganrm, line 273: are there cases where 
```suggestion
        const title = data.trajectoryTitle  || fileId;
```
didn't work? just curious about your check for both undefined and the length

src/containers/ViewerPanel/index.tsx
interim17: That code hasn't been put in PR yet for either viewer or conversion status, I have both changes in the same dev branch.

src/state/trajectory/constants.ts
interim17: I don't know! It should be.

interim17: Oh I remember, these are n't action types they aren't specificlly used in redux, they're just used in conditional checks in the application related to conversion. I don't know that they should use `makeTrajectoryConstant` but maybe they should be declared in a different file?

src/containers/ViewerPanel/index.tsx
interim17: made this change

src/state/trajectory/constants.ts
interim17: Either way they are going to be changed when we add the enum in the forthcoming PR...

src/state/trajectory/logics.ts
interim17: yes `changeFile` returns a Promise

src/containers/ViewerPanel/index.tsx
interim17: The existing check for whether a file is networked and not a local file is that it has a "title" stored in redux.

The trajectory that comes back from the server after conversion has both a fileName and title, the fileName is typically the uuid provided during conversion and the title is (currently) derived from the file name that the user uploaded. In future we will let them choose their preferred title.

I put the fileId as a backup/failsafe for the title to ensure that the trajectory is recognized as a networked file and not a local file if for some reason it comes back from the server with that field empty, but it shouldn't!

src/state/trajectory/logics.ts
meganrm: then you should either await it's return or have a `then` block that encompasses everything you want to do if it's successful. and have a fail state if it doesn't work

src/constants/index.ts
```diff
@@ -15,8 +15,9 @@ export const URL_PARAM_BASE_TYPES = "base_types.json";
 export const URL_PARAM_CUSTOM_TYPES = "custom-types";
 // URLs
 export const BASE_API_URL = `/api/${API_VERSION}`;
+// todo octopus launch: this is pointing to the staging octopus http endpoint
 export const DATA_BUCKET_URL =
-    "https://aics-simularium-data.s3.us-east-2.amazonaws.com";
+    "https://staging-simularium-ecs.allencell.org:443/download";
```
ascibisz, line 20: Since we're handling download requests through octopus now, we should just use the `BACKEND_SERVER_IP`, like we do for defining the net connection settings, and remove this constant. Hard coding it to look at staging like we have here would break in production, because there won't be the same files in each for the case of autoconversion

interim17: Ok, fixed this! We talked abou this, but I'm posting here for context for other reviewers that the `.simularium` file extension needed to be appended to the `fileId` earlier in the process to ensure that the download links for converted and other networked files also looked the same.

ascibisz (approved):
I may not be a redux expert, but this looks good to me!

src/constants/index.ts
```diff
@@ -15,8 +15,7 @@ export const URL_PARAM_BASE_TYPES = "base_types.json";
 export const URL_PARAM_CUSTOM_TYPES = "custom-types";
 // URLs
 export const BASE_API_URL = `/api/${API_VERSION}`;
-export const DATA_BUCKET_URL =
-    "https://aics-simularium-data.s3.us-east-2.amazonaws.com";
+export const DATA_BUCKET_URL = `https://${process.env.BACKEND_SERVER_IP}:443/download`;
```
ascibisz, line 18: nit: this isn't really a bucket, `DOWNLOAD_URL` may be more accurate

meganrm (approved):

PR #487, opened 2024-03-28
autoconversion: fix/domnesting error

interim17:
Time estimate or Size
=======
_xsmall_

Problem
=======
Getting a validate dom nesting error because list elements are inside a `p` tag.
Also the text is not true, MVP will only convert Smoldyn so changing to reflect that and leaving UX's intended text in to restore as we add other converters.

Solution
========
`p` -> `div`, adjusted styles

### DIFF
--- a/src/components/NoTrajectoriesText/index.tsx
+++ b/src/components/NoTrajectoriesText/index.tsx
@@ -57,10 +57,7 @@ const NoTrajectoriesText = ({
                             <li>SpringSaLaD</li>
                         </ul>
                     </li> */}
-                    <li>
-                        {" "}
-                        Import a Smoldyn file through the Load Model menu.{" "}
-                    </li>
+                    <li>Import a Smoldyn file through the Load Model menu.</li>
                     <li>
                         Convert other file types using{" "}
                         <a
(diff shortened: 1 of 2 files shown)


src/components/NoTrajectoriesText/index.tsx
```diff
@@ -28,7 +28,7 @@ const NoTrajectoriesText = ({
     return (
         <div className={styles.container}>
             <h3>Ways to get started:</h3>
-            <p>
+            <div>
                 <ol>
```
meganrm, line 32: another similar change: this isn't an ordered list

src/components/NoTrajectoriesText/style.css
```diff
@@ -2,20 +2,17 @@
     margin: 1em;
 }
 
-.container p, .container h3 {
+.container div, .container h3 {
```
meganrm, line 5: I would give this a classname instead of targeting such a general tag type. this definitely opens you up to unintended application of style

src/components/NoTrajectoriesText/index.tsx
interim17: well actually i think the design does show these bullet points as numbered, i guess that's why we had an `ol`?

src/components/NoTrajectoriesText/style.css
interim17: k i'll make a couple changes the component styling  make it more class focused

src/components/NoTrajectoriesText/index.tsx
meganrm: oh, you're right, nm

meganrm (approved):

frasercl (approved):
LGTM!

PR #488, opened 2024-04-04
autoconversion: fix/refactor after merging main

interim17:
Time estimate or Size
=======
_Medium/Small: many changes are obvious, there's essentially no logic_

Problem
=======
I merged main into 346 and now the styling is a bit whacked up, because there have been recent changes to color vars, custom modal props, etc. This deals with merge issues and refactors conversion branch to work with current styling approaches.

_Note: there is a bug/issue where the converted file is not recognized upon load and so it does not display a title and the Download and Share buttons remain disabled despite a file being loaded. The changes for that bug are in process pending discussion of changes to Octopus and TrajectoryInfo typing. I think its acceptable to proceed with this PR (and merge to main) with that resolution pending, as this will not go to production until Octopus is live._

Solution
========
This PR fixes the handful of erroneous merge artifacts, and brings the `ConversionForm`, `ConversionOverlay`, and various conversion modals into alignment with the recent changes to `CustomModal`, color variable naming, etc. etc.

Most of the changes to the modals are just using the new props for `CustomModal` and a couple style tweaks.

To review styling on on modals and processing overlay, outside of just looking at screenshots I suggest just pasting the following code blocks into the `ConversionForm` render function outside of the usual conditional rendering:
```
             <ConversionServerErrorModal
                    closeModal={toggleServerCheckModal}
                />
                
                                <ConversionFileErrorModal
                    closeModal={toggleFileTypeModal}
                    engineType={conversionProcessingData.engineType}
                />
                
                                <ConversionProcessingOverlay
                    fileName={conversionProcessingData.fileName}
                    cancelProcessing={cancelProcessing}
                />
```
You can upload an invalid file type to test the file type error, but triggering the server error and getting it to hang out on the processing overlay is a little more fussy.

Steps to Verify:
----------------
1. Convert a file and confirm everything still works after merging main.
2. Check out the conversion form, overlay and modals and see that they look like the current light-theme guidance.

Screenshots (optional):
Screenshot 2024-04-03 at 5 05 18 PM
Screenshot 2024-04-03 at 4 53 59 PM

-----------------------
Screenshot 2024-04-03 at 5 05 44 PM

Screenshot 2024-04-03 at 5 05 58 PM

### DIFF
--- a/src/components/ConversionServerErrorModal/index.tsx
+++ b/src/components/ConversionServerErrorModal/index.tsx
@@ -27,7 +27,7 @@ const ConversionServerErrorModal: React.FC<ConversionServerErrorModalProps> = ({
             closeHandler={closeModal}
         >
             <p className={styles.warningText}>
-                {Exclamation}{" "}
+                {Exclamation}
                 {` We're sorry, the server is currently
                 experiencing an issue.`}
             </p>
--- a/src/components/ConversionServerErrorModal/index.tsx
+++ b/src/components/ConversionServerErrorModal/index.tsx
@@ -1,5 +1,5 @@
-import { Button } from "antd";
 import React from "react";
+import { Button } from "antd";
 
 import CustomModal from "../CustomModal";
 import { Exclamation } from "../Icons";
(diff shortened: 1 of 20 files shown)


src/components/ConversionServerErrorModal/index.tsx
```diff
@@ -13,26 +13,27 @@ interface ConversionServerErrorModalProps {
...
-            onCancel={closeModal}
+            titleText="File import cannot be completed"
+            footerButtons={footerButtons}
+            closeHandler={closeModal}
         >
             <p className={styles.warningText}>
-                {Exclamation} We&apos;re sorry, the server is currently
-                experiencing an issue.
+                {Exclamation}{" "}
+                {` We're sorry, the server is currently
```
meganrm, line 31: nit, but it seems like you have double spaces here? I would use just one method for creating the space

src/containers/ConversionForm/index.tsx
```diff
@@ -237,9 +237,11 @@ const ConversionForm = ({
                     )}
                 </div>
                 <Divider className={styles.divider} orientation="right" />
-                <Button ghost onClick={cancelConversion}>Cancel</Button>
+                <Button className="secondary-button" onClick={cancelConversion}>
+                    Cancel
+                </Button>
                 <Button
-                    type="primary"
+                    className="primary-button"
```
meganrm, line 244: this seems like we should have a constant for it, I'm surprised we don't already. I wouldn't do it in this PR since this PR is just trying to get it to be the same as main, but maybe make an issue for it

meganrm (approved):

PR #492, opened 2024-04-10
Feature/unsupported engines

interim17:
Time estimate or Size
=======
_Small_

Problem
=======
1) Autoconversion only supports Smoldyn for now, so we need to comment out the other engines, and not show them in the dropdown menu.
2) A few more little tweaks to bring us into tighter alignment with `main` pre-merge... I found these when I realized we would not be able to directly merge `346` due to an extremely stale diff that showed an archaic version of main.

The branch we're going to be merging is called `feature/autoconversion` and this PR should make that branch and `346` identical.

Solution
========

Steps to Verify:
----------------
1. Check that only Smoldyn appears in the conversion drop down.
Screenshot 2024-04-09 at 6 36 33 PM


src/components/Icons/index.tsx
```diff
@@ -35,18 +33,16 @@ export const Reset = <HomeOutlined />;
 export const ZoomIn = <PlusOutlined />;
 export const Cancel = <PlusOutlined style={{ transform: "rotate(45deg)" }} />;
 export const ZoomOut = <MinusOutlined />;
-export const Share = <ShareAltOutlined />;
+export const Share = <ShareAltOutlined style={{ fontSize: "15px" }} />;
 export const Warn = <WarningOutlined />;
 export const Link = <LinkOutlined />;
-export const Download = <DownloadOutlined size={32} />;
+export const Download = <DownloadOutlined style={{ fontSize: "15px" }} />;
```
interim17, line 39: changed to match main

meganrm (approved):

frasercl (approved):

PR #500, opened 2024-04-23
point at octopus cluster

interim17:
Time estimate or Size
=======
_xsmall_

Problem
=======
This is the exact same code we merged into the `346` autoconversion branch last week. Following up from dev meeting today we want to get main pointed at Octopus so we can debug/test with staging.


ShrimpCryptid (approved):

meganrm:
I think it would be good to merge and release the movie record PR so that any issues with octopus don't block that feature

frasercl (approved):

PR #501, opened 2024-04-23
Feature/button constant

### LINKED ISSUE #489: use a constant to tell buttons apart
## Use Case

Buttons are currently taking string props like "primary-button" and "action-button" to determine their styling. We should use a constant to enforce this instead, this design pattern any other similar use cases

interim17:
Time estimate or Size
=======
_medium_

Problem
=======
Closes #489 

Solution
========
#### `ButtonClass`
Rather than passing in a string like "primary-button" that will be handled by something like `ButtonClass.Primary`. 

Bringing `NavButton` into this new system meant tweaking the css selectors to say "primary-button" instead of just "primary" but styling is not actually changed.

#### `IconGlyphs`
While I was at it I made the enum `IconGlyphs` so that using icomoon glyphs doesnt require manually applying several strings in a class name, seemed like a related enough change for this PR.

Steps to Verify:
----------------
1. Everything should look and act the same as before.


src/constants/interfaces.ts
```diff
@@ -38,3 +38,25 @@ export type TooltipPlacement =
     | "leftBottom"
     | "rightTop"
     | "rightBottom";
+
+export enum ButtonClass {
+    Primary = "primary-button",
+    Secondary = "secondary-button",
+    Action = "action-button",
+}
+
```
meganrm, line 47:
```suggestion
create iconGlyphClass = (name) => return `icon-moon anticon ${name}`
```

interim17: I made that function but moved it to utils and typed the argument as `IconGlyphs` so we could keep the type safety and not use strings directly, does that capture what you were thinking?

src/components/StarCheckbox/index.tsx
```diff
@@ -19,13 +21,16 @@ const StarCheckbox = ({
 
     const parentClassnames = className ? className.split(" ") : [];
     const wrapperClassnames = classNames([...parentClassnames, styles.wrapper]);
-    const checkboxClassNames = classNames(
-        ["icon-moon", "star-empty-icon", styles.checkbox],
-        {
-            ["star-full-icon"]: checked,
-            ["star-dashed-icon"]: indeterminate,
+    const getIcon = () => {
```
meganrm, line 24: nit: `getIconClasses`

meganrm (approved):

src/components/ViewportButton/index.tsx
```diff
@@ -9,7 +9,7 @@ interface ViewportButtonProps extends ButtonProps {
     tooltipText?: string;
     tooltipPlacement?: TooltipProps["placement"];
     tooltipWhenDisabled?: boolean;
-    icon?: ReactNode | string; // When using an icomoon icon, pass the icon name as defined in selectors in src/styles.css
+    icon?: ReactNode | string; // When using an icomoon icon, pass the right IconGlyphs member as defined in src/constants/interfaces
```
toloudis, line 12: Should this comment also mention the wrapper function that adds the icomoon class names?

toloudis (approved):
👍

PR #518, opened 2024-05-16
Feature/max file size

### LINKED ISSUE #516: Autoconversion Max File Size Messaging
## Use Case
Octopus doesn't accept messages larger than 200 MB. If a user tries to upload a file over 200 MB via the autoconversion form, there should be an error message pop up indicating that we only handle files <= 200 MB

## Details
This should match with the other error pop ups, like what shows up if you try to use a file with the wrong extension

interim17:
Time estimate or Size
=======
_small_

Problem
=======
Closes #516 

Solution
========
Add a check for file size in `handleFileSelection`

This also would have been the fourth conditional render at the top of the JSX, which was looking messy to my eye,.

I made an enum for error types, and consolidated all the modal open/closed states into a single typed value `conversionError` and a single function can now close all the modals, I think it's a lot cleaner this way.


src/containers/ConversionForm/index.tsx
```diff
@@ -167,27 +175,41 @@ const ConversionForm = ({
...
+            case ConversionError.FILE_TYPE_ERROR:
+                return (
+                    <ConversionFileErrorModal
+                        closeModal={closeErrorModal}
+                        engineType={conversionProcessingData.engineType}
+                    />
+                );
+            case ConversionError.FILE_SIZE_ERROR:
+                return (
+                    <ConversionFileSizeErrorModal
```
meganrm, line 193: are these modals all super different from each other? it seems like you could just send in a message instead of making a whole new component for each one

interim17: Not super different no, could definitely consolidate to one component. The original designs showed some slight differences that might not be relevant since the Figma SSOT came out, and there are couple prop differences, nothing major.

meganrm: the one prop difference seems to be the engine type which could just be part of the generic `errorMessage` prop. especially considering they all have identical style sheets I would definitely combine them into one

src/containers/ConversionForm/index.tsx
```diff
@@ -77,11 +77,20 @@ const ConversionForm = ({
...
+        ConversionError.SERVER_ERROR
+    );
 
     const engineSelected = !!conversionProcessingData.engineType;
+    const errorModalOpen = conversionError !== ConversionError.NO_ERROR;
+
+    const errorMessage = {
+        [ConversionError.SERVER_ERROR]: "Please try again at a later time.",
+        [ConversionError.FILE_TYPE_ERROR]: `You may want to double check that the file you selected is a valid ${conversionProcessingData.engineType}
+                        file and try again.`,
```
ShrimpCryptid, line 90: I'm seeing that this got broken up into two lines; will that introduce extra whitespace when this message is printed?

src/containers/ConversionForm/index.tsx
```diff
@@ -77,11 +77,20 @@ const ConversionForm = ({
 }: ConversionProps): JSX.Element => {
     const [fileToConvert, setFileToConvert] = useState<UploadFile | null>();
     const [isProcessing, setIsProcessing] = useState<boolean>(false);
-    const [serverErrorModalOpen, setServerErrorModalOpen] =
-        useState<boolean>(false);
-    const [fileTypeErrorModalOpen, setFileTypeErrorModalOpen] = useState(false);
+    const [conversionError, setConversionError] = useState<ConversionError>(
+        ConversionError.SERVER_ERROR
```
ShrimpCryptid, line 81: Should the default be `ConversionError.NO_ERROR`?

src/containers/ConversionForm/index.tsx
```diff
@@ -125,6 +129,11 @@ const ConversionForm = ({
     };
 
     const handleFileSelection = async (file: UploadFile) => {
+        // 200 MB limit
```
ShrimpCryptid, line 132: Make this a constant?

src/containers/ConversionForm/index.tsx
```diff
@@ -125,6 +129,11 @@ const ConversionForm = ({
...
+        if (file.size !== undefined && file.size > 2e8) {
```
ShrimpCryptid, line 133: Is it possible for the file's size to be undefined, but the file is over 200 MB?

interim17: I don't think so?

src/containers/ConversionForm/index.tsx
interim17: Yes! thank you

src/containers/ConversionForm/index.tsx
interim17: No it looks ok on my machine

src/containers/ConversionForm/index.tsx
meganrm: I'm curious what the case is where there's a file but size is undefined? why can't it just be the size check?

src/containers/ConversionForm/index.tsx
```diff
@@ -78,16 +79,17 @@ const ConversionForm = ({
...
 
     const errorMessage = {
-        [ConversionError.SERVER_ERROR]: "Please try again at a later time.",
+        [ConversionError.SERVER_ERROR]: `We're sorry, the server is currently
+                experiencing an issue. Please try again at a later time. `,
         [ConversionError.FILE_TYPE_ERROR]: `You may want to double check that the file you selected is a valid ${conversionProcessingData.engineType}
-                        file and try again.`,
+                        file and try again. `,
         [ConversionError.FILE_SIZE_ERROR]:
             "Your file exceeds the maximum allowed size of 200 MB, please try uploading a smaller file.",
```
ShrimpCryptid, line 94: Nit: Why do two of these have an extra space after the final punctuation?

ShrimpCryptid (approved):

src/containers/ConversionForm/index.tsx
interim17: size is an optional property on the type `UploadFile` so its possibly undefined

meganrm (approved):

PR #520, opened 2024-05-17
Fix/send cancel autoconversion msg

ascibisz:
Time estimate or Size
=======
small

Problem
=======
Now that we have cancel autoconversion functionality in Octopus, and a cancel autoconversion message that the viewer can send to trigger that, the website should trigger this when a user selects to cancel a conversion request that's processing

[Related to this ticket](https://github.com/simularium/octopus/issues/87)

Solution
========
Added call to `SimulariumController.cancelConversion()` when user has cancelled their conversion request.

Did this using redux (for my very first time) since we will likely need to add more functionality to cancelConversionLogic in the future to properly restore the previous state of the viewer when a user has cancelled a conversion request.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "0.11.0",
             "license": "ISC",
             "dependencies": {
-                "@aics/simularium-viewer": "^3.7.4",
+                "@aics/simularium-viewer": "^3.8.0",
                 "@ant-design/css-animation": "^1.7.3",
                 "@ant-design/icons": "^4.0.6",
                 "antd": "^4.23.6",
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "0.11.0",
             "license": "ISC",
             "dependencies": {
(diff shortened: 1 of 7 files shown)


ShrimpCryptid (approved):
Looks good to me! I still don't fully understand Redux's model so I'll defer to Joe for any other comments.

package.json
```diff
@@ -94,7 +94,7 @@
         "webpack-dev-server": "^4.10.1"
     },
     "dependencies": {
-        "@aics/simularium-viewer": "^3.7.3",
+        "@aics/simularium-viewer": "^3.7.4",
```
meganrm, line 97: should update this again to the latest version to double check

interim17 (approved):
Front end wise this LGTM.

If I open a trajectory, then cancel a conversion during the processing, I can trigger a little front end error working that the server is down, and I see multiple websocket connections in the network tab. So there may be some more fiddling to do on either octopuys or the viewer to polish that, but it's nothing blocking here, and it doesn't affect any thing loading or converting correctly, just throws a little warning modal.

PR #521, opened 2024-05-20
Feature/state enums

### LINKED ISSUE #455: Use enums for viewer and conversion status constants
## What needs to happen?
Convert some types and constants to `enums`

## Why should we do this?

During review on autoconersion, @ShrimpCryptid pointed out that it would be efficient and readable to use an `enum` where we currently use types in some cases, notably viewer `status` and `conversionStatus`. When converting these, we should also check for other opportunities to include legibility and efficiency of typing in the website repo.

Not urgent, appropriate to fold into auto-conversion work

interim17:
Time estimate or Size
=======
_small_

Problem
=======
Closes #455 

Solution
========
We are using the enums `ConversionStatus` and `ViewerStatus` to manage typing a bit better.


toloudis (approved):
Looks good, I like this change! I'm not sure if the automated coverage annotation warnings are something we are concerned about.  If not, we should probably remove them as they add noise,

src/containers/ConversionForm/index.tsx
```diff
@@ -114,11 +109,11 @@ const ConversionForm = ({
 
     const cancelProcessing = () => {
         setIsProcessing(false);
-        setConversionStatus({ status: CONVERSION_NO_SERVER });
+        setConversionStatus({ status: ConversionStatus.Inactive });
```
meganrm, line 112: should be `NoServer`?

meganrm: or was it wrong before?

src/state/trajectory/types.ts
```diff
@@ -117,8 +111,9 @@ export interface HealthCheckTimeout {
...
 
-export type ConversionStatus =
-    | typeof CONVERSION_INACTIVE
-    | typeof CONVERSION_NO_SERVER
-    | typeof CONVERSION_SERVER_LIVE
-    | typeof CONVERSION_ACTIVE;
+export enum ConversionStatus {
+    Inactive = "CONVERSION_INACTIVE",
+    NoServer = "CONVERSION_NO_SERVER",
+    ServerLive = "CONVERSION_SERVER_LIVE",
```
meganrm, line 117: is this one necessary? it doesn't seem mutually exclusive with the others. Presumably the server is live while conversion in all cases other than "no server". I also don't see anything that uses `if CONVERSION_STATUS == ServerLive`

interim17: When we go the conversion form we assume "no server" until the health check comes back, but conversion isn't "active" till we send the file, so "server live" is the in between state that where we are good to proceed. The way things are written now I believe we do need it.

meganrm (approved):
Left a comment about a name change, otherwise looks good

PR #522, opened 2024-05-21
Fix/plot data update

meganrm:
Time estimate or Size
=======
 xsmall

Problem
=======
the typing of the return of plot data was inconsistent and incorrect. has been addressed in[ this PR](https://github.com/simularium/simularium-viewer/pull/386#issuecomment-2123090748) but that is a breaking change that needs to be handled here

Solution
========
update the plot data unpacking. needs to wait until we release a new version of the viewer

## Type of change

Steps to Verify:
----------------
1. install a local version of the viewer from[ this branch ](https://github.com/simularium/simularium-viewer/tree/fix/plotdata)
2. [file to load ](http://localhost:9001/viewer?trajUrl=https://aics-simularium-data.s3.us-east-2.amazonaws.com/trajectory/binding-affinity_antibodies.simularium)
1. the plot should show 
2. you can also load the two test files I just added to verify that they are both loading correctly (the JSON doesn't load plots on the main site now)


.github/workflows/test-coverage.yml
```diff
@@ -12,4 +12,4 @@ jobs:
             - uses: ArtiomTr/jest-coverage-report-action@v2.2.4
               with:
                   github-token: ${{ secrets.GITHUB_TOKEN }}
-                  annotations: coverage
+                  annotations: failed-tests
```
meganrm, line 15: to get rid of the coverage annotations

frasercl (approved):

toloudis (approved):

PR #532, opened 2024-06-12
Feature/information panel

### LINKED ISSUE #411: Implement show info for selected agent
## Use Case
As a user, I want to be able to see information about the current selected agent, so that I can more easily debug or explore the trajectory.

## Acceptance Criteria
https://github.com/simularium/simularium-planning/issues/141

## Design
https://www.figma.com/file/VMyr4KilvLAZF6XqKVv12l/Agent-Info-Design?type=design&node-id=320%3A16323&mode=design&t=eL5zQK1RHNpLEqVE-1

interim17:
Time estimate or Size
=======
_medium: 10-15_

Problem
=======
Closes #411 

Solution
========
New redux actions/type/constant/state for `selectedAgent`

### `ViewerPanel`
Retrieve agent meta data via the new viewer callback `onFollowObjectChanged` and put into redux as `AgentMetadata`.

### `ModelPanel` -> `SidebarContents`
Get metadata from state and pass it (and UIDisplayData) down to new Metadata Panel.
Refactored both of these components from Classes into Functional Components, trivial because no lifecycle methods.
Small tweaks to html structure and CSS

### `MetadataPanel`
Hopefully it reads easily, is fairly self explanatory.
Use helper functions and manage a few pieces of state/constants to track whether the panel is open and what it should display.

`selectedAgentNotRendered` is truthy only when the agent type is a positive number, greater than the number of agents in UIDisplayData. I don't know if there are any other ways to check for an agent that exists, but has "left the simulation" that don't collide with the null state where the id is -1.

Steps to Verify:
----------------
1. Open and close the panel.
2. Select/deselect agents, play the sim and verify that data updates

Screenshots (optional):
-----------------------
Screenshot 2024-06-12 at 4 40 08 PM
Screenshot 2024-06-12 at 4 40 16 PM
Screenshot 2024-06-12 at 4 40 35 PM
Screenshot 2024-06-12 at 4 40 26 PM

Thanks for contributing!


src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+        key: string,
+        value: number | PositionRotation
+    ): JSX.Element | string => {
+        if (selectedAgentNotRendered) {
+            return "-";
+        }
+        if (key === "agentType") {
+            return uiDisplayData[selectedAgent.agentType].title;
+        }
+        if (typeof value === "object") {
```
meganrm, line 65: could this also be `key === [KEY_NAME]`? just to make it more explicit?

meganrm: oh, I see, it's for both rotation and position. I might even switch it to if (key === position or key=== rotation) just to make it clear

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+            Rotation
+            <br />
+            (degrees)
+        </>
+    ),
+    radius: <span>Radius</span>,
+};
+
+const formatFloatForDisplay = (float: number): string => {
+    return parseFloat(float.toPrecision(2)).toString();
```
meganrm, line 26: I think it would be good to put this in utils, and format these numbers when you save the data instead of saving them in state in a format you'll never use

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { Divider } from "antd";
+import classNames from "classnames";
+
+import { AgentMetadata, PositionRotation } from "../../constants/interfaces";
+import { FilledCaret } from "../Icons";
+import { AgentDisplayNode } from "../AgentTree";
+
+import styles from "./style.css";
+
+const MetadataLabels: Record<keyof AgentMetadata, JSX.Element> = {
```
meganrm, line 11: I think it would be good to move this within the functional component, it just confuses me up here

src/components/MetadataPanel/index.tsx
interim17: yeah will do, i tried to do without the key argument but added it back in the last change set, so i'll make this more explicit

interim17: FYI: this change requires keeping `typeof value === "object"` when adding the key checks, so that I can access the properties `value.x` etc, and I need to typecast the argument in the final return as `value as number`

src/components/SideBar/index.tsx
```diff
@@ -64,7 +64,7 @@ export default class SideBar extends React.Component<SiderProps, SiderState> {
                 <div className={triggerClass} onClick={this.handleTriggerClick}>
                     {PurpleArrow}
                 </div>
-                {children}
+                {!this.state.collapsed && children}
```
interim17, line 67: this prevents flickering content when opening and closing the sider

src/components/SideBar/index.tsx
```diff
@@ -43,6 +43,7 @@ export default class SideBar extends React.Component<SiderProps, SiderState> {
             [styles.sider]: true,
             [styles[type]]: true,
             [styles.embed]: isEmbedded,
+            [styles.collapsed]: this.state.collapsed,
```
interim17, line 46: so we can only have the padding when collapsed, and the divider can reach across the length of the panel when open

src/containers/ModelPanel/index.tsx
```diff
@@ -62,71 +64,67 @@ interface ModelPanelProps {
...
+        <div className={styles.container}>
+            <SideBarContents
+                mainTitle="Agents"
+                content={[contentMap[viewerStatus]]}
+                selectedAgent={selectedAgent}
+                uidisplayData={uiDisplayDataTree}
+            />
+        </div>
+    );
+};
```
interim17, line 127: These are the relevant changes here, the rest relates to converting Class to FC

src/constants/interfaces.ts
```diff
@@ -67,3 +67,15 @@ export enum ConversionError {
     FILE_SIZE_ERROR = "fileSizeError",
     NO_ERROR = "noError",
 }
+export interface PositionRotation {
```
toloudis, line 70: I suggest renaming this to Vector3D or XYZNumbers or something like that.

src/containers/ModelPanel/index.tsx
```diff
@@ -62,71 +64,67 @@ interface ModelPanelProps {
...
-        );
-    }
-}
+    return (
+        <div className={styles.container}>
+            <SideBarContents
+                mainTitle="Agents"
+                content={[contentMap[viewerStatus]]}
+                selectedAgent={selectedAgent}
+                uidisplayData={uiDisplayDataTree}
```
toloudis, line 123:
```suggestion
                uiDisplayData={uiDisplayDataTree}
```
looks like the d should be uppercased for consistency.

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+        }
+        if (
+            typeof value === "object" &&
+            (key === "position" || key === "rotation")
+        ) {
+            return (
+                <>
+                    <div> x = {formatFloatForDisplay(value.x)}</div>
+                    <div> y = {formatFloatForDisplay(value.y)}</div>
+                    <div> z = {formatFloatForDisplay(value.z)}</div>
```
toloudis, line 67: any intention to put units with these numbers some day?

src/components/Icons/index.tsx
```diff
@@ -50,6 +51,7 @@ export const DownCaret = (
         style={{ width: "16px", transform: "rotate(180deg)" }}
     />
 );
+export const FilledCaret = <CaretDownFilled />;
```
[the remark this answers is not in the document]
interim17: 1) the caret can be rotated so isn't always pointing down. 2) slightly more readable since all the other icons are declared/renamed in this way. Neither reason is obligatory. We could export and use `{CaretDownFilled}` directly.

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+    uiDisplayData: AgentDisplayNode[];
+}
+
+const MetadataPanel: React.FC<MetadataPanelProps> = ({
+    selectedAgent,
+    uiDisplayData,
+}) => {
+    const [panelExpanded, setPanelExpanded] = React.useState(false);
+
+    const agentSelected = !!selectedAgent.uniqueId;
```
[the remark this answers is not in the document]
interim17: Yes, this line was a requested change from @meganrm. Previously I had this line :
`const agentSelected = followObject.instanceId !== -1;`, and initialized the redux state of selectedAgent with an id of -1. 

Now, it's initialized as an empty object, and in the viewer callback anytime the agent ID is -1 we set the `selectedAgent` in redux to be an empty object, so in theory this should never be -1 in the component. 

This touches on type enforcement in redux in the website which is a little odd...

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+            );
+        }
+        return (
+            <>
+                {selectedAgentNotRendered && (
+                    <p className={styles.agentMissingText}>
+                        Agent no longer in simulation
+                    </p>
+                )}
+                {getMetadataRows}
```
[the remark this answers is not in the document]
interim17: I wrote it initially as a function without `useMemo`, but now I guess that function is the callback argument in `useMemo` and this assignment is really to the return value, so I will rename it!

src/components/MetadataPanel/index.tsx
interim17: Seems useful, wasn't in design

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+            <>
+                Rotation
+                <br />
+                (degrees)
+            </>
+        ),
+        radius: <span>Radius</span>,
+    };
+
+    const getFormattedValue = (
```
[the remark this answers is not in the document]
interim17: Yes, I see what you mean here.

I've consolidated the two into `displayValues: Record<keyof AgentMetadata, MetadataDisplay>` and `metadataRows` just does its lookup in that object, much better I think.

src/components/MetadataPanel/index.tsx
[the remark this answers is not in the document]
meganrm: shouldn't this just be `const agentSelected = !!selectedAgent` ?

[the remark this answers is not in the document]
meganrm: based on what you said that it is being set to an empty object if there is not a selected agent

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const selectedAgentNotRendered =
+        selectedAgent.agentType >= uiDisplayData.length;
+
+    useEffect(() => {
+        if (!agentSelected) {
+            setPanelExpanded(false);
+        } else {
+            setPanelExpanded(true);
+        }
+    }, [selectedAgent.uniqueId]);
```
meganrm, line 33: This effect is a little hard for me to follow. I think it would be clearer to just have 
`if (!selectedAgent) { setPanelExpanded(false)}`
or `if(!selectedAgent.uniqueId) { setPanelExpanded(false)}`
I don't think you need the constant `agentSelected`?

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+                    <div className={styles.value}>
+                        {displayValues[key].value()}
+                    </div>
+                </div>
+                {index < keys.length - 1 && (
+                    <Divider className={styles.divider} />
+                )}
+            </React.Fragment>
+        ));
+    }, [selectedAgent, selectedAgentNotRendered]);
```
meganrm, line 97: why is `selectedAgentNotRendered` in the requirements list? it doesn't seem like it's being used here. And I don't think you need useMemo here because this is going to get recalculated anyway every time the props change (ie, you're not saving any re-renders)

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    const panelContent = (): JSX.Element => {
+        if (!agentSelected) {
```
meganrm, line 100:
```suggestion
        if (!agentSelected) {
```
```suggestion
        if (!selectedAgent) {
```

src/components/MetadataPanel/style.css
```diff
@@ -0,0 +1,84 @@
...
+    min-height: 265px;
+}
+
+.title {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: 4px;
+}
+
+h3 {
```
meganrm, line 20: this is going to be applied to all h3 tags. this needs to be `.container h3`

src/components/MetadataPanel/style.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.row {
+    display: flex;
+    font-size: 12px;
+    font-weight: 100;
+}
+
+.key {
+    color: var(--medium-gray);
```
meganrm, line 72: there are a few repeated `font-weights` `font-size` and colors (ie you also set the `.panel` color to medium-gray. I would consolidate the styles: figure out what the default font with, size and color are, apply those in general, and then use classes to do overrides when needed. also what's the general app font weight?

src/components/MetadataPanel/style.css
```diff
@@ -0,0 +1,84 @@
...
+.agent-missing-text {
+    font-size: 13px;
+    font-weight: 400;
+}
+
+.row {
+    display: flex;
+    font-size: 12px;
+    font-weight: 100;
+}
```
meganrm, line 69: I would put this higher, put all the default text styling here: like the display, font-size, font-weight, color, etc then I don't think you never need the `no-agent-text` class because it's the same and the only deviation is the agent-missing-text

src/state/selection/types.ts
```diff
@@ -76,3 +76,8 @@ export interface SetRecentColorsAction {
     payload: string[];
     type: string;
 }
+
+export interface SetSelectedAgentAction {
+    payload: number;
```
meganrm, line 81: is this type right? isn't the payload either an empty object or an object with all the metadata?

src/state/selection/reducer.ts
```diff
@@ -33,6 +35,7 @@ export const initialState = {
     agentHighlightMap: {},
     colorChange: null,
     recentColors: [],
+    selectedAgent: {},
```
meganrm, line 38: nit, but I would call this `selectedAgentMetadata` and the selector then `getSelectedAgentMetadata`

src/components/MetadataPanel/index.tsx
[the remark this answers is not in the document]
interim17: No I don't think so, empty objects are truthy in JS.

src/components/MetadataPanel/index.tsx
interim17: I can remove the constant, I thought having `agentSelected` was more semantic/readable than `selectedAgent.uniqueId`. Either will work. Could also make it even more explicit, `agentIsSelected` would that be more clear?

Can't use `!selectedAgent` because its always truthy in this current version of the code.

src/components/MetadataPanel/index.tsx
interim17: Ah you're right!

src/components/MetadataPanel/index.tsx
interim17: As discussed above this check will always come back true.

src/state/selection/reducer.ts
interim17: Yeah that's definitely better.

src/state/selection/types.ts
interim17: You're right, it should be `AgentMetadata`. Changing this, and changing everything to be `selectedAgentMetadata` and `setSelectedAgentMetadata` instead of `selectedAgent` and `setSelectedAgent`

src/components/MetadataPanel/style.css
interim17: You're correct, thank you!

src/components/MetadataPanel/style.css
interim17: That's fine, but will override the styling on the text that says "Nothing to see here" so I'll need the same number of classes in the end to fix that

interim17: Ok so I pushed some rules up to `.container` and `.panel-contents` and trimmed where I could, also switched everything to css vars.

Still needed `.no-agent-text` but moved the class onto the previous line and its just font-size now.

src/components/MetadataPanel/index.tsx
interim17: Deleted the constant and I'm using `selectedAgentMetadata.uniqueId` in its place per your feedback! Hopefully this is more readable to you!

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+                </div>
+                {index < keys.length - 1 && (
+                    <Divider className={styles.divider} />
+                )}
+            </React.Fragment>
+        ));
+    };
+
+    const panelContent = (): JSX.Element => {
+        if (!selectedAgentMetadata.uniqueId) {
```
meganrm, line 103: the other option here (although I think this is fine) is use `lodash.isEmpty(selectedAgentMetadata)`

meganrm (approved):

src/components/MetadataPanel/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        radius: {
+            label: <span>Radius</span>,
+            value: () => formatFloatForDisplay(selectedAgentMetadata.radius),
+        },
+    };
+
+    const getMetadataRows = () => {
+        const keys = Object.keys(
+            selectedAgentMetadata
+        ) as (keyof AgentMetadata)[];
```
toloudis, line 84: I like how you are using `keyof AgentMetadata` both here and in displayValues to make sure this is safe.  The `displayValues[key]` lookups look risky to me but the typescript typings are doing their job.

toloudis (approved):

PR #547, opened 2024-07-30
Feature/share embed

### LINKED ISSUE #146: Update supported engines on tutorial page
## Use Case

A user trying to convert their MCell data, or data from another simulator for which we just added support, needs to know the simulator they're using is supported and be able to find the relevant jupyter notebook.

## Solution

On the tutorial page https://simularium.allencell.org/tutorial, we list supported engines and corresponding jupyter notebooks. Add the bold text:

We support the following simulators:
- ReaDDy (https://readdy.github.io/)
- PhysiCell (http://physicell.org/)
- CytoSim (https://gitlab.com/f.nedelec/cytosim)
- **MCell (https://mcell.org/)**
- **Smoldyn (http://www.smoldyn.org/)**
- **SpringSaLaD (https://vcell.org/ssalad)**
- **MEDYAN (http://medyan.org/)**

If you used one of our supported simulators to generate your data, choose the notebook for that simulator:
- ReaDDy
- PhysiCell
- CytoSim
- **[MCell[https://github.com/allen-cell-animated/simulariumio/blob/master/examples/Tutorial_mcell.ipynb]]**
- **[Smoldyn[https://github.com/allen-cell-animated/simulariumio/blob/master/examples/Tutorial_smoldyn.ipynb]]**
- **[SpringSaLaD[https://github.com/allen-cell-animated/simulariumio/blob/master/examples/Tutorial_springsalad.ipynb]]**
- **[MEDYAN[https://github.com/allen-cell-animated/simulariumio/blob/master/examples/Tutorial_medyan.ipynb]]**

interim17:
Time estimate or Size
=======
_medium_

Problem
=======
[Closes #146](https://github.com/simularium/simularium-planning/issues/146)

This is the MVP feature, forthcoming work to be prioritized includes options to adjust what controls are visible, autoplay, default to looping, and addressing visual/UX concerns in the embedded page.

Solution
========
Update the `ShareTrajectoryModal` to generate an iframe that with a link to the embedded viewer, merged a set of updates by @meganrm that improved organization and design, notably pulling the code for the embed snippet into its own component.

Added `VerticalFlexbox`: small reusable component that makes the JSX more readable and prevents having lots of very similar divs each with their own class name on the stylesheet. I think we can add a few more similar container utils over time and do some refactoring to clean up other components.

## Type of change

Steps to Verify:
----------------
1. Poke around the options on the modal.
2. To test the modal you can paste it into the html of the website or another webpage. To use it in a react componant change `allowfullscreen` to `allowFullScreen`

Screenshots:
-----------------------
Screenshot 2024-07-30 at 2 07 37 PM
Screenshot 2024-07-30 at 2 07 28 PM


src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -93,18 +81,218 @@ const ShareTrajectoryModal = ({
...
+        setting: "height" | "width",
+        input: string
+    ) => {
+        const numericInput = input ? parseInt(input) : 0;
+        let roundedLimitedInput;
+
+        if (setting === "width") {
+            roundedLimitedInput = numericInput >= 3840 ? 3840 : numericInput;
+        } else if (setting === "height") {
+            roundedLimitedInput = numericInput >= 2160 ? 2160 : numericInput;
```
interim17, line 204: Setting the max size for the embedded snippet to 4k resolution.

src/styles/utils.tsx
```diff
@@ -0,0 +1,32 @@
...
+            display: "flex",
+            flexDirection: "column",
+            gap: `${gap}px`,
+            alignItems: alignItems,
+            justifyContent: justifyContent,
+        }}
+    >
+        {children}
+    </div>
+);
```
interim17, line 30: Reusable container util prevents a lot of repetitive classes in stylesheets and hopefully makes JSX more immediately readable.

src/components/ShareTrajectoryModal/style.css
```diff
@@ -25,28 +40,58 @@
...
+    font-size: 16px;
+    cursor: pointer;
+}
+
+.embed-settings {
+    margin-bottom: 10px;
+}
+
+.accent-text {
+    font-size: 16px;
```
meganrm, line 60: are all these font-size: 16 s necessary? Same question for the font-weight

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -93,18 +81,218 @@ const ShareTrajectoryModal = ({
...
+                                            e.target.value
+                                        );
+                                    }}
+                                />
+                                <p> pixels </p>
+                                <div className={styles.constrainProportions}>
+                                    <Checkbox
+                                        onChange={() => {
+                                            setEmbedSettings({
+                                                ...embedSettings,
```
meganrm, line 262: constrain proportions shouldn't disable/enable the inputs. it means you keep the ratio of height to width constant. ie, if they change the width and it's checked, it will scale the height appropriately

src/components/ShareTrajectoryModal/style.css
```diff
@@ -25,28 +45,62 @@
...
-    gap: 6px;
+    display: flex;
+    gap: 8px;
 }
 
-.copy-button span:global(.anticon) {
-    margin: 0px;
-    padding-top: 2px;
+.number-inputs {
+    text-align: right !important;
```
meganrm, line 82: can this just be targeted more directly instead of using an important?

src/styles/ant-vars.less
```diff
@@ -142,14 +142,14 @@ https://github.com/ant-design/ant-design/blob/master/components/style/themes/def
...
-@radio-disabled-button-checked-color: @disabled-color;
-@radio-wrapper-margin-right: 8px;
+// @radio-button-bg: @black;
+// @radio-button-checked-bg: @dark-three;
+// @radio-button-color: @grayish-brown;
+// @radio-button-hover-color: @primary-5;
+// @radio-button-active-color: @dark-three;
+// @radio-disabled-button-checked-bg: tint(@black, 90%);
+// @radio-disabled-button-checked-color: @disabled-color;
+// @radio-wrapper-margin-right: 8px;
```
meganrm, line 152: why are these commented out? if w're not going to use them they should just be deleted. but it does look like we had set these up at some point, are there radio buttons somewhere in the app that are using these?

interim17: So I couldn't find anywhere that they seemed to be used, but there are lots of parts of this file that are commented out rather than deleted when un-used... I wasn't sure if we were trying to retain some past config for use when switching to antd 5/away from less vars?

src/util/index.ts
```diff
@@ -121,3 +121,26 @@ export const getIconGlyphClasses = (name: IconGlyphs) => {
...
+    } catch (err) {
+        console.error("Failed to copy text: ", err);
+    }
+};
+
+export const getUrlParamValue = (url: string, param: string) => {
+    const urlObj = new URL(url);
+    const params = new URLSearchParams(urlObj.search);
+    return params.get(param);
+};
```
meganrm, line 137: nit, because there are other param functions in this file, so maybe for another PR, but this should probably be in urlHandling instead

src/components/ShareTrajectoryModal/style.css
interim17: was able to remove!

src/components/ShareTrajectoryModal/style.css
interim17: Was able to remove a few of these size and weight rules

src/util/index.ts
```diff
@@ -121,3 +121,26 @@ export const getIconGlyphClasses = (name: IconGlyphs) => {
...
+
+export const roundToTimeStepPrecision = (
+    input: number,
+    timestep: number
+): number => {
+    const precision = (timestep.toString().split(".")[1] || "").length;
+    const multiplier = Math.pow(10, precision);
+    return Math.round(input * multiplier) / multiplier;
```
toloudis, line 145: kind of surprised this sort of function isn't already somewhere else.

src/styles/utils.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface VerticalFlexboxProps {
+    children: ReactNode;
+    className?: string;
+    gap?: number;
+    alignItems?: string;
+    justifyContent?: string;
+}
+
+const VerticalFlexbox: React.FC<VerticalFlexboxProps> = ({
```
toloudis, line 11: As a component, should this live inside a `styles` directory?  I suggest putting it someplace where utils or other components live.

src/styles/ant-vars.less
toloudis, line 152: might as well delete them

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -74,37 +68,79 @@ const ShareTrajectoryModal = ({
...
     };
 
-    const copyToClipboard = async (): Promise<void> => {
-        try {
-            await navigator.clipboard.writeText(url);
-        } catch (err) {
-            console.error("Failed to copy text: ", err);
-        }
-    };
+    const UrlLinkPanel: JSX.Element = (
```
toloudis, line 84: is there anything reusable about this UrlLinkPanel that it could go into its own tsx file?

toloudis (approved):
A few organization suggestions but overall LGTM.

src/styles/ant-vars.less
interim17: Ok, will do. Makes it tempting to also delete all the other commented out rules...

src/styles/utils.tsx
interim17: Fair point, I'll make a folder for utility components, that's more accurate

src/util/index.ts
interim17: https://github.com/simularium/simularium-website/pull/551

src/styles/ant-vars.less
interim17: https://github.com/simularium/simularium-website/pull/550

src/util/index.ts
meganrm: there is `roundTimeForDisplay` in the same file, but they're slightly different: I'm not sure if we need both or of they could be combined

meganrm (approved):

PR #549, opened 2024-08-05
proposed changes to share-embed

meganrm:
reasons for changes:

1. there seemed to be a lot of extra things stored in state (radio button state, trajectory url, last time etc), reduced it to minimal needed 
2. I had trouble following what was happening because of the nested components, state being set halfway down the file (typically states should be set at the top of the component for readability) => extracted out embed snippet to it's own component 
3. Check with Lyndsay, but it seems like you should just be able to click into the time box, instead of having to first click the radio button and then editing the time 
4. constrain proportions was implemented incorrectly 
5. it seemed that the embed snippet should include the time change too since that's at the top of the modal


src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -0,0 +1,128 @@
...
+import VerticalFlexbox from "../../styles/utils";
+import { getUrlParamValue, copyToClipboard } from "../../util";
+import { DownArrow } from "../Icons";
+
+import styles from "./style.css";
+
+const DEFAULT_WIDTH = 834;
+const DEFAULT_HEIGHT = 480;
+const MAX_WIDTH = 3840;
+const MAX_HEIGHT = 2160;
```
meganrm, line 13: if something is an unchanged constant, just defining them like this is clearer than having an object (which makes it less clear something is not supposed to be edited because attributes can be changes, while `const` can't be)

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -0,0 +1,128 @@
...
+        const currentRatio = width / height;
+        const numericInput = inputIsValid(input) ? parseInt(input) : width;
+        const roundedLimitedInput = Math.min(MAX_WIDTH, numericInput);
+        setWidth(roundedLimitedInput);
+        if (constrainProportions) {
+            setHeight(Math.round(roundedLimitedInput / currentRatio));
+        }
+    };
+
+    const handleChangeHeight = (input: string) => {
```
meganrm, line 49: these two functions have a lot of the same code, but you don't gain a lot by combining them because you still need numerous if/else statements (for the max value, and for setting the opposite value), which makes things less clear. In my opinion, the two functions is easier to read than 1 with multiple if elses

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -29,27 +35,27 @@ const ShareTrajectoryModal = ({
...
-    };
+    const [userEnteredTime, setUserEnteredTime] = React.useState(currentTime);
+
+    const startTime =
+        startTimeToUse === TimeToUse.BEGINNING ? 0 : userEnteredTime;
+    const linkUrl = editUrlParams(
+        window.location.href,
+        startTime.toString(),
+        URL_PARAM_KEY_TIME
+    );
```
meganrm, line 58: this is completely derived from state, so you don't need to also set it in state

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -62,62 +68,46 @@ const ShareTrajectoryModal = ({
...
-                window.location.href,
-                timeValue.toString(),
-                URL_PARAM_KEY_TIME
-            )
-        );
-    };
-
-    const [timeRadioGroupValue, setTimeRadioGroupValue] = React.useState(1);
-    const [allowTimeInput, setAllowTimeInput] = React.useState(true);
-    const currentTime = roundToTimestepPrecision(
```
meganrm, line 86: useState functions should be at the top of the component

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -62,62 +68,46 @@ const ShareTrajectoryModal = ({
...
-            window.location.href,
-            currentTime.toString(),
-            URL_PARAM_KEY_TIME
-        )
-    );
-    const [lastEnteredNumber, setLastEnteredNumber] =
-        React.useState(currentTime);
-    const onRadioChange = (e: RadioChangeEvent) => {
-        setTimeRadioGroupValue(e.target.value);
-        handleAllowTimeInput();
```
meganrm, line 101: these were redundant

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -62,62 +68,46 @@ const ShareTrajectoryModal = ({
...
-                    onChange={onRadioChange}
+                    value={startTimeToUse}
+                    onChange={(e: RadioChangeEvent) => {
+                        setStartTimeToUse(e.target.value);
+                    }}
                 >
                     <Space direction="vertical">
-                        <Radio value={1}>From beginning</Radio>
-                        <Radio value={2}>
+                        <Radio value={TimeToUse.BEGINNING}>
```
meganrm, line 95: it's nicer to have reader friendly values for the radio buttons instead of numbers

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -62,62 +68,46 @@ const ShareTrajectoryModal = ({
...
+                                className={classNames(styles.timeInput, {
+                                    [styles.disabled]:
+                                        startTimeToUse === TimeToUse.BEGINNING,
+                                })}
                                 defaultValue={currentTime}
                                 onChange={handleTimeInput}
+                                value={userEnteredTime}
+                                onClick={() => {
+                                    setStartTimeToUse(TimeToUse.USER_ENTERED);
+                                }}
```
meganrm, line 110: this is the thing I say to ask UX about: but it made sense to me to be  able to just click into this input box

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -146,148 +136,6 @@ const ShareTrajectoryModal = ({
...
-        return `<iframe height="${snippetHeight}" width="${snippetWidth} "src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
-    };
-
-    const [embedSnippet, setEmbedSnippet] = React.useState(
-        generateEmbedSnippet()
-    );
-
-    useEffect(() => {
-        setEmbedSnippet(generateEmbedSnippet());
-    }, [embedSettings]);
```
meganrm, line 192: again, you don't need to save something in state that is completely derived form state

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -146,148 +136,6 @@ const ShareTrajectoryModal = ({
...
-        let roundedLimitedInput;
-
-        if (setting === "width") {
-            roundedLimitedInput = numericInput >= 3840 ? 3840 : numericInput;
-        } else if (setting === "height") {
-            roundedLimitedInput = numericInput >= 2160 ? 2160 : numericInput;
-        }
-
-        setEmbedSettings({ ...embedSettings, [setting]: roundedLimitedInput });
-    };
```
meganrm, line 208: I said it above, but this spread, and changing the value seems to take away the simplicity of just `setWidth` and `setHeight`

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -146,148 +136,6 @@ const ShareTrajectoryModal = ({
...
-                                    value={embedSettings.width}
-                                    className={styles.numberInputs}
-                                    onChange={(e) => {
-                                        handleEmbeddedSizeInputs(
-                                            "width",
-                                            e.target.value
-                                        );
-                                    }}
-                                />
-                                <p> pixels </p>
```
meganrm, line 257: `<p>` should be for an actual paragraph. <span> or <label> are more appropriate semantically here

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -146,148 +136,6 @@ const ShareTrajectoryModal = ({
...
-                                    <Checkbox
-                                        onChange={() => {
-                                            setEmbedSettings({
-                                                ...embedSettings,
-                                                ["userProvidedProportions"]:
-                                                    !embedSettings.userProvidedProportions,
-                                            });
-                                        }}
-                                    ></Checkbox>
-                                    <p>Constrain proportions</p>
```
meganrm, line 268: if you wrap the checkbox in a label, then you can click the constrain and it will uncheck/check the box

interim17 (approved):
Thanks again for these edits it all looks good to me, happy to merge it if you feel good about it and update the other PR

src/components/ShareTrajectoryModal/style.css
```diff
@@ -47,13 +52,19 @@
...
 .settings-button {
     border: none;
-    background-color: (var(--light-theme-bg));
+    background-color: var(--light-theme-bg);
     font-size: 16px;
     cursor: pointer;
 }
 
 .embed-settings {
+    transition: height 0.3s;
```
interim17, line 61: much better with this

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
+    const trajectory = getUrlParamValue(
+        window.location.href,
+        URL_PARAM_KEY_FILE_NAME
+    );
+
+    const [width, setWidth] = React.useState(DEFAULT_WIDTH);
+    const [height, setHeight] = React.useState(DEFAULT_HEIGHT);
+    const [constrainProportions, setConstrainProportions] =
+        React.useState(false);
```
interim17, line 29: This is the one thing I feel a bit confused on. I think I understand what you've done here, basically maintaining the current aspect ratio whenever a new height or width changes. Is that what UX wanted for this?

I took it to mean allowing the user to provide height and width at all, rather than just using the default, which I guess was a misunderstanding.

I find the phrase constrain proportions a bit misleading in both cases. Maintain aspect ratio, or allow size inputs or something like that would be clearer imo.

meganrm: I feel like that's the wording for photoshop, but yeah, I'm 99% certain this is the functionality she meant. if you think the wording is confusing you can ask UX for a change

interim17: I'll take your word for it, thanks for the fix.

Issue #510, opened 2024-04-30
Store user color selections


interim17:
Matthew Akamatsu [has requested](https://github.com/simularium/simularium-viewer/issues/148#issuecomment-2078208706) that color selections persist in the site and/or in a downloadable file.

These two criteria should close [viewer issue #148](https://github.com/simularium/simularium-viewer/issues/148#issuecomment-2078208706). Browser session is probably a website only PR, downloadable trajectory could be a multi-repo issue.

- [ ] Save user color choices in browser session.
- [ ] Save user color choices in downloaded trajectory.

PR #554, opened 2024-08-23
Feature/color session

### LINKED ISSUE #511: retain color selections in browser session
## Description

Users would like their color choices to be saved in the browser so that they can navigate to and from a trajectory after making color selections, and not lose their choices.

Part of closing #510 

## Acceptance Criteria

After making color selections, a user in the same browser session should be able to navigate back to that trajectory and find their color choices have been remembered and applied.

interim17:
Time estimate or Size
=======
_large_

I split some of the changes that were pure renaming/organization into a leading PR to ease this one, but there is still quite a bit of redux boiler plate, as well as tests that inflate the number of lines here.

It's not a small PR either way, but not quite as bad as it seems.

Problem
=======
Users want to store their color selections for later use, this solves one of two approaches to the issue, which is to store those selections in local browser storage.

Closes #511 and advances #510, design review complete.

Solution
========
#573 makes gets some organizing and renaming out of the way, pointing at that branch, the feature is here.

This branch:
 - `currentColorSettings` enum member `Default` or `UserSelected` indicates which settings should be applied
 - upon parsing a trajectory store the initial result of `getUIDisplayData` as the default settings
  - default UI data lives in `trajectory` branch and settings and user selected UI data live in `selection`
- `selection` branch logics manage putting user selections into and out of local storage in the browser
- file names and data structures are compared before applying any stored settings, to prevent applying settings in cases where file names match but data structures do not: util `compareAgentTrees`
- several asynchronous processes have bearing on our ability to compare these data structures, so some logic for handling that lives in more than one place
- in response to user color selections made via colorPicker: update both user selections redux and local storage for this trajectory
- whenever user color selections are applied, expose a "restore defaults" button on the left side panel that previews the default when hovered over, and restores default and clears user selections from redux and local storage when clicked

## Type of change

Steps to Verify:
----------------
1. Make selections via color picker and see that they are applied correctly
2. Load a new trajectory, then return to the previous, selections should still be applied
3. Reload the tab, selections should still be applied.
4. Open the trajectory in a new tab and the same settings should still be applied
5. When defaults are applied "Restore default colors" button should be disabled
6. When active, hovering over "Restore default colors" should temporarily apply default settings, and removing hover should go back to user selections
7. Clicking on "Restore default colors" button should permanently apply defaults, clear out previous selections, and disable button
8. Extra credit: download a trajectory, "Example 1", and change its name to match a networked trajectory "Example 2". Store settings for original "Example 2", then load renamed version that is called "Example 2" but is actually "Example 1" under the hood. Settings should not be applied!


src/containers/ModelPanel/index.tsx
```diff
@@ -114,11 +130,47 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+                            titleText={"Restore color defaults"}
+                            buttonType={ButtonClass.Action}
+                            isDisabled={defaultUiSettingsApplied}
+                            clickHandler={handleRestorDefaultClick}
+                            onMouseEnter={handleRestoreDefaultMouseEnter}
+                            onMouseLeave={handleRestoreDefaultMouseLeave}
+                        />
+                    </div>,
+                    null,
+                ]}
```
interim17, line 173: This is the only new piece of UI work, a button to preview and restore default settings that is active when user selected settings are currently applied.

src/state/trajectory/logics.ts
```diff
@@ -637,6 +661,34 @@ const cancelConversionLogic = createLogic({
...
+                })
+            );
+        } else {
+            dispatch(setUserSelectedUIData([]));
+        }
+        done();
+    },
+    type: SET_DEFAULT_UI_DATA,
+});
+
```
interim17, line 684: One of two places we are handling checks for different trajectories with the same name.

If browser settings were retrieved from storage before this logic ran, we did not apply them as we couldn't verify if they were appropriate for the current trajectory in a rigorous way. We do the check here and apply them if so. If the check fails, we restore initial state for userSelectedUIData to make way for incoming color changes or settings.

src/state/selection/logics.ts
```diff
@@ -32,10 +51,75 @@ const storeColorsLogic = createLogic({
...
+                return;
+            }
+            dispatch(setUserSelectedUIData(uiData));
+            dispatch(
+                setCurrentColorSettings({
+                    currentColorSettings: ColorSettings.UserSelected,
+                })
+            );
+        }
+        done();
```
interim17, line 116: One of two places we check if stored color settings match the current trajectory.

If viewer hasn't provided default data yet, we store browser settings without applying them. The check for whether to apply them will happen in a redux logic when the default arrives.

If default is already in state, we can compare and either apply the stored settings or discard them if the trajectories don't match.

src/containers/ModelPanel/index.tsx
```diff
@@ -114,11 +130,47 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+    };
+
+    const handleRestoreDefaultMouseLeave = () => {
+        if (!defaultUiSettingsApplied) {
+            setCurrentColorSettings({
+                currentColorSettings: ColorSettings.UserSelected,
+            });
+        }
+    };
+
```
interim17, line 155: Hovering previews the default color settings, clicking the button applies them and deletes the old user selected settings.

src/state/trajectory/actions.ts
```diff
@@ -196,3 +195,36 @@ export function cancelAutoconversion(): CancelConversionAction {
...
+}
+
+export function setCurrentColorSettings(payload: {
+    currentColorSettings: ColorSettings;
+}): SetCurrentColorSettingsAction {
+    return {
+        payload,
+        type: SET_CURRENT_COLOR_SETTINGS,
+    };
+}
```
interim17, line 224: When dispatching actions from inside redux logics I have found it necessary to wrap payloads in an object rather than apply them directly, to avoid a type error, the `dispatch` call rejects the type of the raw payload.

So in this case `payload: ColorSettings` threw an error but `{currentColorSettings: ColorSettings}` did not.

I don't love that this is not obvious when reading this file and looking at the different actions and how their payloads are typed, but couldn't find a way to make it work without.

src/containers/ModelPanel/index.tsx
```diff
@@ -2,44 +2,53 @@ import * as React from "react";
...
 
-import SideBarContents from "../../components/SideBarContents";
+import { State } from "../../state/types";
+import { ViewerStatus } from "../../state/viewer/types";
+import { getStatus } from "../../state/viewer/selectors";
+import {
+    RequestNetworkFileAction,
+    ColorSettings,
+    SetCurrentColorSettingsAction,
+} from "../../state/trajectory/types";
```
interim17, line 12: Lot's of the changes to imports in this file are just organization!

src/containers/ModelPanel/index.tsx
```diff
@@ -114,11 +128,47 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+                            titleText={"Restore color defaults"}
+                            buttonType={ButtonClass.Action}
+                            isDisabled={defaultUiSettingsApplied}
+                            clickHandler={handleRestoreDefaultClick}
+                            onMouseEnter={handleRestoreDefaultMouseEnter}
+                            onMouseLeave={handleRestoreDefaultMouseLeave}
+                        />
+                    </div>,
+                    null,
+                ]}
```
interim17, line 171: The only new UI for this feature.

Issue #548, opened 2024-07-31
embed snippet advanced features and design concerns


interim17:
## Use Case

After closing the MVP scope of [generating an embed snippet](https://github.com/simularium/simularium-planning/issues/146) some next steps/possible refinements are:

- [ ] [design/implement minimal controls for small embeds](https://github.com/simularium/simularium-website/issues/553)
- [ ] [full screen button for embedded windows](https://github.com/simularium/simularium-website/issues/586)
- [ ] possibility to preview embedded window that current snippet would generate
- [ ] functionality and ui to toggle autoplay and default looping
- [ ] option to toggle controls on and off

Individual tickets and prioritization TBD

PR #556, opened 2024-09-03
Feature/embed controls

### LINKED ISSUE #553: limited controls for small embedded windows
## Use Case

Embedded trajectories are functional when their size is large enough, but small embedded windows currently have crowded/non-functional controls. UX has addressed this by designing limited control sets to use below different height/width breakpoints.

## Acceptance Criteria

- Below a certain height remove all camera controls except for the home button.
- Below a certain width breakpoint do not render side panels.
- Below another width breakpoint playback controls should be limited to: play/pause button, time slider, home button, and fullscreen button.

interim17:
Time estimate or Size
=======
_medium-large_

Problem
=======
Advances #548 
Closes #553 

Embedded windows are likely to be so small that controls and side panels are non-functional.

I think this code could be a useful step forward for use in any small window and/or mobile contexts, but is limited to embedded windows for now.

Solution
========

- refactored `VerticalFlexbox` to use a higher order component to get both vertical and horizontal flexbox utility components
- small tweaks to Show settings button styling
- added `EmbedPreview` tool to see what snippets will look like when used
- all embedded windows get a full screen button, full features available in fullscreen mode, `ViewerPanel` handles toggling fullscreen
-- Small bit of redux state, `embedFullScreen` added for full screen changes which happen in `ViewerPanel` but need to trigger re-render in `SImularium`
- `Simularium` handles the side panels
- Checks of breakpoints and embedded path location are handled in `Simularium` and `ViewerPanel`
- below a certain height camera controls (right vertical) are reduced to just the home button
- below a certain width playback controls (bottom horizontal) are reduced to a minimal set: play/pause, slider, fullscreen
- when height and width are both low, integrate home button into playback controls
- only render scale bar when there is room for it
- `PlaybackControls` and `CameraControls` each take some new props that control whether they render the standard controls or modified versions.
- `HomeButton` code copied into `PlaybackControls` as it can render in either control set depending on window size.

Screenshots of different states below (screenshots are slightly dated, only meant to show embedded window, preview panel has improved margins in actual code!):
Full size:
Screenshot 2024-09-03 at 10 06 46 AM
Only height reduced below cutoff (minimal camera controls):
Screenshot 2024-09-03 at 10 09 05 AM
First width cutoff (minimal playback controls):
Screenshot 2024-09-03 at 10 07 37 AM
Second width cutoff (no side panels, minimal playback controls):
Screenshot 2024-09-03 at 10 07 49 AM
Third width cutoff (no side panels, no scale bar, minimal playback controls):
Screenshot 2024-09-03 at 10 08 05 AM
Height and medium width cutoff (no scale bar, no camera controls, minimal playback controls with home buttons):
Screenshot 2024-09-03 at 10 08 50 AM
Height and width cutoff (no side panels, no scale bar, no camera controls, minimal playback controls with home buttons):
Screenshot 2024-09-03 at 10 08 18 AM

## Type of change


src/components/CustomModal/index.tsx
```diff
@@ -64,7 +64,7 @@ const CustomModal: React.FC<CustomModalProps> = ({
             onCancel={closeHandler}
             closable={false}
             title={title}
-            footer={footer}
+            footer={footerButtons ? footer : null}
```
interim17, line 67: preview modal has no footer

src/components/FlexboxUtility/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </div>
+    );
+
+    FlexboxUtility.displayName = displayName;
+    return FlexboxUtility;
+};
+
+const VerticalFlexbox = createFlexboxComponent("column", "VerticalFlexbox");
+const HorizontalFlexbox = createFlexboxComponent("row", "HorizontalFlexbox");
+
```
interim17, line 45: These are tangential to the feature but I think will improve readability wherever we can use them, especially in highly nested JSX with many styling classes.

src/components/ShareTrajectoryModal/EmbedPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+    setPreview,
+}) => {
+    document.documentElement.style.setProperty(
+        "--embed-preview-modal-width",
+        `${width + 50}px`
+    );
+    document.documentElement.style.setProperty(
+        "--embed-preview-modal-height",
+        `${height + 112}px`
+    );
```
interim17, line 28: Need a way to get these changing values into the .antd selector, am I missing an easier way?

src/components/ShareTrajectoryModal/EmbedPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return (
+        <CustomModal
+            closeHandler={() => {
+                setPreview(!preview);
+            }}
+            titleText="Preview"
+            className={styles.embedPreviewModal}
+        >
+            <div dangerouslySetInnerHTML={{ __html: embedSnippet }} />
+        </CustomModal>
```
interim17, line 39: There are a few spots with `dangerouslySetInnerHTML` in our codebase, I want to be careful though.

The html input that generate the snippet is disabled, but is this a security risk? We can use a library like `dompurify` to sanitize the html but I didn't want to add a dependency if not necessary.

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -59,75 +62,105 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
...
+        <>
+            {showPreview && (
+                <EmbedPreview
+                    embedSnippet={embedSnippet}
+                    height={height}
+                    width={width}
+                    preview={showPreview}
+                    setPreview={setShowPreview}
+                />
+            )}
```
interim17, line 83: Preview component, I preferred the organization having it in its own file.

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -116,7 +116,7 @@ const ShareTrajectoryModal = ({
             </VerticalFlexbox>
             <VerticalFlexbox gap={10}>
                 <h4> Copy link {Link} </h4>
-                <div className={styles.urlInputContainer}>
+                <HorizontalFlexbox gap={6}>
```
interim17, line 119: Starting to add these in where there are style rules that only specify flex, flex direction, gap etc.

src/components/ShareTrajectoryModal/style.css
```diff
@@ -101,3 +94,8 @@
     font-size: 19px;
     margin: 0px;
 }
+
+.embed-preview-modal :global(.ant-modal-content) {
+    width: var(--embed-preview-modal-width); 
+    height: var(--embed-preview-modal-height);
+}
```
interim17, line 101: These CSS vars are not defined globally until the `EmbedPreview` component sets them

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+            minimalPlaybackControls: belowControlsWidth,
+            minimalCameraControls: belowControlsHeight,
+            onlyRenderBottomControls: belowControlsWidth && belowControlsHeight,
+            showScaleBar:
+                !belowScaleBarWidth &&
+                (!belowControlsWidth || !belowControlsHeight),
+            showCameraControls: !belowControlsWidth || !belowControlsHeight,
+        };
+    }
+
```
interim17, line 520: I debated lots of different ways of handling these settings. Didn't put them in state directly because they would be derived state, although that could reduce lines.

Didn't put the breakpoint state in Redux because only one of the breakpoint values is currently accessed in one another container and the lower components would still need them provided by this container.

src/containers/ViewerPanel/index.tsx
```diff
@@ -503,28 +599,67 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+                                    simulariumController.startRecording
+                                }
+                                stopRecording={
+                                    simulariumController.stopRecording
+                                }
+                            />
+                        )}
+                        {location.pathname === EMBED_PATHNAME && (
+                            <ViewportButton
+                                tooltipText={
```
interim17, line 630: The large amount of conditional rendering this changeset introduces, and the more complex logic behind it makes me think a `ControlsModule` that is a container and could access the redux store directly would be a useful refactor.

It could handle most of what is needed for rendering: `PlaybackControls`, `CameraControls`, `ScaleBar`, `RecordMoviesComponents`, and `FullScreenButton` and could reduce complexity in this file.

This changeset is already large though.

src/components/FlexboxUtility/index.tsx
meganrm: curious: why this instead of antd `Flex`?

interim17: My habit is to use my own JSX/CSS for this kind of thing, and was working backwards from that to a more DRY solution. Also taking inspiration from @ShrimpCryptid styling utilities in timelapse-colorizer.

Are we into antd `Flex`? We can go that way instead, I'm not married to this although it was fun to make.

meganrm: it basically has all the same props/capabilities, and is already imported. I'd say unless it doesn't do something that you need it to do, might as well use the library were already importing

interim17: Bumping to 5 would be so fun, I'd love to do it. 😭

src/components/CustomModal/index.tsx
meganrm: I think it makes more sense to do this where you define the footer. and this line would remain unchanged

meganrm:
is there a design for the preview? It seems weird to me that it's off center, and I was wondering why that choice was made but I can't find a design document linked in the issues

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -59,75 +62,105 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
...
-                                <label> pixels </label>
-                                <label className={styles.constrainProportions}>
-                                    <Checkbox
-                                        onChange={() => {
-                                            setConstrainProportions(
-                                                !constrainProportions
-                                            );
+                    {showEmbedSettingsPanel && (
+                        <VerticalFlexbox gap={4}>
+                            <p className={styles.accentText}> Size </p>
```
meganrm, line 103: shouldn't be a `p`

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -59,75 +62,105 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
...
+                                        onChange={(e) => {
+                                            handleChangeHeight(e.target.value);
                                         }}
-                                    ></Checkbox>
-                                    Constrain proportions
-                                </label>
-                            </div>
+                                    />
+                                    <span> x </span>
+                                    <Input
```
meganrm, line 114: this is my fault, but these two should be switched: it should be w x h, not h x w. especially since they're not labeled

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+        } else {
+            document.documentElement.requestFullscreen().catch((err) => {
+                console.error(
+                    `Error attempting to enable fullscreen: ${err.message}`
+                );
+            });
+        }
+    };
+
+    private get embedDisplaySettings() {
```
meganrm, line 507: if I'm reading this right, this will happen regardless of if it's embedded or not? ie, if they're looking at it on a phone, but it seems like this name isn't correct then

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         }
     };
 
+    private updateDisplayBreakpoints = () => {
+        if (location.pathname === EMBED_PATHNAME) {
+            this.setState({
+                embedDisplayBreakpoints: {
+                    belowControlsHeight:
+                        window.innerHeight <=
+                        EMBEDDED_STANDARD_CONTROLS_MINIMUM_HEIGHT,
```
meganrm, line 471: nit: but this could just be `CONTROLS_MIN_HEIGHT` to keep these lines from being super long

src/components/CameraControls/index.tsx
```diff
@@ -208,13 +224,14 @@ const CameraControls = ({
...
-                tooltipText={"Home view (H)"}
-                tooltipPlacement="left"
-                icon={getIconGlyphClasses(IconGlyphs.Reset)}
-                clickHandler={resetCamera}
-            />
+            {homeViewButton}
         </div>
     );
+
+    const renderControls = minimalControls
```
meganrm, line 231: nit but `minimalControls` and `mininalEmbeddedControls` are really close in name. I would probably change the prop to `showFullControls` for clarity.

src/components/PlaybackControls/index.tsx
```diff
@@ -116,7 +122,50 @@ const PlayBackControls = ({
...
+            disabled={loading || isEmpty}
+            range={false}
+        />
+    );
+
+    const minimalEmbeddedControls = (
+        <div className={styles.embedContainer}>
+            {PlayPauseButton}
+            {TimeSlider}
+            {showHomeViewButton && (
```
meganrm, line 155: why is this separate from `showMinimalControls`?

src/components/ShareTrajectoryModal/EmbedPreview.tsx
meganrm: does setting a padding or margin not work?

src/components/PlaybackControls/index.tsx
```diff
@@ -116,7 +122,50 @@ const PlayBackControls = ({
...
+                <div className={styles.buttonContainer}>
+                    <ViewportButton
+                        tooltipText="PlaybackHome view (H)"
```
meganrm, line 158: shouldn't this just be `Home view (H)`?

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         }
     };
 
+    private updateDisplayBreakpoints = () => {
+        if (location.pathname === EMBED_PATHNAME) {
```
meganrm, line 466: to me it makes more sense to just not call this function than to have this here. since the pattern set up above by the code is to have the checks there. Otherwise, looking at it I would assume you're already updating the breakpoints on mount

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -32,8 +33,10 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
     const [showEmbedSettingsPanel, setShowEmbedSettingsPanel] =
         React.useState(false);
 
+    const [showPreview, setShowPreview] = React.useState(false);
+
     const url = `https://simularium.allencell.org/embed?trajFileName=${trajectory}&t=${startTime}`;
-    const embedSnippet = `<iframe height="${width}" width="${height}" src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
+    const embedSnippet = `<iframe height="${height}" width="${width}" src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
```
meganrm, line 39: also switch height and width here so it matches the inputs

meganrm:
when I run this, I don't ever get minimal controls, even with the smallest screen settings. 
Screenshot 2024-09-24 at 11 30 04 AM

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
+        const { belowControlsHeight, belowControlsWidth, belowScaleBarWidth } =
+            this.state.embedDisplayBreakpoints;
+        return {
+            minimalPlaybackControls: belowControlsWidth,
+            minimalCameraControls: belowControlsHeight,
+            onlyRenderBottomControls: belowControlsWidth && belowControlsHeight,
+            showScaleBar:
+                !belowScaleBarWidth &&
+                (!belowControlsWidth || !belowControlsHeight),
+            showCameraControls: !belowControlsWidth || !belowControlsHeight,
```
meganrm, line 517: I'm trying to follow this, but it's really confusing: partially because you have some things that are in the minimize direction, and some that are in opposite. So you have both ! and ors and ands. I'm going to take a stab at simplifying this

src/containers/ViewerPanel/index.tsx
```diff
@@ -485,7 +574,14 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
                 />
                 {firstFrameTime !== lastFrameTime && (
-                    <div className={styles.bottomControlsContainer}>
+                    <div
+                        className={classNames(
+                            styles.bottomControlsContainer,
+                            this.embedDisplaySettings.onlyRenderBottomControls
+                                ? styles.fullWidthBottomControls
+                                : styles.defaultControlSpacing
+                        )}
```
meganrm, line 583: These names are a bit confusing because minimized controls has been set up to be the deviation from the normal, but here you have both 'fullWidth' and "default". I think just checking the width here makes more sense than using the embedDisplaySettings.

src/containers/Simularium/index.tsx
```diff
@@ -240,8 +240,13 @@ class App extends React.Component<AppProps, AppState> {
             simulariumController,
             changeToLocalSimulariumFile,
             conversionStatus,
+            embedFullscreen,
         } = this.props;
         const isEmbedded = location.pathname === EMBED_PATHNAME;
+        const hideSidePanels =
+            isEmbedded &&
+            window.innerWidth <= EMBEDDED_SIDE_PANEL_MINIMUM_WIDTH &&
+            !embedFullscreen;
```
meganrm, line 249: I'm not sure I understand all these conditionals. why is embedFullscreen an exception?

meganrm:
the X on the modals is disappearing on hover for me

meganrm:
I don't really love the modal that opens another modal. As I said before, it also seems like it should be at least centered. And unless it's a choice, I would remove the default iframe beveled edges too

src/containers/ViewerPanel/index.tsx
meganrm: https://github.com/simularium/simularium-website/pull/568

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +462,62 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         }
     };
 
+    private updateDisplayBreakpoints = () => {
+        if (location.pathname === EMBED_PATHNAME) {
+            this.setState({
```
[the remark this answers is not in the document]
interim17: I did propose that to UX, it's definitely what I would like to do!

interim17:
> is there a design for the preview? It seems weird to me that it's off center, and I was wondering why that choice was made but I can't find a design document linked in the issues

There isn't a design at the moment because I built it for myself for use during development and it got folded into the feature last minute during design review.

meganrm:
> > is there a design for the preview? It seems weird to me that it's off center, and I was wondering why that choice was made but I can't find a design document linked in the issues
> 
> There isn't a design at the moment because I built it for myself for use during development and it got folded into the feature last minute during design review.

For both simplicity of the PR and because I don't think the way it looks is up to our design standards, I would remove the preview feature from this PR. Full screen could have also been separate, but it at least falls under the category of embed controls.

interim17:
I've merged the updated suggestions, and pulled the preview component.

I still feel pretty strongly that the preview is a major quality of life feature for anyone actually wanting to generate snippets, and UX seemed into it, so I'd love to address your design concerns and get it into a PR soon if possible.

meganrm:
> I've merged the updated suggestions, and pulled the preview component.
> 
> I still feel pretty strongly that the preview is a major quality of life feature for anyone actually wanting to generate snippets, and UX seemed into it, so I'd love to address your design concerns and get it into a PR soon if possible.

In my experience, anyone who is copying an embed link already has a place they're planning on pasting it, so I don't think it's a big blocker. I'll put my comments in the issue

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -32,8 +32,8 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
     const [showEmbedSettingsPanel, setShowEmbedSettingsPanel] =
         React.useState(false);
 
-    const url = `https://simularium.allencell.org/embed?trajFileName=${trajectory}&t=${startTime}`;
-    const embedSnippet = `<iframe height="${width}" width="${height}" src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
+    const url = `http://${location.host}/embed?trajFileName=${trajectory}&t=${startTime}}`;
```
meganrm, line 35: origin is probably better here so you don' have to change the protocol

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -59,75 +59,86 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
...
-                                            setConstrainProportions(
-                                                !constrainProportions
-                                            );
+                    {showEmbedSettingsPanel && (
+                        <VerticalFlexbox gap={4}>
+                            <p className={styles.accentText}> Size </p>
+                            <VerticalFlexbox gap={24}>
+                                <HorizontalFlexbox gap={8} alignItems="center">
+                                    <Input
+                                        value={width}
```
meganrm, line 95: one of these is using `value` and the other `defaultValue`

src/containers/ViewerPanel/index.tsx
```diff
@@ -109,6 +123,11 @@ interface ViewerPanelState {
     height: number;
     width: number;
     movieURL: string;
+    embedDisplayBreakpoints: {
+        belowControlsHeight: boolean;
+        belowControlsWidth: boolean;
+        belowScaleBarWidth: boolean;
+    };
```
meganrm, line 130: these aren't used anymore

src/components/ShareTrajectoryModal/style.css
meganrm: these shouldn't be in this PR anymore

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
meganrm: I would also add `frameBorder="0"` to get rid of the auto iframe border

interim17: ok yes to frame border, and for now yes to origin, but just to check my understanding, we don't want to merge either version of using `location` we want to point directly to `simularium.allencell.org` once we get done with review

src/state/viewer/reducer.ts
```diff
@@ -97,6 +99,14 @@ const actionToConfigMap: TypeToDescriptionMap = {
             isLooping: action.payload,
         }),
     },
+    [SET_EMBED_FULLSCREEN]: {
```
meganrm, line 102: can you pull this out into another PR? I have comments/review about this too and I think it would be nicer to get the controls merged and then add the full screen feature

src/components/CameraControls/index.tsx
```diff
@@ -141,7 +144,21 @@ const CameraControls = ({
         }
         lastKeyPressed.current = keyPressed;
     }, [keyPressed]);
-    return (
+
+    const homeViewButton = (
```
meganrm, line 148: In my PR I factored this out into a component within CameraControls so you don't have it copied in the PlaybackControls. Is there a reason you didn't keep that change?

src/components/PlaybackControls/index.tsx
```diff
@@ -116,7 +123,50 @@ const PlayBackControls = ({
...
+            step={timeStep}
+            min={firstFrameTime}
+            max={lastFrameTime}
+            disabled={loading || isEmpty}
+            range={false}
+        />
+    );
+
+    const minimalEmbeddedControls = (
+        <div className={styles.embedContainer}>
```
meganrm, line 153: nit: className could be `minimalControlsContainer` since this isn't going to be only embedded screens

src/components/CameraControls/index.tsx
interim17: I merged in your PR and don't think I changed that in my small tweaks to your PR, so not sure why it didn't come along for the ride.

interim17: Not seeing it in here, am I missing something? https://github.com/simularium/simularium-website/pull/568

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
meganrm: we do want to merge it. when it's on the website, location will be simularium.allencell.org. or staging.simularium.allencell.org which will be more consistent for testing

src/components/CameraControls/index.tsx
meganrm: I forgot to push it! I opened a new PR, the histories are now a little messed up so I wouldn't try to merge it, I would just look at it as the suggestion

meganrm: https://github.com/simularium/simularium-website/pull/571

src/containers/ViewerPanel/index.tsx
```diff
@@ -431,6 +440,41 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
         }
     };
 
+    private get embedDisplaySettings() {
+        const { height, width } = this.state;
+        const belowControlsHeight = height <= CONTROLS_MIN_HEIGHT;
+        const belowControlsWidth = width <= CONTROLS_MIN_WIDTH;
+        const showScaleBar =
+            width > SCALE_BAR_MIN_WIDTH &&
+            (!belowControlsWidth || !belowControlsHeight);
```
meganrm, line 449: this is still too confusing. Is there a cutoff where you hide this? 
Ie can you have `const hideScaleBar < SCALE_BAR_MIN_WIDTH`?
Otherwise you can set it in the conditionals below. ie line 455: `showScaleBar: false` 
line 461 and 467 `showScaleBar: width > SCALE_BAR_MIN_WIDTH`. line 473: `showScaleBar: true`

meganrm (changes requested):
looking good, just two requests left: 
- [x] factor out the repeated code for the home view button 
- [x] find a simpler conditional for the showScaleBar bool

src/containers/ViewerPanel/index.tsx
interim17: The point of these checks is to make sure the scale bar doesn't overlap any controls. The height is relevant because when we drop below that breakpoint, we move the home button into the bottom controls, which changes the width occupied by the bottom controls.

So if we are
below scale bar cutoff: false
above scale bar cut off, above the width control cut off, we have room for scale bar: true
above scale bar cut off, and below width cutoff, we need to know if the bottom controls are padded by the presence of the home button (i.e. check the height cutoff)

Anyway, I like your last suggestion so I'm going to go with that!

meganrm (approved):

src/components/Icons/index.tsx
```diff
@@ -28,6 +28,7 @@ export const Play = <CaretRightOutlined />;
 export const Pause = <PauseOutlined />;
 export const UploadFile = <ImportOutlined />;
 export const DownArrow = <DownOutlined />;
+export const UpArrow = <DownOutlined style={{ transform: "rotate(180deg" }} />;
```
tyler-foster, line 31: Cheeky, but I'm not complaining.

tyler-foster: Is that paren supposed to be closed? I see `rotate(180deg`, not `rotate(180deg)`

tyler-foster (approved):
Thanks for the overview yesterday! It really helped me understand what I was looking at.

LGTM. I really like this PR as a concept. Reactive embed controls are a good idea.

+1 to the idea of upgrading to Ant V5 to avoid having to roll custom flexbox utilities.

src/components/Icons/index.tsx
interim17: Weirdly seems to work without closing it, but awful to look at, I'll close it.

PR #557, opened 2024-09-11
don't pass unneeded custom props to native elements

interim17:
Time estimate or Size
=======
_tiny_

Problem
=======
We are passing custom props to the native checkbox and getting this warning in the console when running main:
```
Warning: React does not recognize the `checkboxLevel` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `checkboxlevel` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

Solution
========
Deleting props instead of destructuring to remove (first commit) because linter was mad about it.


ShrimpCryptid (approved):
LGTM!

meganrm (approved):

PR #568, opened 2024-09-26
a simplified way of handling min controls

meganrm:
some simplifications after reviewing the PR. I don't think this is working perfectly but I don't know where the designs for this is so I can't tell exactly what the functionality is supposed to be. 

One assumption I did make is some of these minimal controls are fine regardless of if it's embedded or not (if the controls don't fit we shouldn't show them). But that can be changed back.


src/components/CameraControls/index.tsx
```diff
@@ -228,10 +229,12 @@ const CameraControls = ({
...
-        : fullControls;
-
-    return renderControls;
+    if (displayType === CameraControlsDisplay.None) {
+        return null;
+    } else if (displayType === CameraControlsDisplay.Min) {
+        return minimalEmbeddedControls;
+    } else {
+        return fullControls;
+    }
```
meganrm, line 238: since you already had conditional rendering it made sense to do it all in once place. alternatively you could make a different component and push the conditional rendering up higher.

src/components/PlaybackControls/index.tsx
```diff
@@ -29,8 +32,7 @@ interface PlayBackProps {
     displayTimes: DisplayTimes;
     timeUnits: TimeUnits;
     isEmpty: boolean;
-    minimalControls: boolean;
-    showHomeViewButton: boolean;
+    displayType: PlaybackControlsDisplay;
```
meganrm, line 35: just combined the two props into one

src/containers/ViewerPanel/index.tsx
```diff
@@ -575,12 +581,11 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
-                        className={classNames(
-                            styles.bottomControlsContainer,
-                            this.embedDisplaySettings.onlyRenderBottomControls
-                                ? styles.fullWidthBottomControls
-                                : styles.defaultControlSpacing
-                        )}
+                        className={classNames(styles.bottomControlsContainer, {
+                            [styles.fullWidthBottomControls]:
+                                playBackControlsType ===
+                                PlaybackControlsDisplay.BottomOnly,
```
meganrm, line 587: if you have a default css property, just put that on the main class, and then override it with a conditional class

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -36,7 +36,7 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
     const [showPreview, setShowPreview] = React.useState(false);
 
     const url = `https://simularium.allencell.org/embed?trajFileName=${trajectory}&t=${startTime}`;
-    const embedSnippet = `<iframe height="${height}" width="${width}" src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
+    const embedSnippet = `<iframe width="${width}" height="${height}" src="${url}" title="Simularium" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>`;
```
interim17, line 39: to see the small controls during development the url needs to be changed to this:
`http://localhost:9001/embed?trajFileName=${trajectory}&t=${startTime}`

once this is merge the line that is committed should show the small controls

meganrm: you could fix this by using `location.host`

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -104,18 +104,18 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
...
                                         className={styles.numberInputs}
                                         onChange={(e) => {
-                                            handleChangeHeight(e.target.value);
+                                            handleChangeWidth(e.target.value);
                                         }}
                                     />
                                     <span> x </span>
                                     <Input
-                                        value={width}
+                                        defaultValue={height}
```
meganrm, line 115: switching these made me notice that one of these has value and one as defaultValue. seems to be working, but should be consistent

src/containers/ViewerPanel/index.tsx
```diff
@@ -505,17 +490,36 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
-            showScaleBar:
-                !belowScaleBarWidth &&
-                (!belowControlsWidth || !belowControlsHeight),
-            showCameraControls: !belowControlsWidth || !belowControlsHeight,
-        };
+        const { height, width } = this.state;
+        const belowControlsHeight = height <= CONTROLS_MIN_HEIGHT;
+        const belowControlsWidth = width <= CONTROLS_MIN_WIDTH;
+        const showScaleBar = width > SCALE_BAR_MIN_WIDTH;
+
```
meganrm, line 497: if we're only doing this for embeded, then you could add a default return up top if it's not at the embed pathname

src/containers/ViewerPanel/style.css
```diff
@@ -10,14 +10,11 @@
...
-}
-
-.default-control-spacing {
     right: 25%;
     width: 50%;
 }
 
 .full-width-bottom-controls {
     right: 5%;
     left: 5%;
```
interim17, line 19:
```suggestion
    left: 5%;
    width: 90%;
```

I like this change, but we also need to override the width property in this case.

src/containers/ViewerPanel/index.tsx
```diff
@@ -505,17 +490,36 @@ class ViewerPanel extends React.Component<ViewerPanelProps, ViewerPanelState> {
...
-            onlyRenderBottomControls: belowControlsWidth && belowControlsHeight,
-            showScaleBar:
-                !belowScaleBarWidth &&
-                (!belowControlsWidth || !belowControlsHeight),
-            showCameraControls: !belowControlsWidth || !belowControlsHeight,
-        };
+        const { height, width } = this.state;
+        const belowControlsHeight = height <= CONTROLS_MIN_HEIGHT;
+        const belowControlsWidth = width <= CONTROLS_MIN_WIDTH;
+        const showScaleBar = width > SCALE_BAR_MIN_WIDTH;
```
interim17, line 496:
```suggestion
        const showScaleBar = width > SCALE_BAR_MIN_WIDTH && (!belowControlsWidth || !belowControlsHeight);
```

Your changes are definitely an improvement, but without this bit of logic the scale bar ends up on top of the controls sometimes.

src/components/PlaybackControls/index.tsx
```diff
@@ -217,7 +218,9 @@ const PlayBackControls = ({
 
     return (
         <div className={styles.container}>
-            {minimalControls ? minimalEmbeddedControls : fullControls}
+            {PlaybackControlsDisplay.Full
```
interim17, line 221:
```suggestion
            {displayType === PlaybackControlsDisplay.Full
```
Small to tweak to ensure we are evaluating the prop as its not a boolean.

src/components/ShareTrajectoryModal/EmbedSnippetPanel.tsx
```diff
@@ -36,7 +36,7 @@ const EmbedSnippetPanel = ({ startTime }: EmbedSnippetPanelProps) => {
     const [showPreview, setShowPreview] = React.useState(false);
 
     const url = `https://simularium.allencell.org/embed?trajFileName=${trajectory}&t=${startTime}`;
```
interim17, line 38:
```suggestion
    const url = `http://${location.host}/embed?trajFileName=${trajectory}&t=${startTime}}`;
```
Per your suggestion on the other PR, this renders the appropriate preview

interim17:
@meganrm thanks for these updates!

I know you weren't seeing the previews working properly when making your suggestions so I found a few things that needed tweaking to get all the rendering working, left them as suggestions for now I know you're heading into the weekend.

I love keeping this changes for all windows, not just embedded, that's definitely where I saw this going!

PR #573, opened 2024-10-02
Feature/color sessions prep

interim17:
Time estimate or Size
_tiny_

Problem
=======
Color sessions PR is large and cumbersome, these changes are all just renaming of existing redux plumbing that should not effect functionality, stripping these non-functional changes out into separate PR to improve readability there.

New names makes sense in the context of new feature if they are not obvious to you, so please refer to that PR or [this closed PR](https://github.com/simularium/simularium-website/pull/545) for context if desired.

Solution
========
agentUiNames/receiveAgentNamesAndState -> defaultUIData
setColorChanges -> applyUserColorSelections

### DIFF
--- a/src/containers/ModelPanel/index.tsx
+++ b/src/containers/ModelPanel/index.tsx
@@ -2,7 +2,10 @@ import * as React from "react";
 import { ActionCreator } from "redux";
 import { connect } from "react-redux";
 
-import SideBarContents from "../../components/SideBarContents";
+import { State } from "../../state/types";
+import { ViewerStatus } from "../../state/viewer/types";
+import { getStatus } from "../../state/viewer/selectors";
+import { RequestNetworkFileAction } from "../../state/trajectory/types";
 import {
     requestTrajectory,
     changeToNetworkedFile,
@@ -11,35 +14,32 @@ import {
     getUiDisplayDataTree,
     getIsNetworkedFile,
 } from "../../state/trajectory/selectors";
-import { getStatus } from "../../state/viewer/selectors";
-import { State } from "../../state/types";
+import {
+    AgentRenderingCheckboxMap,
+    ChangeAgentsRenderingStateAction,
+    SetVisibleAction,
+    SetRecentColorsAction,
+    ApplyUserColorSelectionAction,
+} from "../../state/selection/types";
+import {
+    turnAgentsOnByDisplayKey,
+    highlightAgentsByDisplayKey,
+    setAgentsVisible,
+    setRecentColors,
+    applyUserColorSelection,
+} from "../../state/selection/actions";
 import {
     getAgentVisibilityMap,
     getAgentHighlightMap,
     getRecentColors,
     getSelectedAgentMetadata,
 } from "../../state/selection/selectors";
-import {
-    turnAgentsOnByDisplayKey,
-    highlightAgentsByDisplayKey,
-    setAgentsVisible,
-    setColorChange,
-    setRecentColors,
-} from "../../state/selection/actions";
-import {
-    ChangeAgentsRenderingStateAction,
-    SetColorChangeAction,
-    SetVisibleAction,
-    AgentRenderingCheckboxMap,
-    SetRecentColorsAction,
-} from "../../state/selection/types";
 import CheckBoxTree, { AgentDisplayNode } from "../../components/AgentTree";
-import { AgentMetadata } from "../../constants/interfaces";
 import NoTrajectoriesText from "../../components/NoTrajectoriesText";
-import { RequestNetworkFileAction } from "../../state/trajectory/types";
-import { ViewerStatus } from "../../state/viewer/types";
 import NetworkFileFailedText from "../../components/NoTrajectoriesText/NetworkFileFailedText";
 import NoTypeMappingText from "../../components/NoTrajectoriesText/NoTypeMappingText";
+import SideBarContents from "../../components/SideBarContents";
+import { AgentMetadata } from "../../constants/interfaces";
 import {
     getSelectAllVisibilityMap,
     getSelectNoneVisibilityMap,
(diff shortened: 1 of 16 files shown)


src/state/selection/constants.ts
```diff
@@ -18,6 +18,8 @@ export const SET_AGENTS_VISIBLE = makeSelectConstant("set-agents-visible");
 export const RESET_AGENT_SELECTIONS_AND_HIGHLIGHTS = makeSelectConstant(
     "reset-selections-and-highlighted-agents"
 );
-export const SET_COLOR_CHANGES = makeSelectConstant("set-color-changes");
+export const APPLY_USER_COLOR_SELECTION = makeSelectConstant(
+    "apply-user-color-selection"
```
meganrm, line 22: nit, but since this is already the selection branch, you could call this `APPLY_USER_COLOR` (or `APPLY_USER_COLORS` whichever makes more sense)

src/state/selection/logics.ts
```diff
@@ -32,10 +32,10 @@ const storeColorsLogic = createLogic({
             }
             return newAgent;
         });
-        dispatch(receiveAgentNamesAndStates(newUiData));
+        dispatch(setDefaultUIData(newUiData));
```
meganrm, line 35: I'm confused by this naming: why are we setting default data when the `APPLY_USER_COLOR_SELECTION` action has been triggered?

src/state/trajectory/reducer.ts
```diff
@@ -37,7 +38,7 @@ export const initialState = {
     timeUnits: null,
     scaleBarLabel: "",
     agentIds: [],
-    agentUiNames: [],
+    defaultUIData: [],
```
meganrm, line 41: is this still an empty array? I don't remember the shape of UIData, but null might make more sense if UIData isn't an array

src/containers/ModelPanel/index.tsx
```diff
@@ -11,35 +14,32 @@ import {
...
-    getAgentHighlightMap,
-    getRecentColors,
-    getSelectedAgentMetadata,
-} from "../../state/selection/selectors";
+    AgentRenderingCheckboxMap,
+    ChangeAgentsRenderingStateAction,
+    SetVisibleAction,
+    SetRecentColorsAction,
+    ApplyUserColorSelectionAction,
+} from "../../state/selection/types";
```
meganrm, line 23: are these all just rearrangements ?

interim17: Yes, should be, the imports were just not in order and it was bothering me and making things hard to track.

src/state/trajectory/reducer.ts
interim17: UIDisplayData is an array.

src/state/selection/logics.ts
interim17: This is slightly confusing as it is somewhat "in between" the old way and the new way.

In this PR where I am not making functional changes, there is only one piece of UI Display Data, what is currently called `agentUiNames` and I propose to call `defaultUiData` in the future. Here in the current paradigm it gets modified when users make color selections. The names only really make sense in the context of the next PR. I'm just pulling these changes out to make that more readable, if its more confusing I can just point the other PR at main.

src/state/selection/constants.ts
interim17: For sure, that's better

src/containers/ModelPanel/index.tsx
```diff
@@ -11,35 +14,32 @@ import {
...
+} from "../../state/selection/selectors";
 import CheckBoxTree, { AgentDisplayNode } from "../../components/AgentTree";
-import { AgentMetadata } from "../../constants/interfaces";
 import NoTrajectoriesText from "../../components/NoTrajectoriesText";
-import { RequestNetworkFileAction } from "../../state/trajectory/types";
-import { ViewerStatus } from "../../state/viewer/types";
 import NetworkFileFailedText from "../../components/NoTrajectoriesText/NetworkFileFailedText";
 import NoTypeMappingText from "../../components/NoTrajectoriesText/NoTypeMappingText";
+import SideBarContents from "../../components/SideBarContents";
+import { AgentMetadata } from "../../constants/interfaces";
```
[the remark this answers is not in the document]
interim17: They aren't added, they are just moved around, trying to get this big list of imports better organized.

src/state/trajectory/reducer.ts
interim17: I remember we disagreed about this once before, but I do wonder if reducer/initialState would be easier to understand at a glance with more explicit typing?

meganrm: it's not a problem as long as this is the correct empty state

src/state/selection/logics.ts
```diff
@@ -1,15 +1,15 @@
...
-import { receiveAgentNamesAndStates } from "../trajectory/actions";
+import { getDefaultUIData } from "../trajectory/selectors";
+import { setDefaultUIData } from "../trajectory/actions";
+import { APPLY_USER_COLOR } from "./constants";
 
 const storeColorsLogic = createLogic({
     process(deps: ReduxLogicDeps, dispatch, done) {
         const { action, getState } = deps;
-        const uiData: UIDisplayData = getAgentDisplayNamesAndStates(getState());
+        const uiData: UIDisplayData = getDefaultUIData(getState());
```
toloudis, line 12: Minor concern about this name shortening from the true UIDisplayData to UIData in function names like getDefaultUIData.  It feels like the function is slightly lying to me.

toloudis (approved):

src/state/selection/logics.ts
interim17: Ok! I'm fine to make the name more explicit

meganrm (approved):

PR #584, opened 2024-10-14
Feature/user colors

### LINKED ISSUE #579: differentiate default color settings from changed color settings
## Use Case

Advances #511 

Color sessions work requires differentiating between default color settings and user selections. We need to add a piece of global state to represent changed/updated color settings as distinct from `defaultUiData`

## Acceptance Criteria

There will be a new piece of data in redux to represent updated/changed color selections, it will initially match the default ui data received when parsing the trajectory. 

As color selections are made in the UI, this new piece of state should be updated and the default should be unchanged.

interim17:
Time estimate or Size
=======
_small_

Problem
=======
Closes #579, advances #511 

We need to differentiate between the default UI data derived by parsing trajectories and the current UI data that includes user color selections. This PR adds `selectedUIDisplayData` to redux, and points the `getSelectionStateInfoForViewer` selector at it. When default arrives it sets the initial state for selected, and then as users make color changes we only update the selected data, and the default remains unchanged.

Steps to Verify:
----------------
1. Nothing should be functionally changed, loading trajectories and changing colors should work as before.


meganrm (approved):

src/state/trajectory/logics.ts
```diff
@@ -638,6 +640,15 @@ const cancelConversionLogic = createLogic({
...
 });
 
+const setInitialSelectedUIData = createLogic({
+    process(deps: ReduxLogicDeps, dispatch) {
+        const { action } = deps;
+        const uiData = action.payload;
+        dispatch(setSelectedUIDisplayData(uiData));
+    },
+    type: SET_DEFAULT_UI_DATA,
+});
```
frasercl, line 650: This logic looks like it just re-dispatches the triggering action with a different type? If not, what am I missing; if so, why not just use that action in place of the one that triggers this logic?

src/state/selection/logics.ts
```diff
@@ -32,7 +32,7 @@ const storeColorsLogic = createLogic({
             }
             return newAgent;
         });
-        dispatch(setDefaultUIData(newUiData));
+        dispatch(setSelectedUIDisplayData(newUiData));
         done();
```
frasercl, line 36: Still a bit bummed to see this synchronous state derivation code in a logic. Everything this code does would work in a reducer branch, and would probably be more at home there, where it wouldn't have to re-dispatch a new action and immediately call `done` to clear out resources held by `redux-logic` for what it assumes to be an `async` operation. That change is well out of scope for this PR though.

frasercl (approved):
Looks okay overall, though I have a couple concerns about the use of logics.

src/state/trajectory/logics.ts
interim17: Your point is valid!

Context: this was a larger PR, and I got the feedback that I should chunk things up in more bite size pieces. You're correct that this doesn't really need this logic yet, we can just call both `setDefaultUIData` and `setSelectedUIDisplayData` in `ViewerPanel`, and I would happily make that change for now.

I know because I have a crystal ball, that we are likely going to need this asynchronous logic because we won't know what happens first: parsing of the trajectory to derive `defaultUIData` and retrieval from browser storage of `selectedUIData`.

This is a problem of my making! Trying to work backwards from bigger feature branches into smaller PRs can lead to issues like this.

Maybe the defining of the subtasks is the real problem, as this whole PR is somewhat contrived and serves no purpose except to lay the groundwork for forthcoming code, so I wonder what is the greater sin: using this logic to set derived state, or correctly storing state that is never accessed in the first place.

src/state/selection/logics.ts
interim17: This seems like a good opportunity for a refactor. This logic was put in place during previous work on color changing, and its role has changed.

We can derive the new `UiDisplayData` and pass it into `redux` directly, redux never needs to handle the `ColorChange`. I am inclined to do this in `ModelPanel` instead of in the reducer because the pattern in this repo is for reducers to simply pass on their action payload without side effects.

Adding (synchronous) logic to a reducer feels like breaking that pattern in a way that puts code in _more places_ which is part of the pain of redux, for me at least. But I accept that this `redux-logic` is not the right place for it either.

interim17: Wrote a brief issue (#596 ) and will address in a separate PR.

PR #606, opened 2024-10-31
Feature/selector org

interim17:
Time estimate or Size
=======
_small/medium_
Changes related to directory organization, connected to #607, will not merge to main until both PRs are approved

Problem
=======
Advances #511 

Solution
========
@frasercl raised some interesting concerns about state branches in a previous PR

We have redux selectors in `/state` that are:
- basic selection of piece of state
- more complex derivation of state within a branch
- container specific selectors (`ModelPanel`, `ViewerPanel`, etc.)

Now adding `compoundSelectors` to `/state` to account for selectors that consume state from multiple branches and are used multiple containers.

This resolution keeps the actual data for `defaultUIDisplayData` in the `trajectory` branch and `selectedUIDisplayData` and `ColorSettings` in the `selection` branch, and avoids circular dependencies when the branches try to import each other's selectors.

This work is essentially part of #607, but so many lines of moving/organizing code was obscuring the readability of that PR. I will not merge either PR until both are approved. It is a non-breaking change, the app should run as before.

### getCurrentUIData
The selector which pulls from multiple branches to determine the current color settings for the app, and to pass to the viewer.

### getUiDisplayDataTree
Previously in `trajectory` moved to `ModelPanel` as it depends on a compound selector, and is only used in once place, could also be in `compoundSelectors` folder...

### DIFF
--- a/src/containers/ModelPanel/selectors.test.ts
+++ b/src/containers/ModelPanel/selectors.test.ts
@@ -1,7 +1,9 @@
+import { initialState, State } from "../../state";
 import {
     getSelectAllVisibilityMap,
     getSelectNoneVisibilityMap,
     getIsSharedCheckboxIndeterminate,
+    getUiDisplayDataTree,
 } from "./selectors";
 
 const mockUiDisplayData = [
@@ -50,9 +52,8 @@ const mockUiDisplayData = [
 describe("ModelPanel selectors", () => {
     describe("getSelectAllVisibilityMap", () => {
         it("Returns an agent visibility map with all possible states", () => {
-            const result = getSelectAllVisibilityMap.resultFunc(
-                mockUiDisplayData
-            );
+            const result =
+                getSelectAllVisibilityMap.resultFunc(mockUiDisplayData);
             const expected = {
                 agentWithChildren1: ["", "state1"],
                 agentWithChildren2: ["", "state1"],
@@ -64,9 +65,8 @@ describe("ModelPanel selectors", () => {
 
     describe("getSelectNoneVisibilityMap", () => {
         it("Returns an agent visibility map with none of the possible states", () => {
-            const result = getSelectNoneVisibilityMap.resultFunc(
-                mockUiDisplayData
-            );
+            const result =
+                getSelectNoneVisibilityMap.resultFunc(mockUiDisplayData);
             const expected = {
                 agentWithChildren1: [],
                 agentWithChildren2: [],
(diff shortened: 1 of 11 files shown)


src/containers/ModelPanel/selectors.test.ts
```diff
@@ -164,4 +164,67 @@ describe("ModelPanel selectors", () => {
             expect(result).toBe(true);
         });
     });
+    describe("getUiDisplayDataTree", () => {
+        it("returns an empty array if ui display data is empty", () => {
+            expect(getUiDisplayDataTree(initialState)).toStrictEqual([]);
+        });
```
interim17, line 170: Just moved

src/state/compoundSelectors/index.ts
```diff
@@ -0,0 +1,31 @@
...
+        const fileHasBeenParsed = defaultData.length > 0;
+        if (!fileHasBeenParsed) {
+            return [];
+        }
+        if (colorSetting === ColorSettings.UserSelected) {
+            return sessionData;
+        }
+        return defaultData;
+    }
+);
```
interim17, line 31: Important in forthcoming work but not functionally relevant until it replaces `getSelectedUIDisplayData` in `getSelectionStateInfoForViewer`

Mostly here to demonstrate the reasoning behind `compoundSelectors`

src/state/compoundSelectors/index.ts
```diff
@@ -0,0 +1,31 @@
+import { createSelector } from "reselect";
+
+import { getDefaultUIDisplayData } from "../trajectory/selectors";
+import {
+    getCurrentColorSettings,
+    getSelectedUIDisplayData,
+} from "../selection/selectors";
+import { ColorSettings } from "../selection/types";
+import { UIDisplayData } from "@aics/simularium-viewer";
```
meganrm, line 9: nit: organize imports

src/state/selection/types.ts
```diff
@@ -87,3 +87,8 @@ export interface SetSelectedUIDisplayDataAction {
     payload: UIDisplayData;
     type: string;
 }
+
+export enum ColorSettings {
```
meganrm, line 91: nit: make this singular. So when it's used `colorSetting.Default` it makes sense. it's a single setting

src/state/selection/selectors/basic.ts
```diff
@@ -13,3 +13,5 @@ export const getSelectedAgentMetadata = (state: State) =>
     state.selection.selectedAgentMetadata;
 export const getSelectedUIDisplayData = (state: State) =>
     state.selection.selectedUIDisplayData;
+export const getCurrentColorSettings = (state: State) =>
+    state.selection.currentColorSettings;
```
meganrm, line 17: should also be singular (otherwise it sounds like it's going to be a list of colors, when it's just the singular binary setting)

src/state/selection/types.ts
meganrm: I don't have a good suggestion for this name, but it could be confusing because at first glance it seems like it's going to be a list of colors. the singular will help, but something like, `colorsToUse` or `colorsOrigin`?

src/state/selection/selectors/basic.ts
interim17: What about `currentColorSetting`/`getCurrentColorSetting` for state and `AgentColorMapping` for the typing.

Examples:
`if (colorSetting === AgentColorMapping.UserSelected)`
or
`currentColorSetting: AgentColorMapping.Default,`

meganrm: I wouldn't call it a Mapping because that generally means it's a lookup "table". but this is just the value you use in another lookup

meganrm (approved):

src/state/compoundSelectors/index.ts
frasercl, line 1: Maybe a short comment at the top of this file that describes what this file is for, like you did in the PR description?

I also see that the parent `state` directory has a README.md that (among other things) goes over the file structure for all this redux stuff - consider updating that as well

frasercl (approved):
I like it!

PR #607, opened 2024-10-31
retrieve stored color data from browser and apply it

### LINKED ISSUE #582: ability to retrieve stored color settings from browser storage
## Use Case

Advances #511 

When users have stored color settings in the local browser storage they should be retrieved and applied after the trajectory has been loaded and parsed

## Acceptance Criteria

Trajectories with color settings in local storage should have their settings verified to confirm the data matches, and if there is a match in agent structure, but the color assignments are different, the stored settings should be applied and put into redux.

Address concerns about appropriate use of `redux-logic` in: https://github.com/simularium/simularium-website/pull/584

interim17:
Time estimate or Size
=======
_medium_

Problem
=======
Advances #511 Closes #582

Solution
========

Color settings stored in the browser are retrieved, validated, and applied.

The final out of scope piece of this feature is returning to the default settings.

### updateSelectedUIDisplayData
- In `ModelPanel` define a callback for handling color changes (passed to `ColorPicker`) that takes a `ColorChange`, derive new `UIDisplayData`, send that to redux, and trigger a change in `ColorSettings` from `Default` to `UserSelected`.

### Selection branch logics
- `storeDisplayDataInBrowserLogic` for stashing settings in the browser storage.
- `applySessionColorsLogic` which retrieves settings from the browser, validates the data, and ensures that the trajectory has been parsed before applying the settings.

### Trajectory branch logics
- File load flows account for grabbing settings from browser storage.
- `receiveDefaultUIDataLogic` checks for valid settings that might have been retrieved from the browser before the trajectory could be parsed. The selection branch logic would have stored these settings for later validation in this case. If present and valid, they are applied.

## Type of change


src/containers/ModelPanel/index.tsx
```diff
@@ -81,10 +89,25 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+            colorChange,
+            currentUIDisplayData
+        );
+        setSelectedUIDisplayData(newUIData);
+        setCurrentColorSettings({
+            currentColorSettings: ColorSettings.UserSelected,
+        });
+        storeDisplayDataInBrowser(newUIData);
+    };
+
```
interim17, line 110: Calling all relevant redux action creators here in the function, and passing it down to `ColorPicker` as a callback. Also doing the conversion from `ColorChange` to `UIDisplayData` before hitting the `redux` plumbing to address some concerns raised by @frasercl in a previous PR about abuse of `redux-logic`.

I think there are other places in the repo where we drill action creators down as props (and sometimes also rename them several times) that are a bit confusing, maybe worth refactoring some of them, I like this approach of defining a callback as it keeps the `containers` as the only components that directly handle the redux functions.

src/state/trajectory/logics.ts
```diff
@@ -640,11 +652,30 @@ const cancelConversionLogic = createLogic({
...
+        /**
+         * If the conditions below are true, then valid color settings were
+         * retrieved from browser storage before the default ui data arrived
+         * and those settings should be applied.
+         * If false then userSelected data either doesn't exist, hasn't arrived yet, or
+         * is invalid: set to empty array for now.
+         */
+        if (
+            browserStoredUIData.length > 0 &&
+            compareAgentTrees(browserStoredUIData, action.payload)
```
meganrm, line 668: I would rename this to `isSameAgentTree` since you're using it as a bool and the naming is ambiguous

src/containers/ModelPanel/index.tsx
```diff
@@ -79,10 +89,25 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+    const updateSelectedUiDisplayData = (colorChange: ColorChange) => {
+        const newUIData = applyColorChangeToUiDisplayData(
+            colorChange,
+            currentUIDisplayData
+        );
+        setSelectedUIDisplayData(newUIData);
+        setCurrentColorSetting({
+            currentColorSetting: ColorSetting.UserSelected,
+        });
+        storeDisplayDataInBrowser(newUIData);
```
meganrm, line 108: without digging into the details, but just looking at this, it seems like a lot of state setting to be happening in a component. I would move this into a logic.

interim17: So this is work I specifically moved _out of logics_  in conversation with @frasercl about being more strict in the use of `redux-logic` for handling async operations only.

Can I request that we converge on an idea of what `logics` are for and then use them that way, because at this point I'll be changing things back and forth between different PR comment cycles.

interim17: Checking the `state/README.md` says that `Logic` is for: "AJAX calls or complex business logic", and not strictly for async code.

@frasercl can you live with this philosophy?

meganrm: I would consider communicating with the browser storage as syncing with an external api

frasercl: I agree with the "AJAX calls" element of the description above but not necessarily with the "complex business logic" part. As I understand it (and once again, I'm not an authority, I'm trying to work all this out too!), part of redux's model is that actions don't necessarily take the form of "here's a specific element of state I want to set." They correspond to the real-world events that might require a state transformation - literally all the "actions" that a user or an external system are able to take. For instance, if clicking some button requires updating the enabled/disabled state of an agent _and_ setting its color, you don't necessarily want to dispatch a "set enabled" action _and_ a "set color" action, you might just want an "enabled/color change" action. So in general, if you're finding yourself dispatching a lot of actions and want to consolidate them, the tool to reach for is a more complex reducer, not necessarily a logic.

In this case though, where _both_ complex business logic and external API synchronization are happening, I think a logic makes sense.

src/state/trajectory/logics.ts
```diff
@@ -640,11 +652,30 @@ const cancelConversionLogic = createLogic({
...
+        ) {
+            dispatch(
+                setCurrentColorSetting({
+                    currentColorSetting: ColorSetting.UserSelected,
+                })
+            );
+        } else {
+            dispatch(setSelectedUIDisplayData([]));
+        }
```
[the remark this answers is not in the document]
frasercl: on a second look, this is not a better idea

src/state/trajectory/logics.ts
```diff
@@ -640,11 +652,30 @@ const cancelConversionLogic = createLogic({
...
 
-const setInitialSelectedUIData = createLogic({
-    process(deps: ReduxLogicDeps, dispatch) {
-        const { action } = deps;
-        const uiData = action.payload;
-        dispatch(setSelectedUIDisplayData(uiData));
+const receiveDefaultUIDataLogic = createLogic({
+    process(deps: ReduxLogicDeps, dispatch, done) {
+        const { getState, action } = deps;
+        const browserStoredUIData = getSelectedUIDisplayData(getState());
```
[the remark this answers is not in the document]
interim17: I feel silly saying this now, but I think there is different part of the loading cycle in the viewer I can hook into to make this irrelevant. There is a lesson for me here, as I got stuck trying to solve this problem and failed to find the place in the code this was previously solved.

We can trigger the `getDisplayDataFromBrowserStorage` action/logic in the `isInitialPlay` block of the `receiveTimeChange` callback in `ViewerPanel` and not have to worry about the default potentially coming in later than the goods from browser storage. No need for competing logics, or Promise handling. This logic will be removed entirely, and the `selection` branch logic can be simplified.

I've even used this code before! One thing I don't love about it is that the naming of `receiveTimeChange` doesn't remind me, at a glance, that it is part of initial loading, even though it is when you look closer. Maybe I'll factor some of that out and name it more explicitly.

src/containers/ModelPanel/index.tsx
interim17: Thanks again for all the thought and consideration here.

After digging into the `redux-logic` docs a bit, I'm coming down on the side of "synchronous business logic is appropriate for the website's `logics`", because 
1) It keeps a lid on the number of redux related files where you have business logic to keep track of.
2) The `transform`/`validate` hook in `redux-logic` seems to make this use case explicit: validating/altering the payload in some way and then either proceeding, rejecting the action, or providing a different action to proceed with, so I'm going to try to make use of that.

src/state/selection/logics.ts
```diff
@@ -1,46 +1,73 @@
...
+            const fileKey = getSimulariumFile(getState()).name;
+            localStorage.setItem(fileKey, JSON.stringify(newDisplayData));
+            allow(setSelectedUIDisplayData(newDisplayData));
+        } else {
+            reject(action);
+        }
+    },
+    process() {
+        return setCurrentColorSetting({
+            currentColorSetting: ColorSetting.UserSelected,
```
interim17, line 36: I haven't been making use of these execution phase hooks (`transform`/`validate`) or necessarily using the `process` hook very intentionally. I think this is a nice improvement and they definitely help clarify what's going on in cases where we may swap one action for another or dispatch actions if another one is successful.

meganrm (approved):

Issue #601, opened 2024-10-28
Bump antd 4 -> antd 5


interim17:
## Use Case
Ant design is one the design library we use for most of our components. The first `antd 5` versions are now two years old, and other AICS repos including Simularium front ends like NBSV use v5. Migrating is a chore but we should do it.

The primary difference, besides the availability of certain new components and features (`Flex`, multi value slider, etc) is moving from `less` to a `CSS-in-JS` approach and the use of `DesignTokens`. 

Hopefully our existing components, props, types and CSS modules/selectors will migrate easily, and we can simply find where we need to add styling to account for what `ant-vars.less` is still doing.

## Acceptance Criteria

Initial steps:
- [ ] bump the dependency
- [ ] remove `ant-vars.less`
- [ ] add styling as need (within current paradigm to cover what was being handled by `less`

Following phase/ongoing:
- [ ] adopt an approach to `CSS-in- JS`
- [ ] start migrating old stylesheets/selectors into `DesignTokens`

PR #609, opened 2024-11-04
bump antd and typescript, and update build tools and components to wo…

### LINKED ISSUE #604: bump antd dependency to v5
## Use Case

Advances #601 

## Acceptance Criteria

- install `antd v5`
- delete `ant-vars.less`
- remove `less` loading code from webpack
- define blank template `simulariumTheme` config to hold design tokens/theming to replace `less`
- wrap `App` in ConfigProvider, taking `simulariumTheme` as prop

Out of scope: actually providing configuration/styling globally or for specific components in `ConfigProvider`

interim17:
Time estimate or Size
=======
_small, but sensitive_

Problem
=======
Advances #601 and closes #604 

Solution
========
Installing `antd v5` led to a cascade of other dependency/build questions, and a variety of small fixes.

- bumping typescript to v5
- bumping node types to make github actions happy
- tweaking `webpack.config.json` to match `antd 5`
- tweaking `ts.config.base` and `.babelrc` to work with new typescript version, and some oddities I found
- removing dead code: `bindAll` function and tests, probably used when we had more class based components?
- small error/linting tweaks in a few components

with @pairperson1

## Type of change

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


package.json
```diff
@@ -35,6 +35,7 @@
         "@types/jest": "^26.0.24",
         "@types/jsdom": "12.2.2",
         "@types/lodash": "4.14.121",
+        "@types/node": "^22.8.4",
```
interim17, line 38: Some changes were running fine locally and hitting type errors when our github actions tried to run type checks, this fixed it.

src/components/AgentTreeSubmenu/index.tsx
```diff
@@ -11,7 +10,7 @@ import { CHECKBOX_TYPE_STAR } from "../../constants";
 interface AgentTreeSubmenuProps {
     checkedAgents: string[];
     options: CheckboxOptionType[];
-    onChange: (values: CheckboxValueType[]) => void;
+    onChange: (values: string[]) => void;
```
interim17, line 13: Could no longer import this type, doesn't seem to affect functionality.

src/util/index.ts
```diff
@@ -1,23 +1,6 @@
...
-) {
-    const setOfMethods = new Set(methods);
-
-    forOwn(obj.constructor.prototype, (value, key) => {
-        if (setOfMethods.has(value) && isFunction(value)) {
-            Object.assign(obj, { [key]: value.bind(obj) });
-        }
-    });
-}
-
```
interim17, line 20: dead code, AFAICT

.babelrc
```diff
@@ -13,7 +13,7 @@
   ],
   "presets": [
     "@babel/preset-react",
-    ["@babel/preset-env", { "modules": false }],
+    ["@babel/preset-env", { "modules": "auto" }],
```
interim17, line 16: see also, comments on `ts.config.base` but:
`babel` docs say when using a bundler like webpack this should always be auto, unless we are shipping native ES modules, which I do not believe we are.

tsconfig.json
```diff
@@ -1,8 +1,27 @@
...
+    "skipLibCheck": true,
+    "noEmit": true,
+    "preserveConstEnums": true,
+    "sourceMap": true,
+    "strict": true,
     "module": "ES6",
     "resolveJsonModule": true,
-    "esModuleInterop": true
-  }
+    "esModuleInterop": true,
```
interim17, line 16: Previously `tsconfig.base` was being extended into `tsconfig.json` and `tsconfig.commonjs.json`

To my eye the commonJS config is also dead code, and we can therefore consolidate the configs?

Also, there is no compile target, so the `target` field can be removed if `module` is provided.

meganrm (approved):
looks like a good start to this

frasercl (approved):
Looks good!

package.json
interim17: Good question.

PR #611, opened 2024-11-06
remove erroneous overflow bar

interim17:
Time estimate or Size
=======
_tiny_

Problem
=======
Empty viewer and trajectories with no plots were showing a scroll bar on the right side panel head.

Remove this overflow-y property. I think overflow styling on the side panels could use a touch up, but this is just buggy and needs a fix.

Screenshot 2024-11-06 at 3 24 45 PM

### DIFF
--- a/src/components/SideBarContents/style.css
+++ b/src/components/SideBarContents/style.css
@@ -2,7 +2,6 @@
     display: flex;
     flex-direction: column;
     height: 100%;
-    overflow-y: auto;
 }
 
 .container label {


frasercl (approved):

interim17:
> does this still scroll when there are more plots than fit in the height of the window?

yes, I think it's possible that design should be freshened up, but it still behaves as it should, as does the other panel

toloudis (approved):

PR #612, opened 2024-11-15
use native onClick in place of clickHandler prop in custom buttons

interim17:
Time estimate or Size
=======
_small_

`NavButton` and `ViewportButton` use a custom prop of `clickHandler` which is essentially the same as the antd `Button`'s native `onClick`

More symmetry in the props of different buttons (basic `antd` as well as our wrappers, and the `<button>` they eventually render in the DOM), is useful when managing focus and click/keyboard navigation behavior.

I don't recall a justification for using `clickHandler` instead of just `onClick` that I can remember so I'd like to switch them all to `onClick`.


toloudis (approved):
I don't know technically why onClick is better than clickHandler or whether there's any difference at all.  Apart from that, if everything's working, I approve!

interim17:
> I don't know technically why onClick is better than clickHandler or whether there's any difference at all. Apart from that, if everything's working, I approve!

In previous code no practical difference, but if for example you want to inject functions or otherwise work with different types of buttons, and some of them have `clickHandler` and some don't it adds complexity. It's nice if their props are the same whenever possible.

Realizing this as I work on keyboard navigation and wanted to just get ahead of it.

tyler-foster (approved):

PR #614, opened 2024-11-20
Feature/config provider

### LINKED ISSUE #603: define colors as ts constants
## Use Case

Advances #601 

antd v5 `ConfigProvider` and theming will injest hex colors defined as typescript constants, not as css vars.

Whatever colors and variables we are using in `ant-vars.less` will need to be ported over.

## Acceptance Criteria

The colors from `ant-vars.less` are declared and expored from a `themeColors.ts` file

interim17:
Time estimate or Size
=======
_small_

Problem
=======
Advances #601 Closes #603 #604

With antd bumped to v5, `ant-vars.less` is deprecated and replaced by `ConfigProvider`.

This PR is for making the switch and capturing the styles that were being handled by `less`.

It is _not_ a comprehensive fix of styling. Migrating lines from `ant-vars.less` is not sufficient to make things look right, and completing that process is out of scope.

Solution
========

- delete `ant.var.less` 
- wrap `App` with `ConfigProvider` and define `SimulariumTheme` with global and component tokens to *mostly* replace the styles applied in ant-vars.less
- TS base color variables updated to match CSS vars, semantic variable names grouped according to light and dark theme to reflect that same distinction in current CSS although there is only one `simulariumTheme` used in `ConfigProvider`

We can decide about having different antd theme algorithms going forward.

I think reasonable approval criteria here are:
- the `ConfigProvider` styles is being applied 
- nothing seems garishly out of place in terms of component tokens or file organization.

The point here is to get rid of `ant-vars.less`, migrate the gist of what styles it was still handling, and establish the `ConfigProvider`


src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,216 @@
...
+    colorLink: BLUE,
+
+    // Form
+    labelColor: DARK,
+
+    // Input
+    inputPlaceholderColor: WARM_GRAY,
+    inputBorderColor: WARM_GRAY,
+    inputBgColor: OFF_WHITE,
+    inputHoverColor: BABY_PURPLE,
```
meganrm, line 37: Overall this is nice, but some of the repetition here is more detailed than I think is necessary. For instance, I would have a single light AppBgColor, and then assign it below for both `colorBgContainer` and `inputBgColor` because you're always going to want those two to change together, instead of having a 1:1 of the token names in here. Similarly I would have a single `pageBgColor: WHITE_SIX` and use it for the base and the layout.  I'm also surprised you have to define the inputHoverColor separately, doesn't it already get the primaryColor?

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,216 @@
...
+    layoutFooterBg: DARK_THREE,
+    layoutLightTriggerColor: GRAYISH_BROWN,
+
+    // Menu
+    menuDarkItemColor: TRANSPARENT_WHITE,
+    menuDarkSelectedBg: BABY_PURPLE,
+    menuDarkSelectedColor: PURE_WHITE,
+
+    // Button
+    buttonPrimaryColor: BABY_PURPLE,
```
meganrm, line 64: again, doesn't this already get set by setting `colorPrimary`?

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,216 @@
...
+    checkboxPrimaryColor: WHITE_SIX,
+    checkboxBgColor: WHITE_SIX,
+    checkboxMarkColor: DARK,
+
+    // Tabs
+    tabsCardBg: LIGHT_PURPLE,
+    tabsInkBarColor: DARK_FOUR,
+    tabsItemHoverColor: DARK_FOUR,
+    tabsItemActiveColor: DARK_FOUR,
+    tabsItemSelectedColor: DARK_FOUR,
```
meganrm, line 80: combine these into one variable `tabsInteractionColor` or something

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,216 @@
...
+    // Tag
+    tagDefaultBg: ALLEN_PURPLE,
+    tagDefaultColor: WHITE_SIX,
+
+    // Slider
+    sliderRailBg: DARK_FOUR,
+    sliderRailHoverBg: WARM_GRAY,
+    sliderTrackBg: WHITE_THREE,
+    sliderHandleColor: WHITE_THREE,
+    sliderDotBorderColor: WHITE_THREE,
```
meganrm, line 99: also combine all these into one `sliderColor`

meganrm (changes requested):
looks good, I think you should reduce your variables in your mappings though. The goal is when we change a color, we can change the fewest lines of code possible. I think it's a safe assumption that if things are currently the same color as each other we'll want to keep them the same color. So your theme definitions can have a lot fewer variables than the antd theme, and just reuse them (which you're doing in some places already)

src/index.tsx
```diff
@@ -71,21 +73,23 @@ const App = () => {
...
-                    <ScrollToTop />
-                    {location.pathname !== EMBED_PATHNAME && (
-                        <Header>
-                            <AppHeader />
-                        </Header>
-                    )}
-                    <RouterSwitch />
-                </BrowserRouter>
-            </Layout>
+            <ConfigProvider theme={simulariumTheme} wave={{ disabled: true }}>
```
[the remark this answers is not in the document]
interim17: wave controls some animations that are hard to turn off individually

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,203 @@
...
+    modalContentBg: LIGHT_PURPLE,
+    modalTitleColor: DARK,
+};
+
+// Dark Theme Tokens
+const darkTheme = {
+    // Layout
+    layoutSiderBg: DARK,
+    layoutTriggerBg: DARK,
+    layoutTriggerColor: BABY_PURPLE,
```
meganrm, line 57: Same thing as before: you positive you need to set this separately when you already have primary color set to baby purple?

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,203 @@
...
+const GRAYISH_BROWN = "#4a4a4a";
+const BLUE = "#0094FF";
+const WARM_GRAY = "#979797";
+const TRANSPARENT_WHITE = "rgba(255, 255, 255, 0.65)";
+const TRANSPARENT = "transparent";
+
+const APP_BG_COLOR = OFF_WHITE;
+const TABS_INTERACTION_COLOR = DARK_FOUR;
+const SLIDER_COLOR = WHITE_THREE;
+const CHECKBOX_BG_COLOR = WHITE_SIX;
```
meganrm, line 22: Nit: just call this `LAYOUT_BG` and use it for `colorBgBase` and `colorBgLayout` below. or don't have both of those in your light theme and just set them in the actual theme.

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,203 @@
...
+    modalHeaderBg: LIGHT_PURPLE,
+    modalContentBg: LIGHT_PURPLE,
+    modalTitleColor: DARK,
+};
+
+// Dark Theme Tokens
+const darkTheme = {
+    // Layout
+    layoutSiderBg: DARK,
+    layoutTriggerBg: DARK,
```
meganrm, line 56: nit: combine these into one variable

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,203 @@
...
+    tabsCardBg: LIGHT_PURPLE,
+    tabsInteractionColor: TABS_INTERACTION_COLOR,
+
+    // Popover
+    popoverBgColor: CHARCOAL_GRAY,
+    popoverTextColor: PURE_WHITE,
+
+    // Dropdown
+    dropdownBgColor: CHARCOAL_GRAY,
+    dropdownTextColor: TRANSPARENT_WHITE,
```
meganrm, line 85: nit: call it `FADED_WHITE` because transparent sounds like it's going to be invisible

meganrm (approved):
looks good, I left a few more suggestions but nothing blocking

src/styles/theme/themeConfig.ts
```diff
@@ -0,0 +1,203 @@
...
+const TRANSPARENT_WHITE = "rgba(255, 255, 255, 0.65)";
+const TRANSPARENT = "transparent";
+
+const APP_BG_COLOR = OFF_WHITE;
+const TABS_INTERACTION_COLOR = DARK_FOUR;
+const SLIDER_COLOR = WHITE_THREE;
+const CHECKBOX_BG_COLOR = WHITE_SIX;
+
+// Light Theme Tokens (Global defaults, Modal, Radio)
+const lightTheme = {
```
frasercl, line 25: I've gotten confused working with this repo in the past by assuming that "light theme" and "dark theme" were different modes for the whole UI that could be switched between dynamically, rather than design concepts that statically apply to different UI elements (light theme for landing page, dark theme for viewer controls, etc.). Consider adding a few extra words to this comment to help out confused devs like my past self.

Also: even given that these "themes" apply to different UI elements, I'm a bit surprised that more effort isn't taken to unify the shapes of these objects, but that's out of scope for this PR anyways.

frasercl (approved):
Looks good!

src/styles/theme/themeConfig.ts
interim17: Hate to say it but I think I have found an even better shiny new way of organizing these variables, so this may code might not last that long. This uses styled-components so is pending some buy-in from reviewer and Megan but I like it and its pointed at this branch. It gets rid of the light and dark theme naming you see here.
https://github.com/simularium/simularium-website/pull/616

I agree that the light/dark theme is confusing, and I am also hoping to get rid of the light-theme.css file in favor of some CSS-in-JS and actually using themes for theming when things are called themes.

PR #615, opened 2024-11-23
Feature/restore default

### LINKED ISSUE #583: restore defaults button/ ability to toggle between default and changed color settings
## Use Case

Advances #511 

Once color settings have been changed (by user selections or by retrieving stored settings from the browser) the user should have the ability to preview and then apply the default color settings.

## Acceptance Criteria

Button in side panel should exist and be deactivated while the defaults are applied.
Once current settings do not match the default the button should become active.
On hover it should retain changes in state, but temporarily apply the default as a preview.
On click the default should be applied permanently and the changed colors should be removed from state and browser storage.

interim17:
Time estimate or Size
=======
_small_

Problem
=======
Closes #583 and #511 

Solution
========
Final steps for user color selections in browser storage!!

The button is in the side panel, hover over it? Preview the default settings. Stop hovering over it? Back to your selections.

Click it?? Restore default and delete selections from redux and browser storage.

If selections are in browser storage they should still be applied on load.

In keeping with recent discussion around the role of actions and logics in our redux flow, the business end of `clearUserSelectedColors` is in the logic, where we can run multiple dispatches without excess re-renders, and with logic free reducers.

Needed to make a few tweaks to `ColorPicker` so that re-renders wouldn't trigger a reset of `selectedUIDisplayData` when just trying to run a preview.

Steps to Verify:
----------------
1. Try a mix of making color changes, hovering over the button, clicking the button, and reloading the page in various states.


src/components/ColorPicker/index.tsx
```diff
@@ -32,7 +32,7 @@ const ColorPicker = ({
 }: ColorPickerProps): JSX.Element => {
     const [currentColor, setCurrentColor] = useState(initialColor);
     const [debouncedColor] = useDebounce(currentColor, 250);
-    const isInitialRender = useRef(true);
+    const isUserInteraction = useRef(false);
```
interim17, line 35: Now that other elements are interacting with color changing, this needed to be a little more robust so prevent re-renders from triggering this and making preview events into permanent selection changes.

src/state/compoundSelectors/index.ts
```diff
@@ -27,3 +28,17 @@ export const getCurrentUIData = createSelector(
             : defaultData;
     }
 );
+
+export const getDefaultUISettingsApplied = createSelector(
```
ShrimpCryptid, line 32: Consider changing to `getIsDefaultUISettingsApplied` (or `Are`) so this sounds more like a boolean/conditional value. Otherwise `getDefaultUISettingsApplied` sounds like it would be returning the default UI settings.

ShrimpCryptid (approved):
Love the new behavior, it's super rad having the instant preview! Tested with browser refresh and it works well.

FYI I am having trouble navigating to the button via keyboard shortcuts, but from the other PRs I've been seeing you open I assume you already know about it.

src/containers/ModelPanel/index.tsx
```diff
@@ -112,11 +125,35 @@ const ModelPanel: React.FC<ModelPanelProps> = ({
...
+        }
+    };
+
     return (
         <div className={styles.container}>
             <SideBarContents
                 mainTitle="Agents"
-                content={[contentMap[viewerStatus]]}
+                content={[
+                    <div key="molecules">
```
toloudis, line 139: is `molecules` maybe too specific a name for this?

toloudis (approved):
LGTM but one suggested renaming.

src/containers/ModelPanel/index.tsx
interim17: I wrote the first drafts of this a long time ago, guess I was feeling granular?

PR #616, opened 2024-12-04
styled-components.dropdown menu items

interim17:
Time estimate or Size
=======
Medium
Advances #601 

Problem
=======
Dropdown menu elements can be fussy to style as the base html elements receive different default, focus and hover rules front `ant`

Solution
========
As part of migrating to `ant5` I wanted to start using some CSS-in-JS and  `styled-components` is in use in other AICS tools like `timelapse-colorizer`. Setting up `styled-components` during the ant migration makes sense to me, if its something we might like to use.

### `DropdownMenuItem`
This file uses `styled-components` to create a uniformly styled element for dropdown menu items, regardless of whether the item is a button, an anchor tag, or a react router link.

It applies the common styles and then exposes three different `React.FC` exports for use in the menus.

### token/theme reorganization
Preferable to use `JS` instead of `CSS` variables in the `styled-components` which led me to once again reorganize the colors and theming.

I think this version is better, the complete set of base colors from the Figma style guide is included, a couple base colors have been consolidated, I've reduced redundancy as much as possible, and the whole set of base, semantic, and component colors is available via the `ThemeProvider`

### StyleProvider
Made a wrapper for the now accumulating styling providers so that `App` doesn't look cluttered.


src/components/CustomDropdown/DropdownMenuItems.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+// Dropdown items can be of a few component
+// varieties, including buttons, router links, and
+// anchor tags. This file unifies their styling,
+// and then divides them into three component exports
+// for semantically explicit usage in dropdowns.
+
+// Common styles
+const baseStyles = css`
+    font-family: ${(props) => props.theme.typography};
```
interim17, line 14: this is our only styled component for now, but we will probably want to add a `GlobalStyles` provider to get things like basic fonts into all styled components, and it could potentially replace some or all of the global style rules in style.css if we do that.

src/components/CustomDropdown/DropdownMenuItems.tsx
```diff
@@ -0,0 +1,144 @@
...
+    border: 2px solid ${({ theme }) => theme.colors.dropdown.background};
+    border-radius: 3px;
+    color: ${({ theme }) => theme.colors.dropdown.text};
+    cursor: pointer;
+    height: 28px;
+    padding: 4px;
+    width: 100%;
+    font-size: 14px;
+
+    &&& {
```
interim17, line 25: This is a nice way in `styled-components` to increase specifiicty, without chasing down lots of antd class names, and without using `!important` which overlooks specificity entirely.

src/styles/theme/colors.ts
```diff
@@ -0,0 +1,144 @@
...
+        transparent: "rgba(255, 255, 255, 0.65)",
+    },
+    // named darkest to lightest
+    dark: {
+        one: "#141219",
+        two: "#19171f",
+        three: "#1e1b25",
+        four: "#25222e", // fka three
+        five: "#2f2a3c", // fka four
+        six: "#3b3649", // fka charcoal gray, its purplish...
```
interim17, line 14: Can remove these comments, but in our *migration era* I think these are time savers when juggling old css vars and new color constants

src/styles/theme/colors.ts
```diff
@@ -0,0 +1,144 @@
...
+    // accent colors
+    primaryPurple: baseColors.purple.six,
+    activePurple: baseColors.purple.five,
+    primaryBlue: baseColors.blue,
+    primaryBorder: baseColors.gray.seven,
+    sliderColor: baseColors.gray.six,
+    buttonTextGray: baseColors.gray.five,
+    tagBg: baseColors.purple.nine,
+    tagText: baseColors.gray.nine,
+};
```
interim17, line 73: I think these can/will continue to evolve a bit as we dial in all styling, but file now has a basic way of exporting base colors, these semantic colors, and then component specific assignments. Spreading all three into the `themeColors` export makes any layer the color hierarchy (base, semantic, component) available via styled components theme props.

src/components/CustomDropdown/DropdownMenuItems.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const StyledRouterLink = styled(Link)`
+    ${baseStyles}
+    ${contentStyles}
+`;
+
+const StyledExternalLink = styled.a`
+    ${baseStyles}
+    ${contentStyles}
+`;
```
ShrimpCryptid, line 77: ✨✨✨✨✨✨

src/components/HelpMenu/index.tsx
```diff
@@ -12,24 +12,26 @@ import {
...
-                target="_blank"
-                rel="noopener noreferrer"
-            >
+            <DropdownRouterLink to={TUTORIAL_PATHNAME} newTab={true}>
                 Quick start
-            </Link>
+            </DropdownRouterLink>
         ) : (
-            <Link to={TUTORIAL_PATHNAME}>Quick start</Link>
+            <DropdownRouterLink key="Tutorial" to={TUTORIAL_PATHNAME}>
```
ShrimpCryptid, line 32: Why does this have a key when the other element doesn't? Should they both have keys?

src/styles/theme/colors.ts
ShrimpCryptid: migration era ✨

src/styles/theme/colors.ts
ShrimpCryptid: Would it make sense to not export base colors directly? That way you push development towards assigning specific semantic colors or component colors instead of relying on the hardcoded color directly.

ShrimpCryptid (commented):
Left a bunch of comments!

src/styles/theme/colors.ts
interim17: So, if I'm understanding you I think I do at the bottom of the file?
```
export const themeColors = {
    ...baseColors,
    ...semanticColors,
    ...componentColors,
};
```
I'm trying to balance both exporting all the most specific instances of the colors for use wherever they end up being needed, and the feedback I've received on previous PRs to consolidate anything that gets used repeatedly in similar contexts into these semantic variables.

src/components/CustomDropdown/DropdownMenuItems.tsx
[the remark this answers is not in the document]
interim17: According to intellisense the type isn't `any` it's `ExecutionContext & object` which I'm not entirely sure about, I can type theme as `DefaultTheme` from styled-components in the file where I export it, but that type is already picked up here in this file.

I'm not getting a typescript error, do you type props more explicitly in your implementation of `styled-components` in TFE?

src/components/HelpMenu/index.tsx
interim17: Not sure why I was inconsistent here, but in taking a look at it I think they aren't really needed, I could add where missing, but might also just remove most of them.

These components, if they were directly populating an array would each need keys, but the components are the values of the `label` properties, not directly filling the arrays so I think it's ok to just remove?

src/components/CustomDropdown/DropdownMenuItems.tsx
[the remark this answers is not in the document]
ShrimpCryptid: I haven't used the `props.theme` before in `StyledComponents`, I usually am getting my theme variables out of a context. Just curious if Intellisense/TS picked up the `Defaulttheme` type!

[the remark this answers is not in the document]
interim17: It did, and the ThemeProvider is a essentially a context

ShrimpCryptid (approved):
LGTM!

meganrm (approved):

PR #618, opened 2024-12-05
bug fix: handle popstate among networked trajectories while on viewer path

### LINKED ISSUE #617: forward/back is buggy
## Description

Hitting forward and back while the viewer is open and loading networked trajectories doesn't navigate, even if the url string updates.

## Expected Behavior

If we load trajectory A then B, then hit the back button, trajectory A should load. It should also work with navigating back to a networked trajectory from a local file

interim17:
Time estimate or Size
=======
small

Problem
=======
Closes #617 
If you are at `/viewer` hitting back won't actually load the previous trajectory. Same for forward, bug reproduces with local files as well (i.e. navigating back to networked trajectory URL after loading local file).

Solution
========
Added a new `useEffect` and `handlePopState` to `useLocationChange`

Didn't work with the `[location]` dependency of previously existing `useEffect` so made a new one.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -44,8 +44,10 @@ function useLocationChange() {
                 dispatch(setIsPlaying(false));
                 dispatch(clearSimulariumFile({ newFile: false }));
             });
-            return;
         }
+    }, [location]);
+
+    React.useEffect(() => {
         const handlePopState = () => {
             if (window.location.pathname === VIEWER_PATHNAME) {
                 const trajectoryId = getUrlParamValue(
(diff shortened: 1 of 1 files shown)


toloudis (approved):
I'm assuming this is working code... I had to google `popstate` event. do we use `pushstate` events too?

interim17:
> I'm assuming this is working code... I had to google `popstate` event. do we use `pushstate` events too?

`popstate` is a type of event, `pushState` is a method that triggers a `popstate` event. So yes, we call `pushState` explicitly in some places, but `popstate` events can be triggered other ways, like hitting the back button, and this PR is about handling those events more explicitly

ShrimpCryptid (approved):
LGTM!

PR #624, opened 2025-01-15
Feature/ant5 styling

interim17:
Time estimate or Size
=======
_small_
_hopefully the penultimate PR on this migration_

Problem / Solution
=======
Rounding up small styling and functionality fixes for `ant` migration.

The only thing that I intentionally left unfixed are buttons, which I decided deserved a bigger refactor, in the same vein as `DropdownMenuItems`. I will PR those changes into this branch.

I made the commit history a bit messy trying to keep various feature branches up to date with `main`, apologies.


toloudis (approved):
I'm not a good css reviewer but nothing looks too crazy here.  Should I be seeing a change in package.json that changes the ant dependency version?

interim17:
> I'm not a good css reviewer, so you may want to get one more set of eyes on it, but nothing looks too crazy here. Should I be seeing a change in package.json that changes the ant dependency version?

No, the dependency was bumped already in the target branch, and caused many small styling problems, this is just chasing those down.

src/components/ShareTrajectoryModal/index.tsx
```diff
@@ -138,9 +138,11 @@ const ShareTrajectoryModal = ({
         errorMessage: {
             content: (
                 <div>
-                    <h4>{Warn} The current file is stored on your device.</h4>
-                    <div>
-                        <h5>
+                    <h5 className={styles.localFileHeading}>
+                        {Warn} The current file is stored on your device.
+                    </h5>
```
ShrimpCryptid, line 143: Is `h5` just being used for styling here? Is there an `h4` element above this somewhere in the HTML hierarchy?

ShrimpCryptid (approved):

PR #625, opened 2025-01-15
Feature/custom button

interim17:
Time estimate or Size
=======
_small/medium_
_the final PR into the feature branch for this migration??_

Problem
=======
Button styling got messed up during `ant` migration

Also our approach to styling buttons is a bit all over the place.

Solution
========
Replace `NavButton` with more general `CustomButton`, a styled-component.

Delete: `NavButton` et. al, and `light-theme.css` and its directory.

The light-theme file was a bit of an anachronism from auto-conversion work.

### `ButtonClass`
Is expanded to cover all five of our basic button styled from the figma master guide (light and dark primary, light and dark secondary, and action buttons). Using the enum instead of raw strings is more type-safe and developer friendly.
Native antd `variant` prop is omitted and replaced with our own custom variants.

### `CustomButton`
As in the case of `DropdownMenuItems`, `antd` can be a bit heavy handed with class names, and also swaps out underlying html in certain cases.
Using a `styled-component` helps address specificity and consolidate the styling of buttons in one place.

Other buttons in the app (`ViewportButton` and `DropdownMenuItems`) are functionally and stylistically distinct, and either weren't much affected in migration or have already been addressed.


src/components/CustomButton/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+interface TooltipText {
+    defaultText: string;
+    disabledText?: string;
+}
+
+interface TooltilButtonProps extends CustomButtonProps {
+    tooltipText: TooltipText;
+    tooltipPlacement?: TooltipPlacement;
+}
```
interim17, line 27: Currently the style rules, basic button, and button with tooltip are all here in the same file. I'm open to the argument that they be split up! But at this stage I like having it all in one place.

src/components/CustomButton/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+export const TooltipButton: React.FC<TooltilButtonProps> = ({
+    tooltipText = { defaultText: "", disabledText: "" },
+    tooltipPlacement,
+    disabled = false,
+    ...buttonProps
+}) => {
+    const [tooltipRenderText, setTooltipRenderText] = useState(
+        disabled ? tooltipText.disabledText : tooltipText.defaultText
+    );
```
ShrimpCryptid, line 171: Does this need to be stateful? Is this to prevent the rendered text from changing while the tooltip is open?

ShrimpCryptid (approved):
Some minor comments but this approach is really elegant! LGTM!

src/components/LocalFileUpload/index.tsx
```diff
@@ -1,11 +1,12 @@
 import React from "react";
 import { Link } from "react-router-dom";
-import { Button, message, Upload, UploadProps } from "antd";
+import { message, Upload, UploadProps } from "antd";
 import { CloseOutlined } from "@ant-design/icons";
 import { RcFile } from "antd/lib/upload";
 
 import { ButtonClass } from "../../constants/interfaces";
 import { VIEWER_PATHNAME } from "../../routes";
+import { CustomButton } from "../CustomButton";
```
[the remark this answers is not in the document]
interim17: yes, I want that

src/components/CustomButton/index.tsx
interim17: Hey good catch, no it doesn't. Also the tooltip behavior wasn't actually working with the new disabled prop so this comment helped me catch that!

Adding a wrapper div to handle the mouse events so disabled buttons still show tooltips, and making `tooltipRenderText` a plain `const` since the button always re renders!

src/styles/colors.css
```diff
@@ -55,41 +56,9 @@
...
-    --dark-theme-secondary-button-hover-border: var(--light-purple-two);
-    --dark-theme-secondary-button-hover-color: var(--dark-two);
-    --dark-theme-secondary-button-focus-bg: var(--light-purple-two);
-    --dark-theme-secondary-button-focus-border: var(--dark-three);
-    --dark-theme-secondary-button-focus-outline: var(--light-purple-two);
-    --dark-theme-secondary-button-focus-color: var(--dark-two);
-    --dark-theme-secondary-button-disabled-bg: transparent;
-    --dark-theme-secondary-button-disabled-border: var(--dark-theme-header-bg);
-    --dark-theme-secondary-button-disabled-color: var(--dim-gray-two);
+    /* --dark-theme-primary-button-bg: var(--baby-purple);
```
tyler-foster, line 61: Are lines 60 and 61 here still necessary?

src/constants/interfaces.ts
```diff
@@ -42,8 +42,10 @@ export type TooltipPlacement =
     | "rightBottom";
 
 export enum ButtonClass {
```
tyler-foster, line 44: Too lazy to check this myself - is this enum used anywhere beyond the `CustomButton` component props? If not, would it be the right idea to just export it from `CustomButton`? That's what I would do, but I don't typically make big constant files like these.

src/components/CustomButton/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+        const buttonTheme = button[variant][theme];
+        const { background, text, border, hover, active, disabled } =
+            buttonTheme;
+
+        return css`
+            background-color: ${background};
+            border: 1px solid ${variant === "primary" ? background : border};
+            color: ${text};
+
+            &&& {
```
tyler-foster, line 67: Yeesh. Which brand of CSS magic does this triple ampersand belong to? Does this mean "parent-parent-parent"..?

tyler-foster (approved):
Smells reasonable to me!

Left some minor comments. Not sure I'm up to speed with the latest CSS technology.

src/components/CustomButton/index.tsx
interim17: No! It's both better and worse than that.

In `styled-components` this just directly raises the specificity of the current selector. Yes, this is arbitrary and smells a bit like using `!important` but for me its a needed workaround for quality of life when styling `antd`.

Because some `antd` components apply many different class names, change element nesting, and even changes which base html element is rendered for different components, it can be tricky to ensure that our styles will be consistently applied. This solves that in a brute force kind of way, but is better than using `!important` because these triple applied class hashes CAN be overridden by something of greater selector specificity, whereas `!important` ignores the specificity hierarchy entirely.

When our styles don't really map onto ant's opinions very well, I prefer to just sort of veto all their styling by using this triple ampersand.

tyler-foster: Thanks, I sort of it get it now.

Does there come a point where we eschew the ant buttons altogether and just roll our own from scratch 🤔 ? I'd be curious as to when the team would consider that to be worth it.
This isn't a request for changes though! Feel free to resolve.

interim17: I am trending towards a "I want to build it from scratch mindset" but also acknowledge the many benefits that do come form using this library. I think migrations like this are a point where we have to "pay" the deferred update cost that we would be paying in ongoing way when maintaining our own components.

The buttons may have reached a place where I would consider getting rid of the ant component, but in general ant does save us a lot of dev time.

PR #626, opened 2025-02-03
Feature/ant5

interim17:
Time estimate or Size
=======
?

Huge change set, but everything has been reviewed once already.
I would say run the brunch, look for any styling issues?

Problem
=======
Migrating ant 4 -> 5

Solution
========
Lot's of changes, happy to talk through previous PRs or any specific questions.

The broad overview:
- bumped `antd`, `typescript` and made some associated config changes
- removed some dead code
- made _lots_ of little changes to style sheets and component props because
- added theming via `antd` `ConfigProvider`
- added `styled-components`, integrated it into theming and used it to style a few components that are hard to target, or require a lot of styling overrides


rugeli (approved):
I have clicked around on the branch and the styling looked good, everything functional as expected! 

Super minor and non-blocking, but thought I'd mention it: when loading the blood plasma model, the `v.1.0` tag shows up first in the title and stays grayed out the whole time.

frasercl:
Poking around the UI, I noticed a few potentially problematic visual changes:
1. On the home page, hovering over the images in the example model cards causes the margin between the images and the description to shrink: 
image
2. The hover state of dropdown items has a new background, rather than a text color change, which reveals that the text is vertically off-center. (This may be intentional, and the centered-ness may be acceptable.)
image
3. The "help" dropdown is missing an outline that is present in the deployed version:
image
4. Plots have a new border radius and margin, which reveals a gap between the header ("Plots") and the first plot in the list which wasn't there before: 
image
5. Like Ruge said, the version tag in the title bar is barely readable.

Up to your discretion whether or not to address any one of these - I think only 1 and maybe 5 are true problems.

interim17:
@frasercl and @rugeli thanks for the comments

- homepage images: fixed
- help button border: fixed
- version tag colors: fixed
- dropdown hover items: background color change is correct, vertical text alignment fixed although this relates to an overpass font issue I'd like a more long term solution for
- plots: top margin is correct, border-radius is correct re convo w ux, I think some work on plot formatting is worth doing soon but out of scope

frasercl (approved):

PR #632, opened 2025-03-05
Landing page updates (homepage feature branch 2/4)

interim17:
Time estimate or Size
=======
_medium_

Problem
=======
Continuing work on homepage redesign. Merging to feature branch. 

Solution
========
Continue to move into dark theming and implementing the content pages redesign.

Add a few reusable components to be shared by landing page, tutorial page, and about page:
- `ContentPage`: contains content, contact panel, divider, footer
- `ContentPanel`
- `ContactPanel`

This is just to dry things out as the structure and styling of these three pages is very similar.

This PR adds these components and implements them along with new designs on the `LandingPage`, will do the About and Tutorial pages next.

Screenshots (optional):
-----------------------

Thanks for contributing!


src/components/ContentPage/style.css
```diff
@@ -0,0 +1,27 @@
...
+@media screen and (max-width: 480px) {
+    .page-content p {
+        font-size: 0.8rem;
+        margin-bottom: 1em;
+    }
+
+    .page-content li {
+        font-size: 0.8rem;
+    }
+}
```
interim17, line 22: Work on dynamic resizing/mobile is still pending, this is existing code moved here for now and unchanged.

src/components/ContentPage/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+    children: React.ReactNode;
+    className?: string;
+}
+
+const ContentPage: React.FC<ContentPageProps> = ({
+    children,
+    className = "",
+}) => {
+    return (
+        <React.Fragment>
```
meganrm, line 20: why not `<>` ? is this still using a very old version of React?

src/components/ContentPage/style.css
```diff
@@ -0,0 +1,27 @@
+.page-content {
+  display: block;
+  margin: auto;
+}
+
+@media screen and (min-width: 576px) {
+    .page-content h1, h2 {
+        margin-top: -0.3em;
```
meganrm, line 8: this is sorta a strange value, maybe add a comment about why this is necessary (I'm reading this before running it)

src/components/ContentPagePanel/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const panelClasses = `
+    ${styles.panel} 
+    ${isDark ? styles.darkSection : ""} 
+    ${className}
+  `;
+
+    const innerClasses = `
+    ${styles.panelInner}
+    ${isWide ? styles.wide : ""}
+  `;
```
meganrm, line 28: I'm pretty sure we have class-names installed, may be better to use that

src/components/ContentPage/index.tsx
interim17: Ah you're right, I just copied it over from the old page how it was an didn't notice to change it to `<>`

src/components/ContentPage/style.css
interim17: So this is old code I just moved and left a note on this PR that I haven't addressed the media queries and responsive sizing in the chunk of work. But I'll look at it when I get to that?

meganrm (approved):

frasercl (approved):

PR #643, opened 2025-05-27
fix: link hover state on homepage

interim17:
Time estimate or Size
=======
_x-small_

Problem
=======
The second of the two buttons in this container was wrapped wrong, so the hover styling wasn't applying to the whole component. I just made it match the one before in the container, which was working correctly.

Screenshot 2025-05-27 at 10 04 12 AM

### DIFF
--- a/src/components/LandingPage/index.tsx
+++ b/src/components/LandingPage/index.tsx
@@ -46,8 +46,8 @@ const LandingPage = (): JSX.Element => {
                     </Link>
 
                     <Link
-                        to={ABOUT_PATHNAME}
                         className={styles.borderedContent}
+                        to={ABOUT_PATHNAME}
                     >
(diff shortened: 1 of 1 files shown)


frasercl (approved):

src/components/LandingPage/index.tsx
```diff
@@ -44,15 +44,19 @@ const LandingPage = (): JSX.Element => {
                             </p>
                         </div>
                     </Link>
-                    <div className={styles.borderedContent}>
+
+                    <Link
+                        to={ABOUT_PATHNAME}
+                        className={styles.borderedContent}
```
ShrimpCryptid, line 50:
```suggestion
                        className={styles.borderedContent}
                        to={ABOUT_PATHNAME}
```
Teeny nit: reverse to match above

ShrimpCryptid (approved):

PR #650, opened 2025-07-08
styling cleanup and remove lingering focus state on homepage button

interim17:
Time estimate or Size
=======
_small_

Problem
=======
A number of PRs have been merged into the `feature/homepage` branch, I'd love to bring it into main.

This is a small set of tweaks on spacing. There is more that can proceed on implementing those designs, but we don't want to fix everything in the feature branch, this is a cleanup of some spacing, font sizing, and typos I noticed.

### DIFF
--- a/src/components/ContactPanel/style.css
+++ b/src/components/ContactPanel/style.css
@@ -1,0 +1,3 @@
+.header-spacing {
+    margin-bottom: 16px;
+}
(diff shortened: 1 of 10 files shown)


src/components/ContactPanel/style.css
```diff
@@ -0,0 +1,3 @@
+.header-spacing {
+    margin-bottom: 16px;
+}
```
ShrimpCryptid, line 3: If this class is being used in multiple places, can you define it once in a global stylesheet and import it wherever it's being used?
https://developer.mozilla.org/en-US/docs/Web/CSS/@import

ShrimpCryptid (approved):
Non-blocking comments!

toloudis (approved):
I didn't test it or anything but the changes look reasonable.  I generally agree with Peyton's comment about h1-h2-h3 but can't evaluate it for this specific case.

PR #655, opened 2025-08-14
update jest and babel to work with es modules from viewer

interim17:
Hacked around on jest and babel configs to get it happy with es modules.

Had to install the es version of lodash, and point the test suite a particular versions of uuid, otherwise seems relatively straightforward?

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
         "webpack-dev-server": "^4.10.1"
     },
     "dependencies": {
-        "@aics/simularium-viewer": "^3.10.2",
+        "@aics/simularium-viewer": "^3.10.3",
         "@ant-design/css-animation": "^1.7.3",
         "@ant-design/icons": "^4.8.3",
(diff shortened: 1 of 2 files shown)


interim17:
@meganrm for context from my discussion with Alli earlier, it seems like the new viewer patch is installing and building successfully, but jest was failing to resolve ES module imports

package.json
```diff
@@ -85,7 +85,6 @@
         "sinon": "7.2.4",
         "style-loader": "3.3.1",
         "ts-jest": "^29.0.3",
-        "ts-node": "10.9.1",
```
ascibisz, line 88: why did `ts-node` get removed?

interim17: I'll just put it back to keep in scope, extra dev dependencies not a huge deal, but I don't think we need it currently.

ascibisz (approved):

meganrm (approved):

